PR #3, opened 2017-02-10
Replace ace calls to standard functions

dgkimura:
Rather than call ACE library functions, lets try using standard functions, like `std::snprintf` which was introduced in C++11.

### DIFF
--- a/src/cppcache/src/CacheXmlParser.cpp
+++ b/src/cppcache/src/CacheXmlParser.cpp
@@ -20,6 +20,7 @@
 #include <gfcpp/PoolFactory.hpp>
 #include "AutoDelete.hpp"
 #include "CacheImpl.hpp"
+#include <cstring>
 
 #if defined(_WIN32)
 #include <windows.h>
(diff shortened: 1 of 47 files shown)


src/cppcache/src/CacheXmlParser.cpp
```diff
@@ -1165,7 +1166,7 @@ void CacheXmlParser::startPersistenceManager(const xmlChar** atts) {
         throw CacheXmlException(s.c_str());
       }
 
-      ACE_OS::strncpy(libraryFunctionName, (char*)atts[i], len);
+      std::strncpy(libraryFunctionName, (char*)atts[i], len);
```
pivotal-jbarrett: Maybe not is a great opportunity to fix these c-style casts that are warnings in C++11 now when you are fixing other things in the file.

src/cppcache/src/TcrMessage.hpp
```diff
@@ -1170,7 +1170,7 @@ class TcrMessageHelper {
     }
     if (compId != expectedPartType) {
       char exMsg[256];
-      ACE_OS::snprintf(exMsg, 255,
+      std::snprintf(exMsg, 255,
```
pivotal-jbarrett: This change violates the C++ formatting we adhere to. Please reformat each changed line or file after global search and replace.

pivotal-jbarrett: Technically its the lines following this change that violate the formatting.

dgkimura: Is there a command line tool (e.g. lint) that I can run to ensure changes adhere to the style guide?  Also to be clear, it's the fact that the following lines are indented more than 2 spaces (than the line above) that is in violation?

src/cppcache/src/CacheXmlParser.cpp
dgkimura: Maybe this would be better as a separate pull request for fixing all c-style casts?  However, if you feel it's more effective to update them incrementally as we make changes then I will add them to this pull request.

src/cppcache/src/TcrMessage.hpp
pivotal-jbarrett: We don't have a good command to execute. I suggest looking for integration into what ever editor you have with clang-format. There appears to be some differences between clang-format on mac and windows so limit the formatting to files you are editing. 

For example, in Eclipse, I have reformat on save configured. 

The following lines used to be aligned with the changed line's open paren. The guide defines specific rules for when to align to the paren or indent.

src/cppcache/src/CacheXmlParser.cpp
pivotal-jbarrett: I am of the camp that believes that large efforts like that are rarely undertaken unless there is an easy tool to do the work for you. The clang-tidy rule fixed some when we did the initial reformat and cleanup to Google C++ Style Guide. There were some it was unable to do automagically, my guess because they were ambiguous. I would make an effort that any file you edit should compile clean, no warnings and no static analyzer warnings/errors. You can enable clang-tidy by passing CLANG_TIDY_ENABLED=ON to CMake.

dgkimura: Alright.  Thanks for the clang-tidy tip.  I'll add these changes and update the pull request.

src/cppcache/src/TcrMessage.hpp
dgkimura: Ah, thanks.  I didn't see that.  I'll update the PR.

src/cppcache/src/CacheXmlParser.cpp
dgkimura: Okay, so `xmlChar` is an `unsigned char` and `strncpy ` requires a `char`, hence a c-style cast/hack here.  That seems broken.  A fix may be using `std::memcpy`.  However, other calls, like `strlen` seem fundamentally broken.  I think fixing these calls may be out of scope for this PR.

pivotal-jbarrett: The C++ way to do this now is `reinterpret_cast<const char*>(atts[i])`.

PR #6, opened 2017-02-12
GEODE-2470: Replace sed dependency with standard C++.

mmartell:
- sed is used in CacheHelper to do string replacement of port number
variables in the input xml configuration files.

### DIFF
--- a/src/cppcache/integration-test/CacheHelper.cpp
+++ b/src/cppcache/integration-test/CacheHelper.cpp
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #include <gfcpp/GeodeCppCache.hpp>
 #include <stdlib.h>
 #include <gfcpp/SystemProperties.hpp>
@@ -21,6 +22,7 @@
 #include <ace/INET_Addr.h>
 #include <ace/SOCK_Acceptor.h>
 #include <fstream>
+#include <regex>
 
 #include "TimeBomb.hpp"
 #include <list>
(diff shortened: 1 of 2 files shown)


pivotal-jbarrett:
Why not use regex_replace in C++11 http://en.cppreference.com/w/cpp/regex/regex_replace?

mmartell:
Simplified by switching to regex_replace instead custom function.

src/cppcache/integration-test/CacheHelper.cpp
```diff
@@ -1382,6 +1377,31 @@ void CacheHelper::createDuplicateXMLFile(std::string& originalFile,
...
 
+void CacheHelper::replacePortsInFile(int hostPort1, int hostPort2,
+                                     int hostPort3, int hostPort4, int locPort1,
+                                     int locPort2, const std::string& inFile,
+                                     const std::string& outFile) {
+  std::ifstream in(inFile, std::ios::in | std::ios::binary);
+  if (in) {
+    std::string contents;
+    in.seekg(0, std::ios::end);
+    contents.resize(in.tellg());
```
pivotal-jbarrett: I think you want `reserve()`. `resize()` causes the string to be resized and initialized with `CharT()`. `reserve()` just allocates enough capacity but does not resize or initialize the string.

src/cppcache/integration-test/CacheHelper.cpp
```diff
@@ -1382,6 +1377,31 @@ void CacheHelper::createDuplicateXMLFile(std::string& originalFile,
...
+    in.seekg(0, std::ios::beg);
+    in.read(&contents[0], contents.size());
```
pivotal-jbarrett:
```
contents.assign((std::istreambuf_iterator<char>(in)),
    std::istreambuf_iterator<char>());
```

src/cppcache/integration-test/CacheHelper.cpp
```diff
@@ -1382,6 +1377,31 @@ void CacheHelper::createDuplicateXMLFile(std::string& originalFile,
...
+
+    contents = std::regex_replace(contents, std::regex("HOST_PORT1"), std::to_string(hostPort1));
+    contents = std::regex_replace(contents, std::regex("HOST_PORT2"), std::to_string(hostPort2));
+    contents = std::regex_replace(contents, std::regex("HOST_PORT3"), std::to_string(hostPort3));
+    contents = std::regex_replace(contents, std::regex("HOST_PORT4"), std::to_string(hostPort4));
+    contents = std::regex_replace(contents, std::regex("LOC_PORT1"), std::to_string(locPort1));
+    contents = std::regex_replace(contents, std::regex("LOC_PORT2"), std::to_string(locPort2));
+
+    std::ofstream out(outFile, std::ios::out);
+    out.write(&contents[0], contents.size());
```
pivotal-jbarrett:
```
out << contents;
out.close();
```

src/cppcache/integration-test/CacheHelper.cpp
mmartell: Good catch.

mmartell:
As per your suggestions, switched to using assign with istreambuf_iterators.

PR #7, opened 2017-02-12
Feature/geode 2467

mmartell:


PivotalSarge:
LGTM

mmartell:
Removed the extraneous commit. This PR now only has the commit to change the Xerces CMakeLists.txt.

src/dependencies/xerces-c/CMakeLists.txt
```diff
@@ -32,10 +32,10 @@ if (WIN32)
     set( _BUILD_DIR Build/Win32/VC${MSVC_VERSION}/${_DEBUG_OR_RELEASE} )
   endif()
 
-  if (MSVC_VERSION GREATER 12)
-    # Only have project files for VS12 and older
-    set(MSVC_VERSION 12)
-  endiF()
+  #if (MSVC_VERSION GREATER 12)
```
pivotal-jbarrett: xerces-c does not have project files for VC>14. You should limit the MSVC_VERSION locally in the module to 14 otherwise xerces-c will not compile. See the `projects\Win32\` directory in the xerces-c distribution for list of supported VC.

pivotal-jbarrett: Also, if we are removing code we should remove it, not comment it out.

mmartell: It actually builds fine with Visual Studio 2017. They probably just don't test with release candidates, so they haven't update their docs. Official release date is Mar 7.

I left the code in but commented out in case we wanted to put it back in for some reason.

mmartell:
Bumped the version check to allow highest VS version (14).

PR #12, opened 2017-02-15
GEODE-2309: Remove or ignore apache-rat flagged files

dgkimura:
`resolv.config` used to work around Solaris SPARC networking issues on a local OpenStack instance.  It should not be needed going forward.

### DIFF
--- a/.ratignore
+++ b/.ratignore
@@ -23,10 +23,12 @@
 
 # expect script
 .*changepasswd$
+.*winrm.cloud-init$
 
 # doxygen
 .*package-list$
 .*testframeworkdox.txt$
+docs
 
 # patches
 .*patches$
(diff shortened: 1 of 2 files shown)


.ratignore
```diff
@@ -23,10 +23,12 @@
 
 # expect script
 .*changepasswd$
+.*winrm.cloud-init$
 
 # doxygen
 .*package-list$
 .*testframeworkdox.txt$
+docs
```
pivotal-jbarrett: This will match any file with "docs" in it. Seems like a rather greedy regex.

dgkimura: Good point.

However, I tested that this seems to work.  I created a file `somedocs` and `docsblah`, both without licenses.  When I ran `rat` it flagged both of those files, which would make it seem like it doesn't treat this change as a regex?

I couldn't seem to find apache-rat documentation for their regex.  If you know where they are could you point me to them?

`$ java -jar ~/tools/apache-rat-0.12/apache-rat-0.12.jar --dir . --exclude-file .ratignore`

```
*****************************************************

 Printing headers for text files without a valid license header...
 
=====================================================
== File: ./docsblah
=====================================================
no license

=====================================================
== File: ./somedocs
=====================================================
no license

=====================================================
```

pivotal-jbarrett: Hmm.. ok.

PR #13, opened 2017-02-15
GEODE-2476: Replace gfcpp with geode.

PivotalSarge:
- Rename directories and files with gfcpp into their name to
  instead use geode and update all references thereto.
- Remove the gfcpp executable.
- Rename gfcpp.properties to geode.properties and update all
  references thereto.
- Ensure formatting style guide compliance.
- Elide duplicate GEODE_ from include guards.


src/CMakeLists.txt
```diff
@@ -224,7 +222,7 @@ add_subdirectory(cppcache)
 add_subdirectory(cryptoimpl)
 add_subdirectory(dhimpl)
 add_subdirectory(sqliteimpl)
-add_subdirectory(gfcpp)
+add_subdirectory(getversion)
```
pivotal-jbarrett: Eeeek! This is the best name we could come up with for this stupid util? May we should just delete it?

src/cppcache/include/geode/AttributesFactory.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-#ifndef GEODE_GFCPP_ATTRIBUTESFACTORY_H_
-#define GEODE_GFCPP_ATTRIBUTESFACTORY_H_
+#ifndef GEODE_GEODE_ATTRIBUTESFACTORY_H_
```
pivotal-jbarrett: Really? GEODE_GEODE doesn't seem right?!

src/cppcache/include/geode/AttributesFactory.hpp
```diff
@@ -20,7 +20,7 @@
  * limitations under the License.
  */
 
-#include "gfcpp_globals.hpp"
+#include "geode_globals.hpp"
 #include "gf_types.hpp"
```
pivotal-jbarrett: Why not gf_types.hpp renamed to geode_types.hpp?

src/cppcache/include/geode/AttributesFactory.hpp
mmartell: +1 for GEODE_ATTRIBUTES_FACTORY_H_

src/cppcache/include/geode/AttributesFactory.hpp
mmartell: +1 for geode_types.hpp

PivotalSarge: That's a separate JIRA ticket and hence will be a separate (upcoming) PR.

src/cppcache/include/geode/AttributesFactory.hpp
PivotalSarge: Google specification is project name, directory name, file name.

src/CMakeLists.txt
PivotalSarge: I'm totally cool with deleting it. I didn't because it's been part of the install; some users may rely on it to determine version?

pivotal-jbarrett: If they are relying on it then their scripts will break either way. There are better ways to get version. I don't think I have ever seen anyone anywhere use it. I don't even see it in our documents. I say kill it!

PR #15, opened 2017-02-16
GEODE-2484: Fix snprintf error. Need #include <cstdio>.

mmartell:


mmartell:
Whoops. Forgot to format. Fixing now.

src/cppcache/src/TcrMessage.hpp
```diff
@@ -1135,7 +1137,7 @@ class TcrMessageHelper {
         return EXCEPTION;
       } else {
         char exMsg[256];
-        std::snprintf(
+        snprintf(
```
dgkimura: Why are you removing the namespace specifier here?

metatype:
Did you tag the wrong JIRA issue?  I don't see anything related to GEODE-2484.

mmartell:
David,

I had to remove the std:: to get it to build on Windows. I tested my fix on
Windows and Mac only.

src/cppcache/src/TcrMessage.hpp
```diff
@@ -1114,10 +1116,10 @@ class TcrMessageHelper {
     } else if (!isObj) {
       // otherwise we're currently always expecting an object
       char exMsg[256];
-      std::snprintf(exMsg, 255,
-                    "TcrMessageHelper::readChunkPartHeader: "
-                    "%s: part is not object",
-                    methodName);
+      snprintf(exMsg, 255,
```
pivotal-jbarrett: If you had to remove std:: to make this work then it isn't working like you think it is. Let's figure out why and fix correctly,

mmartell:
Something is screwed up in our #includes. TcrMessage.cpp compiles fine
leaving the std:: in there. However, other classes that include it don't
compile (like Distributed System).

I'd rather spend time switching to newer constructs (<stream> and
<upstream>). Simple change, and for now we can convert the streamstrimg
back to char* when passing to logger.

src/cppcache/src/TcrMessage.hpp
```diff
@@ -20,6 +20,10 @@
  * limitations under the License.
  */
 
+//#include <cstdio>
```
dgkimura: No need for this line in current revision.

src/cppcache/src/TcrMessage.hpp
```diff
@@ -1113,12 +1117,11 @@ class TcrMessageHelper {
...
-      char exMsg[256];
-      std::snprintf(exMsg, 255,
-                    "TcrMessageHelper::readChunkPartHeader: "
-                    "%s: part is not object",
-                    methodName);
-      LOGDEBUG("%s ", exMsg);
+
+      std::stringstream s;
+      s << "TcrMessageHelper::readChunkPartHeader: " << methodName << ": part is not object\n";
+      LOGDEBUG("%s ", s.str().c_str());
```
dgkimura: I find this line a little suspicious.  If I understand correctly, `stringstream.str()` will return a temporary string object and then `c_str()` will return a pointer to that temporary object.  Then it seems like it may be possible that temporary string is freed after the expression is evaluated.  And then we would be referencing a freed object inside `LOGDEBUG` which would lead to undefined behavior.  If I am misunderstanding then please educate me how this works and why you chose to do it!  :)

pivotal-jbarrett: Good question. I am still fascinated as to why std::snprintf isn't working correctly on Windows too.

PR #17, opened 2017-02-17
GEODE-2440: Switch hashcode() return type.

mmartell:
- Convert from uint32_t to int32_t to match server.


pivotal-jbarrett:
Please close this pull request and resubmit when you have cleaned up all the places that use hashcode.

mmartell:
Close for now. Resubmit after implementing jbarrett's recommendations.

PR #18, opened 2017-02-17
Revert "GEODE-2484 Replace ace calls to standard functions"

dgkimura:
Revert build break on VS 2013 which does not have complete
C++11 features, e.g. std::snprintf.

This reverts commit d4e0a8447b1e54d5c34a35069bc24a30d8258e2e.


echobravopapa:
Wise choice, this can revisited in the future when the community has more time and inclination to make the jump to a toolchain with complete or more complete C++11 support.

PR #19, opened 2017-02-17
GEODE-2508: Inital work on making lib names generic.

echobravopapa:
Just a sneak peak, I think I may have a couple tweaks/additions next week.


src/cppcache/integration-test/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ target_link_libraries(${TEST_UTILS_LIB}
   PRIVATE
     ACE
   PUBLIC
-    apache-geode
+    ${PRODUCT_LIB_NAME}
```
pivotal-jbarrett: Rather than changing the target name of the library throughout CMake by using a variable I suggest just controlling the output name of the binary using the OUTPUT_NAME property.

set_target_properties(apache-geode PROPERTIES OUTPUT_NAME ${PRODUCT_LIB_NAME})

This way the target stays "apache-geode" but the results of the output will be adjusted.

echobravopapa:
Closing to take different approach.

PR #21, opened 2017-02-22
GEODE-2441: Remove pdx auto serializer from core sources

dgkimura:
Purpose of this pull-request is to simplify core dependencies and core source size by removing PdxAutoSerializer and related tests.  Reasoning is that PdxAutoSerializer is more of a standalone utility project that adds nearly 40,000 lines of extra complexity.

A followup change will be to move PdxAutoSerializer as a stand-alone utility either into a `contrib` directory in `geode-native` repo or somewhere inside `geode-examples` repo.


davinash:
LGTM

PR #23, opened 2017-02-22
GEODE-2478: Replace gf with geode.

PivotalSarge:
- Rename directories and files with gf into their name to
  instead use geode and update all references thereto.
- Ensure formatting style guide compliance.
- Elide duplicate GEODE_ from include guards.


PivotalSarge:
Good idea. I think you're right about redundancies. I'll look into that.

Sarge

PR #24, opened 2017-02-22
GEODE-2508: Initial work on new approach to generic lib naming.

echobravopapa:
Keeping the changes to CppCacheLibrary.cpp - did not, yet, find a CMake way to do it more seamlessly.

### DIFF
--- a/src/cppcache/src/CMakeLists.txt
+++ b/src/cppcache/src/CMakeLists.txt
@@ -125,7 +125,9 @@ target_include_directories(apache-geode
     $<TARGET_PROPERTY:_apache-geode,INTERFACE_INCLUDE_DIRECTORIES>
 )
 add_dependencies(client-libraries apache-geode)
-set_target_properties(apache-geode PROPERTIES PUBLIC_HEADER "${PUBLIC_HEADERS}" OUTPUT_NAME ${PRODUCT_LIB_NAME} )
+set_target_properties(apache-geode PROPERTIES
+  PUBLIC_HEADER "${PUBLIC_HEADERS}"
+  OUTPUT_NAME ${PRODUCT_LIB_NAME})
 
 add_library(apache-geode-static STATIC ${SOURCES_ALL})
 target_link_libraries(apache-geode-static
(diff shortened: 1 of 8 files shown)


pivotal-jbarrett:
@PivotalSarge yes.

src/cppcache/src/CMakeLists.txt
```diff
@@ -125,7 +125,7 @@ target_include_directories(apache-geode
     $<TARGET_PROPERTY:_apache-geode,INTERFACE_INCLUDE_DIRECTORIES>
 )
 add_dependencies(client-libraries apache-geode)
-set_target_properties(apache-geode PROPERTIES PUBLIC_HEADER "${PUBLIC_HEADERS}")
+set_target_properties(apache-geode PROPERTIES PUBLIC_HEADER "${PUBLIC_HEADERS}" OUTPUT_NAME ${PRODUCT_LIB_NAME} )
```
pivotal-jbarrett: I would only suggest avoiding long lines in the CMake files by putting each property on a new line.

```
set_target_properties(apache-geode PROPERTIES
    PUBLIC_HEADER "${PUBLIC_HEADERS}"
    OUTPUT_NAME ${PRODUCT_LIB_NAME})
```

src/cppcache/src/CppCacheLibrary.cpp
```diff
@@ -127,15 +127,18 @@ std::string CppCacheLibrary::getProductLibDir() {
   for (int i = 0; i < PATH_MAX && path[i] != 0; i++) {
     path[i] = ::tolower(path[i]);
   }
-  dllNamePtr = strstr(path, "apache-geode.dll");
+  std::string cppName = PRODUCT_LIB_NAME;
+  cppName += ".dll";
+  std::string dotNetName = PRODUCT_DLL_NAME;
+  dotNetName += ".dll";
+  dllNamePtr = strstr(path, cppName.c_str());
```
dgkimura: Nit: `strstr` seems old-school.  If you turn `path` into a `std::string` then you can use `std::basic_string::find` and do something like.

```cpp
if (pathAsString.find(cppName) == std::string::npos)
{
    // ...
}
```

src/cppcache/src/CMakeLists.txt
echobravopapa: will adjust that.  cmake did not like duplicate calls to set_target_properites

src/cppcache/src/CppCacheLibrary.cpp
echobravopapa: fair nit 👍 will change it

PR #25, opened 2017-02-22
GEODE-2382: Replace gemstone with geode.

PivotalSarge:
- Remove AssemblyCompany.
- Migrate package hierarchy from com.gemstone.gemfire to org.apache.geode.
- Rename from gemstone.txt to geode.txt.


src/xsds/gfcpp-cache-9.0.xsd
```diff
@@ -33,7 +33,7 @@ limitations under the License.
       version="9.0">
 
 The contents of a declarative XML file correspond to APIs found in the
-com.gemstone.gemfire.cache and com.gemstone.gemfire.cache.client
+org.apache.geode.cache and org.apache.geode.cache.client
```
pivotal-jbarrett: Have you verified that these are the correct packages after GemFire became Geode?

PivotalSarge: I got those packages names off a checkout of the develop branch of geode yesterday so I'm pretty confident they're right.

PR #30, opened 2017-02-23
GEODE-2532: Create config files in runtime directory.

mmartell:
Some config xml files were being created in the source directory instead of the runtime test directory.

### DIFF
--- a/src/clicache/integration-test/ThinClientPoolTestsN.cs
+++ b/src/clicache/integration-test/ThinClientPoolTestsN.cs
@@ -234,7 +234,7 @@ namespace Apache.Geode.Client.UnitTests
     {
       string xmlLocation = CacheHelper.TestDir + Path.DirectorySeparatorChar + "valid_cache_pool.xml";
 
-      string duplicateXMLFile = CacheHelper.TestDir + Path.DirectorySeparatorChar + Util.Rand(3432898).ToString() + "valid_cache_pool.xml";
+      string duplicateXMLFile = Util.Rand(3432898).ToString() + "valid_cache_pool.xml";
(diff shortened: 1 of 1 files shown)


pivotal-jbarrett:
Does this change create the xml files in each of the individual integration test's runtime directory of the root of the all the tests?

echobravopapa:
@pivotal-jbarrett good question, @mmartell where do the files go to now?

mmartell:
The xml files are created in the runtime working directory. For a Debug build this is:
BUILD_DIR\clicache\integration-test\Debug, where BUILD_DIR is the cmake build directory.

This fix moves those few xml files that were being created in the source directory to the working directory to be consistent with the rest of the tests.

mmartell:
@echobravopapa There is an inconsistency in how a few of the tests compute the path for the configured xml file. After inspection, there didn't appear to be a simple high level way to fix this. My proposed fix is to fix it in the isolated file that causes the problem.

pivotal-jbarrett:
The C++ integration test framework keeps the test files in a directory specific to the current test. Can we not do the same thing here and avoid this random thing?

mmartell:
@pivotal-jbarrett It seems like a very good suggestion to keep all test files local to each test subdirectory. That would avoid potential name collision of these configured xml files. It looks like an attempt was made to avoid such file name collisions by appending a random number, but that isn't a guarantee.

I've created a new ticket (GEODE-2540) to move these xml config files to each local test subdirectory. That is a much bigger task which will require refactoring the NUnit based tests to extract the current test name. 

For now, I'd like to suggest we accept this PR which solves the problem of polluting the src tree with these test files.

PR #33, opened 2017-02-24
GEODE-2441: Remove antlr reference from core sources

dgkimura:
References to antlr are already deleted, however in the verbosity of the previous pull-request this part was missed.  It may be worth noting that the standalone PdxAutoSerializer utility has its own antlr CMakeLists.txt.


dgkimura:
@PivotalSarge This is the relevant code after the patch is applied so my guess is there was a workaround in `antlr` that was probably non-portable.

```cpp
#if ( _MSC_VER == 1200 )
// VC6 seems to need this
// note that this is not a standard C++ include file.
# include <stdio.h>
#else
#include <cstring>
#include <cstdio>
#endif
```

pivotal-jbarrett:
@PivotalSarge CMakeList.txt.in is there to build Antlr with CMake rather than the very complicated and broken build process it ships with.

PR #36, opened 2017-02-25
GEODE-2494: Replace SpinLock with spinlock_mutex.

pivotal-jbarrett:
Replaces our non-standard SpinLock with spinlock_mutex that implements the C++11 standard BasicLockable. Where applicable the spin lock was completely removed and replaced with other thread safe constructs. Singleton protection, until singletons are removed entirely, converted to C++11 thread synchronized static initialization. Atomic operations converted to std::atomic.

Can I get some reviews form the other C++ contributors before I merge this lovely change?


src/cppcache/src/CqQueryVsdStats.cpp
```diff
@@ -108,21 +98,21 @@ CqQueryStatType::CqQueryStatType()
...
 
 CqQueryVsdStats::CqQueryVsdStats(const char* cqqueryName) {
-  CqQueryStatType* regStatType = CqQueryStatType::getInstance();
+  auto& regStatType = CqQueryStatType::getInstance();
 
-  StatisticsType* statsType = regStatType->getStatType();
+  StatisticsType* statsType = regStatType.getStatType();
 
-  GF_D_ASSERT(statsType != NULL);
+  GF_D_ASSERT(statsType != nullptr);
```
dgkimura: Would `static_assert` work here?

src/cppcache/src/LRUList.cpp
```diff
@@ -15,12 +15,18 @@
...
-#define LOCK_HEAD SpinLockGuard headLockGuard(m_headLock)
-#define LOCK_TAIL SpinLockGuard tailLockGuard(m_tailLock)
+namespace apache {
+namespace geode {
+namespace client {
+
+using util::concurrent::spinlock_mutex;
+
+//#define LOCK_HEAD SpinLockGuard headLockGuard(m_headLock)
+//#define LOCK_TAIL SpinLockGuard tailLockGuard(m_tailLock)
```
dgkimura: Delete?

src/cppcache/src/LRUList.cpp
```diff
@@ -96,33 +102,38 @@ void LRUList<TEntry, TCreateEntry>::getLRUEntry(LRUListEntryPtr& result) {
 
 template <typename TEntry, typename TCreateEntry>
 typename LRUList<TEntry, TCreateEntry>::LRUListNode*
-LRUList<TEntry, TCreateEntry>::getHeadNode(bool& isLast) {
-  LOCK_HEAD;
+LRUList<TEntry, TCreateEntry>::getHeadNode(bool* isLast) {
```
dgkimura: Why did this need to be changed to a raw pointer?

src/cppcache/src/MapSegment.cpp
```diff
@@ -19,18 +19,22 @@
...
-#include "SpinLock.hpp"
 #include "Utils.hpp"
 #include "ThinClientPoolDM.hpp"
 #include "ThinClientRegion.hpp"
 #include "TombstoneExpiryHandler.hpp"
 #include <ace/OS.h>
 #include "ace/Time_Value.h"
-using namespace apache::geode::client;
 
-#define _GF_GUARD_SEGMENT SpinLockGuard mapGuard(m_spinlock)
```
dgkimura: Yay, one less macro!

src/cppcache/src/MapSegment.hpp
```diff
@@ -41,6 +40,9 @@
 #include <ace/Versioned_Namespace.h>
 #include "TombstoneList.hpp"
 #include <unordered_map>
+
+#include "util/concurrent/spinlock_mutex.hpp"
```
dgkimura: Nit: Why are you using "?  And why is it placed down here?

src/cppcache/src/LRUList.cpp
```diff
@@ -15,12 +15,18 @@
...
  */
 #include "LRUList.hpp"
-#include "SpinLock.hpp"
+#include "util/concurrent/spinlock_mutex.hpp"
 
-using namespace apache::geode::client;
+#include <mutex>
 
-#define LOCK_HEAD SpinLockGuard headLockGuard(m_headLock)
-#define LOCK_TAIL SpinLockGuard tailLockGuard(m_tailLock)
```
PivotalSarge: Good to see the removal of more macros.

src/cppcache/src/LRUList.hpp
```diff
@@ -20,32 +20,32 @@
...
 namespace apache {
 namespace geode {
 namespace client {
+
 // Bit mask for recently used
-#define RECENTLY_USED_BITS 1ul
+#define RECENTLY_USED_BITS 1u
 // Bit mask for evicted
-#define EVICTED_BITS 2ul
+#define EVICTED_BITS 2u
```
PivotalSarge: Could these be const variables in an anonymous namespace instead?

src/cppcache/src/MapSegment.cpp
```diff
@@ -316,16 +320,16 @@ GfErrType MapSegment::remove(const CacheableKeyPtr& key, CacheablePtr& oldValue,
                              MapEntryImplPtr& me, int updateCount,
                              VersionTagPtr versionTag, bool afterRemote,
                              bool& isEntryFound) {
-  //  _GF_GUARD_SEGMENT;
+  //  std::lock_guard<spinlock_mutex> lk(m_spinlock);
```
PivotalSarge: Since the code is commented out, should it just be deleted?

src/cppcache/src/MapSegment.hpp
```diff
@@ -164,9 +166,9 @@ class CPPCACHE_EXPORT MapSegment {
     m_entryFactory->newMapEntry(key, newEntry);
     newEntry->setValueI(newValue);
     if (m_concurrencyChecksEnabled) {
-      if (versionTag != NULLPTR && versionTag.ptr() != NULL) {
+      if (versionTag != NULLPTR && versionTag.ptr() != nullptr) {
```
PivotalSarge: It looks like `nullptr` is being used, not `NULLPTR`, no?

src/cppcache/src/LRUEntriesMap.cpp
```diff
@@ -297,7 +295,7 @@ GfErrType LRUEntriesMap::put(const CacheableKeyPtr& key,
       }
     }
     // SpinLock& lock = segmentRPtr->getSpinLock();
-    // SpinLockGuard mapGuard( lock );
+    // std::lock_guard<spinlock_mutex> mapGuard( lock );
```
PivotalSarge: Remove commented out code?

src/cppcache/src/LRUEntriesMap.cpp
```diff
@@ -20,6 +20,9 @@
 #include "MapSegment.hpp"
 #include "CacheImpl.hpp"
 
+#include <mutex>
+#include "util/concurrent/spinlock_mutex.hpp"
```
PivotalSarge: Similar to a previous comment from @dgkimura.

src/tests/cpp/security/Security.cpp
```diff
@@ -41,16 +41,18 @@
...
 
-namespace FwkSecurity {
+#include <mutex>
+#include <util/concurrent/spinlock_mutex.hpp>
+
+namespace apache {
+namespace geode {
+namespace client {
+namespace testframework {
+namespace security {
```
PivotalSarge: Could these constants be moved into an anonymous namespace?

pivotal-jbarrett: Probably but I didn't want to change their current scope without understanding the test. I wanted to just correct the warning around `using namespace`.

src/cppcache/src/LRUEntriesMap.cpp
pivotal-jbarrett: Yes, thanks!

src/tests/cpp/security/Security.cpp
```diff
@@ -1014,12 +1016,15 @@ int32_t Security::doEntryOperations() {
                 reinterpret_cast<const unsigned char *>(valBuf),
                 static_cast<int32_t>(strlen(valBuf)));
             int32_t *val =
-                (int32_t *)(dynCast<CacheableBytesPtr>(tmpValue)->value());
+                const_cast<int32_t *>(reinterpret_cast<const int32_t *>(
+                    dynCast<CacheableBytesPtr>(tmpValue)->value()));
```
pivotal-jbarrett: `dynCast` is something Geode does to unwrap and rewrap the nasty `SharedPtr` so I am not going to touch that. The `const_cast<int32_t *>(reinterpret_cast<const int32_t *>` is the real change here to do the C++ version of `(int32_t*)` in the original. `dynCast` results in a `const X *`, so the obvious `reinterpret_cast<Y &*>` is actually a compiler error because `reinterpret_cast` can not drop modifiers like `const`. The `const_cast` then drops the modifier.

src/cppcache/src/MapSegment.hpp
pivotal-jbarrett: `"` includes are for "local" includes. If `spinlock_mutex` was an external library then `<` would be used.

If "down here" means why the sub directories, then the answer is to be consistent with the namespace. The utility of that is debatable.

src/cppcache/src/MapSegment.hpp
pivotal-jbarrett: `NULLPTR` is Geode constant from "empty" `SharedPtr`. 

`nullptr` is C++11 for "empty" pointer.

pivotal-jbarrett: The original `NULL` is C for `0` and was allowed in C++ as an implicit cast to `(void *)0`.

src/cppcache/integration-test/testSpinLock.cpp
```diff
@@ -20,48 +20,44 @@
...
-}  // namespace apache
+using apache::geode::util::concurrent::spinlock_mutex;
 
 DUNIT_TASK(s1p1, Basic)
   {
-    void* lock = apache::geode::client::testSpinLockCreate();
-    apache::geode::client::testSpinLockAcquire(lock);
-    apache::geode::client::testSpinLockRelease(lock);
+    spinlock_mutex s;
+    { std::lock_guard<spinlock_mutex> lk(s); }
```
pivotal-jbarrett: I agree but we chose Google C++ Style Guide, we should adopt our own. I would follow up on the thread on dev@geode discussing other issues with the guide.

src/cppcache/integration-test/testSpinLock.cpp
```diff
@@ -20,48 +20,44 @@
...
-    LOG("ThreadA: Acquired lock x.");
-    triggerM->release();
-    triggerA->acquire();
-    apache::geode::client::testSpinLockRelease(lock);
+    {
+      std::lock_guard<spinlock_mutex> lk(lock);
+      LOG("ThreadA: Acquired lock x.");
+      triggerM->release();
+      triggerA->acquire();
+    }
```
pivotal-jbarrett: Yes.

src/cppcache/src/CqQueryVsdStats.cpp
```diff
@@ -20,43 +20,38 @@
...
 #include <ace/Singleton.h>
 
-const char* cqStatsName = (const char*)"CqQueryStatistics";
-const char* cqStatsDesc = (const char*)"Statistics for this cq query";
+#include <mutex>
+
+#include "util/concurrent/spinlock_mutex.hpp"
+
+const char* cqStatsName = "CqQueryStatistics";
+const char* cqStatsDesc = "Statistics for this cq query";
```
pivotal-jbarrett: All I was doing here was removing the warning about redundant casts. I am not entirely familiar with the difference of `const` and `constexpr` at this location.

src/cppcache/src/CqQueryVsdStats.cpp
pivotal-jbarrett: `static_assert` is a compile time assertion. In this case the assertion can't be resolved at compile time since the variables and functions involved are not `constexpr`, templates or other compiler resolved things.

src/cppcache/src/LRUList.cpp
pivotal-jbarrett: Yes.

src/cppcache/src/LRUList.cpp
pivotal-jbarrett: Good questions... Adherence to Google C++ Style Guide on references vs. pointers. I think I did this before writing my questions about the practice on the dev@geode. I can back this change out.

For reference, the guide dictates that references must be const and should be used for in variables only. For out variable you should use pointer.

src/cppcache/src/MapSegment.cpp
pivotal-jbarrett: Death to macros!

src/cppcache/src/LRUList.hpp
pivotal-jbarrett: They are macros, but they could be converted to `const` in anonymous namespace or even in this namespace or `constexpr` on the class itself. I will look at what that looks like and maybe update it.

src/cppcache/src/MapSegment.cpp
pivotal-jbarrett: Yup

src/cppcache/src/LRUEntriesMap.cpp
pivotal-jbarrett: See response to @dgkimura .

src/cppcache/src/CqQueryVsdStats.cpp
pivotal-jbarrett: Good call here though. I can actually move them into the class using `constexpr`, which keeps them safely scoped to this class.

src/cppcache/integration-test/testSpinLock.cpp
echobravopapa: +1 to separate lines, @dgkimura you want to bump the thread and add this

PR #37, opened 2017-02-25
GEODE-2531: Replace HostAsm::atomic with std::atomic.

pivotal-jbarrett:
Focus on commit cf1ba85 since the others are from pull #36.

Replaces the HostAsm::atomic* functions and ACE::Atomic* templates with std::atomic<> template. Removes HostAsm and AtomicInc.


src/cppcache/include/geode/CacheStatistics.hpp
```diff
@@ -102,8 +102,8 @@ class CPPCACHE_EXPORT CacheStatistics : public SharedBase {
   virtual void setLastAccessedTime(uint32_t lat);
   virtual void setLastModifiedTime(uint32_t lmt);
 
-  volatile uint32_t m_lastAccessTime;
-  volatile uint32_t m_lastModifiedTime;
+  std::atomic<uint32_t> m_lastAccessTime;
+  std::atomic<uint32_t> m_lastModifiedTime;
```
PivotalSarge: std::atomic is pretty cool. It looks like the performance impact is negligible, right?

src/cppcache/include/geode/SharedBase.hpp
```diff
@@ -56,11 +57,12 @@ class CPPCACHE_EXPORT SharedBase {
 
  protected:
   inline SharedBase(bool noInit) {}
+  inline SharedBase(const SharedBase&) {}
 
   virtual ~SharedBase() {}
 
  private:
-  mutable volatile int32_t m_refCount;
+  std::atomic<int32_t> m_refCount;
```
PivotalSarge: Is the mutable unnecessary? Or does std::atomic have const operators that can be used to mutate the reference count?

src/cppcache/src/statistics/AtomicStatisticsImpl.cpp
```diff
@@ -330,7 +329,13 @@ double AtomicStatisticsImpl::_incDouble(int32_t offset, double delta) {
     throw IllegalArgumentException(s);
   }
 
-  return (doubleStorage[offset] += delta);
+  double expected = doubleStorage[offset];
+  double value;
+  do {
+    value = expected + delta;
+  } while (!doubleStorage[offset].compare_exchange_weak(expected, value));
```
PivotalSarge: If compare_exchange_weak() fails, should expected be refreshed with doubleStorage[offset]?

src/cppcache/include/geode/CacheStatistics.hpp
pivotal-jbarrett: The original code was using atomic operations already, just from multiple platform specific sources. std::atomic gets compiled into whatever is most performant for the platform. I would guess that this is actually faster than whatever we were doing on each platform since many were function calls into another library. This code will get compiled to atomic put opcodes on most platforms (with appropriate memory synchronization between caches).

src/cppcache/include/geode/SharedBase.hpp
pivotal-jbarrett: It didn't appear that the `mutable` attribute was necessary for operations on this value. This `SharedBase` thing is a little funky so I didn't poke too deep. The compiler did not find any issues where a `const` function or reference was making changes to this value. It may be worth more investigation.

src/cppcache/src/statistics/AtomicStatisticsImpl.cpp
pivotal-jbarrett: Yes. Some of the documentation for std::atomic suck out there and don't clearly state that expected is updated if the compare fails.

PR #39, opened 2017-02-27
Add rat checks to travis-ci build

metatype:


PivotalSarge:
Where is .ratignore? I don't see it.

echobravopapa:
@PivotalSarge .ratignore is in the base dir of geode-native

PivotalSarge:
Ah. `find` does not behave like I think it does with respect to finding dot files in the current directory.

echobravopapa:
I quoted the target, suspecting find would not like it otherwise...

PR #40, opened 2017-02-27
GEODE-2552: Remove NanoTimer

pivotal-jbarrett:
Please focus on commits e541e31 ... daf4234, others are artifacts of unpulled requests.


pivotal-jbarrett:
> Would adding const to variables like spentWorking and sleepDuration help the compiler generate more efficient code?

I doubt the compilers would care since it will know the scope is limited to the block and there are no other assignment operations. It does however provide more readable and safe coding style. I could make adjustments.

pivotal-jbarrett:
@PivotalSarge 
> Why the switch to ACE_OS::localtime() from localtime()?

`localtime` AKA POSIX `localtime` is not thread safe. `ACE_OS::localtime` wraps the platform specific `localtime_r` or `localtime_s` functions that are thread safe. Ultimately we want to replace these something like `boost::locale` until something is adopted into the `std`.

pivotal-jbarrett:
@PivotalSarge 
> Nice fix of the broken window of catching exceptions by value.

If you enable `clang-tidy` you will see plenty of them. I am cleaning up lots of these bad things along the way.

PR #44, opened 2017-03-02
GEODE-2578: Remove 64 KiB limit on query strings.

PivotalSarge:
- Remove artificial cap of 65535 for query string
  length by using 32 bits for the length of query
  strings in DataOutput::writeFullUTF().
- Rename parameter to TcrMessage::writeStringPart()
  whose name is misleading due to copy-and-paste.
- Add three unit tests for query-related messages:
  * testConstructorEXECUTECQ_MSG_TYPE
  * testConstructorWithGinormousQueryEXECUTECQ_MSG_TYPE
  * testConstructorEXECUTECQ_WITH_IR_MSG_TYPE


pivotal-jbarrett:
@PivotalSarge looks like some unit tests are failing.

PivotalSarge:
The unit tests failed on Linux but not other platforms. It turns out that DataOutput has some not-very-obvious (and apparently platform-specific) behavior around buffer management that results in the return values from getBufferLength() and getRemainingBufferLength() depending on the behavior of other unit tests that may precede them. Hence, the test of ginormous query strings was causing the return values of those methods to differ from what would be returned if that test was *not* run. So I reworked how the DataOutput cursor advancement unit tests work and created GEODE-2598 to understand and test how DataOutput's buffer management works.

pivotal-jbarrett:
@PivotalSarge can you rebase your branch off of develop. We backed out all those C++11 changes and now they are showing up in your pull.

PivotalSarge:
Roger that. Expect a new PR forthwith...

Sarge

PR #48, opened 2017-03-07
GEODE-2578: Remove 64 KiB limit on query strings.

PivotalSarge:
- Remove artificial cap of 65535 for query string
  length by using 32 bits for the length of query
  strings in DataOutput::writeFullUTF().
- Rename parameter to TcrMessage::writeStringPart()
  whose name is misleading due to copy-and-paste.
- Add three unit tests for query-related messages:
  * testConstructorEXECUTECQ_MSG_TYPE
  * testConstructorWithGinormousQueryEXECUTECQ_MSG_TYPE
  * testConstructorEXECUTECQ_WITH_IR_MSG_TYPE
- Fix the cursor advance tests to not depend on
  specific values for buffer length and not to
  test remaining buffer length.

### DIFF
--- a/src/cppcache/include/geode/DataOutput.hpp
+++ b/src/cppcache/include/geode/DataOutput.hpp
@@ -378,9 +378,8 @@ class CPPCACHE_EXPORT DataOutput {
      */
   inline void writeFullUTF(const char* value, uint32_t length = 0) {
     if (value != NULL) {
-      int32_t len = getEncodedLength(value, length);
-      uint16_t encodedLen = static_cast<uint16_t>(len > 0xFFFF ? 0xFFFF : len);
-      writeInt(static_cast<int32_t>(encodedLen));
+      int32_t encodedLen = getEncodedLength(value, length);
+      writeInt(encodedLen);
       ensureCapacity(encodedLen);
       write(static_cast<int8_t>(0));  // isObject = 0 BYTE_CODE
       uint8_t* end = m_buf + encodedLen;
--- a/src/cppcache/test/DataOutputTest.cpp
+++ b/src/cppcache/test/DataOutputTest.cpp
@@ -305,15 +305,9 @@ TEST_F(DataOutputTest, TestCursorAdvance) {
       "001B596F7520686164206D65206174206D65617420746F726E61646F2E",
       dataOutput.getByteArray());
 
-  EXPECT_EQ((2 + 27), dataOutput.getBufferLength());
-
-  // buffers are pre-allocated 8k and have 2 bytes to hold the data length
-  EXPECT_EQ(((8 * 1024) - (2 + 27)), dataOutput.getRemainingBufferLength());
-
+  const uint32_t originalLength = dataOutput.getBufferLength();
   dataOutput.advanceCursor(2);
-  EXPECT_EQ((2 + 27 + 2), dataOutput.getBufferLength());
-
-  EXPECT_EQ(((8 * 1024) - (2 + 27 + 2)), dataOutput.getRemainingBufferLength());
+  EXPECT_EQ((originalLength + 2), dataOutput.getBufferLength()) << "Correct length after advance";
 }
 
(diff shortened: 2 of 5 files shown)


src/cppcache/include/geode/DataOutput.hpp
```diff
@@ -378,9 +378,8 @@ class CPPCACHE_EXPORT DataOutput {
      */
   inline void writeFullUTF(const char* value, uint32_t length = 0) {
     if (value != NULL) {
-      int32_t len = getEncodedLength(value, length);
-      uint16_t encodedLen = static_cast<uint16_t>(len > 0xFFFF ? 0xFFFF : len);
```
echobravopapa: what was the corner case impact of this line?

src/cppcache/test/DataOutputTest.cpp
```diff
@@ -305,15 +305,9 @@ TEST_F(DataOutputTest, TestCursorAdvance) {
       "001B596F7520686164206D65206174206D65617420746F726E61646F2E",
       dataOutput.getByteArray());
 
-  EXPECT_EQ((2 + 27), dataOutput.getBufferLength());
-
-  // buffers are pre-allocated 8k and have 2 bytes to hold the data length
-  EXPECT_EQ(((8 * 1024) - (2 + 27)), dataOutput.getRemainingBufferLength());
```
echobravopapa: What led you to removing the call to getRemainingBufferLength()?

src/cppcache/test/TcrMessage_unittest.cpp
```diff
@@ -559,3 +560,66 @@ TEST_F(TcrMessageTest, testConstructorEXECUTE_FUNCTION) {
...
+  EXPECT_EQ(TcrMessage::EXECUTECQ_MSG_TYPE, testMessage.getMessageType());
+
+  EXPECT_MESSAGE_EQ(
+      "0000002A0000004000000005FFFFFFFF0000000009004578656375746543510000001800"
+      "73656C656374202A2066726F6D202F736F6D657768657265000000040000000001000000"
+      "010000000000010001",
+      testMessage);
+}
+
+TEST_F(TcrMessageTest, testConstructorWithGinormousQueryEXECUTECQ_MSG_TYPE) {
```
echobravopapa: I'm not certain that this query's magnitude qualifies for ginormity... maybe we just call it test*64KiBQuery* - lest we set the bar low for ginormous...

src/cppcache/include/geode/DataOutput.hpp
PivotalSarge: If the query string happened to be longer that 0xFFFF, it would be artificially truncated which often led to errors on the server side.

src/cppcache/test/DataOutputTest.cpp
PivotalSarge: getRemainingBufferLength() is the call that exhibited less-than-obvious behavior. In particular, it appears to be affected by some global state. Its return value was not predictable enough to be tested in that manner.

src/cppcache/test/TcrMessage_unittest.cpp
PivotalSarge: I'm unaware of the SI definition of ginormous...

src/cppcache/include/geode/DataOutput.hpp
```diff
@@ -378,9 +378,8 @@ class CPPCACHE_EXPORT DataOutput {
      */
   inline void writeFullUTF(const char* value, uint32_t length = 0) {
```
pivotal-jbarrett: If we are being sticklers and consistent this method has no business here. It is only used by the Message writeStringPart and should really be implemented there.

PivotalSarge: I agree and tried to move it to TcrMessage. However, its implementation depends on private members of DataOutput.

echobravopapa: ICKY, sounds like a separable refactoring chore...

PR #50, opened 2017-03-09
GEODE-2636: Switch to CMake variable for library name.

PivotalSarge:
- Parameterize the name of the library for which the quickstarts look.
- Fix warnings that obfuscate when the quickstarts can not find a library.


src/quickstart/cpp/PdxSerializer.cpp
```diff
@@ -165,15 +165,15 @@ int main(int argc, char** argv) {
     LOGINFO("Registered Person Query Objects");
 
     // Populate the Region with some Person objects.
-    Person* p1 = new Person("John", 1 /*ID*/, 23 /*age*/);
+    Person* p1 = new Person((char *)"John", 1 /*ID*/, 23 /*age*/);
```
echobravopapa: What led to adding this cast?

PivotalSarge: Those methods pass a string literal to a method whose first parameter is char *. An alternative would be to modify the type of the parameter but that was a more invasive change when the goal was to reduce the clutter in the log that got in the way of determining the root cause.

src/quickstart/cpp/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ elseif(UNIX)
     set(DYNAMIC_LIBRARY_PATH LD_LIBRARY_PATH=${NATIVECLIENT_DIR}/lib)
 endif()
 
+set(PRODUCT_LIB_NAME "apache-geode" CACHE STRING "Binary name")
```
metatype: Ah, makes sense.  I was wondering how that got set.

PR #52, opened 2017-03-13
GEODE-2513 Rename gfcpp.properties file to geode.properties - updated…

davebarnes97:
… references, moved and renamed files


karensmolermiller:
Please also change the headings in the file setting-properties/chapter-overview.html.

The 7 link names in file setting-properties/native-client-config.html do not match the headings at the top (or in section header) of the linked file.  I would like to see them match.  Same for 1 link name in file setting-properties/cache-server-config.html.

davebarnes97:
Good catch, @karensmolermiller  - corrections incorporated.

PR #54, opened 2017-03-14
GEODE-2657: Execute Region Function sends incorrect message format

dgkimura:
`TcrMessageExecuteRegionFunction` is missing a call to set `m_hasResults`. This causes the message response to be parsed synchronously and not asynchronously chunked. The synchronous parser does not support this message type and it barfs.

### DIFF
--- a/src/cppcache/src/TcrMessage.cpp
+++ b/src/cppcache/src/TcrMessage.cpp
@@ -2447,6 +2447,7 @@ TcrMessageExecuteRegionFunction::TcrMessageExecuteRegionFunction(
   m_tcdm = connectionDM;
   m_regionName = region == NULL ? "INVALID_REGION_NAME" : region->getFullPath();
   m_region = region;
+  m_hasResult = getResult;
 
   if (routingObj != NULLPTR && routingObj->size() == 1) {
     LOGDEBUG("setting up key");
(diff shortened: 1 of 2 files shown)


dgkimura:
@echobravopapa We're only human.

Our unittest didn't catch it originally because the output that's written to DataOutput in the constructor is the same with/without this variable set.  It's likely when the message is sent that client checks this variable and performs some magic.  I'm guessing our integration-tests that cover this aren't running.  Else we might have had a chance.

pivotal-jbarrett:
@dgkimura I would suggest getting a few of those integration tests updated to test this. They were marked flaky a long time ago when the tests ran through several iterations with different configurations. Now that we don't do that anymore it should be easy to stabilize the tests.

dgkimura:
@pivotal-jbarrett Okay. Fair enough. That's going to add some cycles to this pull-request.

pivotal-jbarrett:
@dgkimura It will give the CI something to chew on to prove that this message actually works with the server now. ;)

pivotal-jbarrett:
@dgkimura I will pull this and work on some of the tests.

pivotal-jbarrett:
This will get closed when pull #63 is committed.

PR #60, opened 2017-03-17
GEODE-2521 geode native docs: Document installation from source

davebarnes97:
Adds instructions for building from source.
Removes binary installation & uninstallation, which were artifacts from the proprietary code.


BUILDING.md
```diff
@@ -77,7 +77,7 @@ Due to limitations in CMake, the documentation must be built as a separate step
 
 ### Required Tools
 * [Visual Studio 2013](https://www.visualstudio.com) or newer
-* [Cygwin](https://www.cygwin.com/)
+* [Cygwin](https://www.cygwin.com/) (for the patch tool)
```
echobravopapa: @davebarnes97 I think we can remove reference to Cygwin, we get the patch tool from GNUWin now... if I am mistaken @dgkimura will set me straight

PR #61, opened 2017-03-17
GEODE-2687: test for ssl auth failure:

echobravopapa:
- Added testThinClientSSLAuthFail to QUICK list.
- removed un-needed test code.
- catching expected exception
- Add test for untrusted server cert


src/cppcache/integration-test/CacheHelper.cpp
```diff
@@ -1812,17 +1812,32 @@ std::string CacheHelper::generateGeodeProperties(const std::string& path,
...
-    msg += "cluster-ssl-require-authentication=true\n";
-    msg += "cluster-ssl-ciphers=TLS_RSA_WITH_AES_128_CBC_SHA\n";
-    msg += "cluster-ssl-keystore-type=jks\n";
-    msg += "cluster-ssl-keystore=" + keystore + "/server_keystore.jks\n";
-    msg += "cluster-ssl-keystore-password=gemstone\n";
-    msg += "cluster-ssl-truststore=" + keystore + "/server_truststore.jks\n";
-    msg += "cluster-ssl-truststore-password=gemstone\n";
-    msg += "security-username=xxxx\n";
-    msg += "security-userPassword=yyyy \n";
+    if (untrustedCert){
```
PivotalSarge: Much like keystore is a parameter to this method, could this if-then-else statement be collapsed to a single set of concatenations that use a (possibly empty) prefix string, e.g., "untrusted_"?

pivotal-jbarrett:
I will go ahead and merge.

PR #63, opened 2017-03-20
GEODE-2657 - Fixes and tests Function Execution.

pivotal-jbarrett:
Please make a quick review.

Fixes all the integration tests around Function Execution and adds test of AppDomain Function Execution.

Test testThinClientPoolExecuteFunctionDisableChunkHandlerThread is split out of testThinClientPoolExecuteFunction and marked OMITTED until GEODE-2691 is addressed.


src/cppcache/integration-test/testThinClientExecuteFunctionPrSHOP.cpp
```diff
@@ -563,7 +563,7 @@ DUNIT_TASK_DEFINITION(CLIENT1, Client1OpTest2)
       LOGINFO("FETimeOut begin onRegion");
       ExecutionPtr RexecutionPtr = FunctionService::onRegion(regPtr0);
       CacheableVectorPtr fe =
-          RexecutionPtr->withArgs(CacheableInt32::create(5000))
+          RexecutionPtr->withArgs(CacheableInt32::create(5000 * 1000))
```
PivotalSarge: Would std::chrono::duration help avoid the seconds-for-milliseconds kinds of problems in future?

src/clicache/src/CacheFactory.cpp
```diff
@@ -84,6 +85,12 @@ namespace Apache
 					//TODO::split
           SafeConvertClassGeneric::SetAppDomainEnabled(appDomainEnable);
 
+          if (appDomainEnable)
+          {
+            // Register managed AppDomain context with unmanaged.
+            apache::geode::client::createAppDomainContext = &Apache::Geode::Client::createAppDomainContext;
```
PivotalSarge: Does moving the registration of the managed app domain context to cache factory from distributed system make it available sooner?

src/cppcache/integration-test/CMakeLists.txt
```diff
@@ -169,8 +169,6 @@ set_property(TEST testThinClientPartitionResolver PROPERTY LABELS FLAKY)
 set_property(TEST testThinClientPdxDeltaWithNotification PROPERTY LABELS FLAKY)
 set_property(TEST testThinClientPdxInstance PROPERTY LABELS FLAKY)
 set_property(TEST testThinClientPdxTests PROPERTY LABELS FLAKY)
```
echobravopapa: 5 more live tests, that is nice

src/cppcache/integration-test/testThinClientExecuteFunctionPrSHOP.cpp
echobravopapa: Should we be marking these areas for posterity and conversion to only using milli?

src/cppcache/integration-test/testThinClientPoolExecuteFunction.cpp
```diff
@@ -1158,9 +1160,10 @@ DUNIT_TASK_DEFINITION(CLIENT1, Client2OpTest)
 
       LOGINFO("FETimeOut begin onServer");
       ExecutionPtr serverExc = FunctionService::onServer(getHelper()->cachePtr);
-      CacheableVectorPtr vec = serverExc->withArgs(CacheableInt32::create(5000))
-                                   ->execute(FETimeOut, 5000)
-                                   ->getResult();
+      CacheableVectorPtr vec =
+          serverExc->withArgs(CacheableInt32::create(5000 * 1000))
```
echobravopapa: might consider making a constant or two for these literals... in case someone is maintaining the test later on and needs/wants to adjust times used

src/cppcache/integration-test/testThinClientPoolExecuteFunction.cpp
```diff
@@ -1506,48 +1353,10 @@ void runFunctionExecution(bool isEndpoint) {
   CALL_TASK(StopC1);
   CALL_TASK(CloseServers);
   CALL_TASK(CloseLocator1);
-
```
echobravopapa: What were these big (removed) chunks testing?

- out of scope?
- redundant?

src/clicache/integration-test/ThinClientFunctionExecutionTestsN.cs
```diff
@@ -959,7 +959,7 @@ public void ExecuteFETimeOut()
         region[i] = "VALUE--" + i;
       }
 
-      object args = 5000;
+      object args = 5000 * 1000;
```
echobravopapa: maybe a constant..

src/cppcache/integration-test/testThinClientExecuteFunctionPrSHOP.cpp
pivotal-jbarrett: Yes. Ultimately we should consider removing the continuity issue between the client API and the server. The client is taking the timeout as seconds but sending to the server as milliseconds. Ultimately taking a chrono::duration would make the API more clear. The test would still need to be very explicit on the use of milliseconds are the argument to the function.

src/clicache/src/CacheFactory.cpp
pivotal-jbarrett: No, it just looked better here since the surrounding code was doing things with the appDomainEnabled flag. Where it came from had no other AppDomain code and didn't check the flag so it was always "enabled". Now it will be enable only when the `appdomain-enabled` property is set.

pivotal-jbarrett: I think we may want to just abandon the `appdomain-enabled` property and always enable AppDomain support when using the .NET client. I don't know that there is any good reason for running disabled since .NET uses AppDomains heavily (prior to .NET Core).

src/clicache/integration-test/ThinClientAppDomainFunctionExecutionTests.cs
```diff
@@ -0,0 +1,282 @@
...
+    {
+      return m_endResultCount;
+    }
+    public MyAppDomainResultCollector()
+    {
+      m_results = new List<TResult>();
+    }
+    public void AddResult(TResult result)
+    {
+      Util.Log("hitesh in MyAppDomainResultCollector " + result + " :  " + result.GetType());
```
mmartell: What's the significance of "hitesh in"?

pivotal-jbarrett: Doh! thanks!

src/clicache/integration-test/ThinClientFunctionExecutionTestsN.cs
pivotal-jbarrett: I felt it was important to call out the disconnect between units here rather than hide it in a constant.

src/clicache/src/CacheFactory.cpp
echobravopapa: +1 make it a built default, less property juggling for the app devs

src/cppcache/integration-test/testThinClientPoolExecuteFunction.cpp
pivotal-jbarrett: Like I said in the review request. This test was split out into tho two tests. The test had two different tests crammed into one file. The big delete in this file is all the crap that became the second test.

src/cppcache/integration-test/testThinClientPoolExecuteFunction.cpp
pivotal-jbarrett: See comment on same request below.

src/clicache/integration-test/ThinClientFunctionExecutionTestsN.cs
echobravopapa: fair choice

PR #70, opened 2017-03-22
GEODE-2470: Fix for Solaris regex short comings.

mmartell:


src/cppcache/integration-test/CacheHelper.cpp
```diff
@@ -1376,6 +1376,41 @@ void CacheHelper::createDuplicateXMLFile(std::string& originalFile,
...
+    std::ofstream out(outFile, std::ios::out);
+    out << contents;
+    out.close();
+  }
+}
+
+void CacheHelper::replaceInPlace(std::string& searchStr, const std::string& matchStr,
+                                 const std::string& replaceStr) {
+    size_t pos = 0;
+    while ((pos = searchStr.find(matchStr, pos)) != std::string::npos) {
```
echobravopapa: @mmartell I think there is an issue with this replacement code... 
I'm not convinced we need the loop and my experiments are showing that a searchStr like so:
"look do leap be sheep" where the matchStr is "leap" and replaceStr is "lose" you will end up with "look do lose"

src/cppcache/integration-test/CacheHelper.cpp
```diff
@@ -1376,6 +1376,41 @@ void CacheHelper::createDuplicateXMLFile(std::string& originalFile,
...
+      searchStr.replace(pos, matchStr.length(), replaceStr);
```
echobravopapa: There we go... second arg to .replace() makes all the difference

mmartell:
This PR has been merged.

PR #82, opened 2017-03-30
GEODE-2736: Fixed orphaned worker threads

mhansonp:
- Force timeouts to be handled like no connection
- Added test

### DIFF
--- a/src/cppcache/integration-test/testThinClientPoolExecuteFunctionThrowsException.cpp
+++ b/src/cppcache/integration-test/testThinClientPoolExecuteFunctionThrowsException.cpp
@@ -1,0 +1,363 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#define ROOT_NAME "testThinClientPoolExecuteFunctionThrowsException"
+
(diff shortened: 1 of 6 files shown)


src/cppcache/integration-test/testThinClientPoolExecuteFunctionThrowsException.cpp
```diff
@@ -0,0 +1,362 @@
...
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+    
+    try {
+        CacheableVectorPtr fil = CacheableVector::create();
+        fil->push_back(CacheableInt32::create(1));
+        ExecutionPtr exe = FunctionService::onRegion(regPtr0);
+        
+        LOGINFO("Executing the exception test it is expected to throw.");
+        CacheableVectorPtr executeFunctionResult3 =
+        funcExec->withArgs(arrList)->withFilter(filter)->execute("ThinClientRegionExceptionTest", 15)->getResult();
```
PivotalSarge: Is the expectation that this method invocation will throw an exception? If so, it probably should be followed in the try block by a FAIL() to catch the cases where it doesn't.

mhansonp: It will always throw and the fact that it does is success. The reason this is good is that it means that the cleaning up of the worker threads in the executeFunctionSH followed its cleanup path. I don't have a way to directly infer that all the workers were cleaned up directly, but the result of that code path is that when a fatal or unknown exception is thrown by the server, that we cleanup properly. If the exception comes through, then we know the exception that is to be thrown at the end of cleanup happened.

echobravopapa: @mhansonp have a look at testThinClientSSLAuthFail for what @PivotalSarge is asking for regarding the FAIL()

src/cppcache/integration-test/testThinClientPoolExecuteFunctionThrowsException.cpp
```diff
@@ -303,6 +303,7 @@ DUNIT_TASK_DEFINITION(CLIENT1, Client1OpTest)
         LOGINFO("Executing the exception test it is expected to throw.");
         CacheableVectorPtr executeFunctionResult3 =
         funcExec->withArgs(arrList)->withFilter(filter)->execute("ThinClientRegionExceptionTest", 15)->getResult();
+        FAIL("Failed to throw expected exception.");
```
echobravopapa: Perfect. @mhansonp may I ask you to please squash your commits as well...? TIA

PR #83, opened 2017-03-31
GEODE-2691: Fix function execution attributes mismatch

dgkimura:
Fixes function execution invocation which previously passed in attributes that
function execute would compare to the server's view of the registered function
attributes. Since the previously passed in attributes were deprecated there is
nothing to compare. Instead we should accept the attributes that the server
returns.

Testing: Function execution tests pass

### DIFF
--- a/src/cppcache/src/ExecutionImpl.cpp
+++ b/src/cppcache/src/ExecutionImpl.cpp
@@ -72,11 +72,6 @@ std::vector<int8_t>* ExecutionImpl::getFunctionAttributes(const char* func) {
   return NULL;
 }
 
-ResultCollectorPtr ExecutionImpl::execute(const char* func, uint32_t timeout) {
-  LOGDEBUG("ExecutionImpl::execute1: ");
-  return execute(func, timeout, true);
-}
-
 ResultCollectorPtr ExecutionImpl::execute(CacheableVectorPtr& routingObj,
                                           CacheablePtr& args,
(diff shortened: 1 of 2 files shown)


src/cppcache/src/ExecutionImpl.cpp
```diff
@@ -100,33 +94,31 @@ ResultCollectorPtr ExecutionImpl::execute(const char* fn, uint32_t timeout,
   bool serverIsHA = false;
   bool serverOptimizeForWrite = false;
 
-  if (verifyFuncArgs) {
-    std::vector<int8_t>* attr = getFunctionAttributes(fn);
-    {
+  std::vector<int8_t>* attr = getFunctionAttributes(fn);
+  {
```
PivotalSarge: What is the purpose of this curly-brace block? Is there an RAII instance in there?

echobravopapa: looks like it is an artifact from the previous if statement and can be removed

echobravopapa: I take that back... @dgkimura what is with the curly braced block?

dgkimura: The curly-brace block was not introduced as part of this change, so your guess is as good as mine. :)

echobravopapa:
@dgkimura merged, please close

PR #84, opened 2017-03-31
GEODE-2513 Unbranding docs: Added Geode logos to doc headers.

davebarnes97:
Added Geode logos to the documentation's top frame and at the top of individual topics. Patterned after the Geode website, with a white-on-black logo in the browser frame and a color-on-white log in the page body.


davebarnes97:
docs-branding-os-client

PR #86, opened 2017-04-04
Feature/geode 2513 security props

davebarnes97:


davebarnes97:
Whoa! Thought I fixed that...

davebarnes97:
I _did_ fix it. Not sure why the old misspelling was visible to you,
@pivotalsarge.

PR #103, opened 2017-06-23
GEODE-2525: Removes MersenneTwister sources.

pivotal-jbarrett:
Can I get a quick review before I merge.


src/tests/cpp/fwklib/GsRandom.cpp
```diff
@@ -18,60 +18,16 @@
...
-  if (seed != seedUsed) {
-    if (seed != -1) {
-      singleton->gen.seed(seed);
-    } else {
-      singleton->gen.seed();
-    }
-    seedUsed = seed;
-  }
+GsRandom &GsRandom::getInstance() {
+  // C++11 initializes statics threads safe
```
dgkimura: Nit, I don't this this comment adds much value.

pivotal-jbarrett: Yeah probably not. I can kill it.

PR #104, opened 2017-06-28
GEODE-3143 : close socket before deteting

fdaniel7:
Closing socket before deleting it.


pivotal-jbarrett:
It looks like your reverts were to undo changes to .gitignore. I'd suggest not building from within the source tree at all. If you do there is an ignore already for "build" that you can use as your build directory. CMake promotes building outside the source directory, you can generate your project files from any location.

fdaniel7:
Hi I'm still new with github, it is indeed a small change. On next contributions I'll pay more attention.

echobravopapa:
@fdaniel7  What is the GEODE ticket you are working on?

fdaniel7:
There's no ticket for this as far as I know. I'm fixing this as part of reviewing geode client code

pivotal-jbarrett:
Please open a JIRA and include it in your commit message.

GEODE-1234: blah blah

fdaniel7:
When connect() is invoked from one of TcpConn::TcpConn the TcpConn object is not created since the exception is thrown from the constructor, so ~TcpConn will not be invoked.

echobravopapa:
@fdaniel7 please close this PR, thx

PR #105, opened 2017-07-02
GEODE-2891 connect-timeout violation in C++ Native Client

gregt5259:
The fix enables to interpret the time measure unit in handshake as milliseconds rather than seconds.

### DIFF
--- a/src/cppcache/src/TcrConnection.cpp
+++ b/src/cppcache/src/TcrConnection.cpp
@@ -318,7 +318,9 @@ bool TcrConnection::InitTcrConnection(
   LOGFINE("Attempting handshake with endpoint %s for %s%s connection", endpoint,
           isClientNotification ? (isSecondary ? "secondary " : "primary ") : "",
           isClientNotification ? "subscription" : "client");
-  ConnErrType error = sendData(data, msgLengh, connectTimeout, false);
+  // GT GEODE-2891
+  //ConnErrType error = sendData( data, msgLengh, connectTimeout, false );
+  ConnErrType error = sendData( data, msgLengh, connectTimeout, false, TcrMessage::HANDSHAKE );
 
   if (error == CONN_NOERR) {
     CacheableBytesPtr acceptanceCode = readHandshakeData(1, connectTimeout);
@@ -383,7 +385,9 @@ bool TcrConnection::InitTcrConnection(
       uint32_t credLen;
       char* credData = (char*)sendCreds.getBuffer(&credLen);
       // send the encrypted bytes and check the response
-      error = sendData(credData, credLen, connectTimeout, false);
+	  // GT GEODE-2891
+      //error = sendData(credData, credLen, connectTimeout, false);
+      error = sendData(credData, credLen, connectTimeout, false, TcrMessage::HANDSHAKE);
 
       if (error == CONN_NOERR) {
         acceptanceCode = readHandshakeData(1, connectTimeout);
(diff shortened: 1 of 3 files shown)


src/cppcache/src/TcrConnection.cpp
```diff
@@ -318,7 +318,9 @@ bool TcrConnection::InitTcrConnection(
   LOGFINE("Attempting handshake with endpoint %s for %s%s connection", endpoint,
           isClientNotification ? (isSecondary ? "secondary " : "primary ") : "",
           isClientNotification ? "subscription" : "client");
-  ConnErrType error = sendData(data, msgLengh, connectTimeout, false);
+  // GT GEODE-2891
```
pivotal-jbarrett: Do not include comments with your name or the ticket number.

src/cppcache/src/TcrConnection.cpp
```diff
@@ -318,7 +318,9 @@ bool TcrConnection::InitTcrConnection(
...
+  //ConnErrType error = sendData( data, msgLengh, connectTimeout, false );
```
pivotal-jbarrett: Do not leave commented out sources, this is what revision control is for.

src/cppcache/src/TcrConnection.cpp
```diff
@@ -703,15 +717,21 @@ inline ConnErrType TcrConnection::sendData(uint32_t& timeSpent,
...
-    }
+    	notPublicApiWithTimeout == TcrMessage::EXECUTE_REGION_FUNCTION_SINGLE_HOP ||
+		// GT GEODE-2891
+		notPublicApiWithTimeout == TcrMessage::HANDSHAKE)
+	{
+	    // then app has set timeout in millis, change it to microSeconds
+	    sendTimeoutSec = sendTimeoutSec * 1000;
+	    isPublicApiTimeout = true;
+	    LOGDEBUG("sendData2 %d ", sendTimeoutSec);
+	}
```
pivotal-jbarrett: Formatting does not conform to  [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html).

src/cppcache/src/TcrMessage.hpp
```diff
@@ -171,7 +173,8 @@ class CPPCACHE_EXPORT TcrMessage {
     GET_DURABLE_CQS_DATA_ERROR = 106,
     GET_ALL_WITH_CALLBACK = 107,
     PUT_ALL_WITH_CALLBACK = 108,
-    REMOVE_ALL = 109
+	REMOVE_ALL = 109,
+	HANDSHAKE = 110
```
pivotal-jbarrett: These numbers correspond to protocol message numbers on the server. We can just add one here and expect it not cause issues later.

src/cppcache/src/TcrMessage.hpp
```diff
@@ -44,6 +44,8 @@
 #include <map>
 #include <vector>
 
+//
```
pivotal-jbarrett: Clean this up.

gregt5259:
I suggest to fix the issue by defining handshake pseudo message within the range probably defined for such pseudo messages by original design i.e.:

  typedef enum {
    /* Server couldn't read message; handle it like a server side
       exception that needs retries */
    HANDSHAKE = -3
    NOT_PUBLIC_API_WITH_TIMEOUT = -2,

WDYT?

Thanks,
Dr. Gregory Turovets

"…We're all mad here. I'm mad. You're mad."
"How do you know I'm mad?" said Alice.
"You must be," said the Cat, "or you wouldn't have come here."
Alice's Adventures in Wonderland by Lewis Carroll<http://www.livelib.ru/author/157108>.

pivotal-jbarrett:
> I suggest to fix the issue by defining handshake pseudo message within the range probably defined for such pseudo messages by original design i.e.:
>  typedef enum {
>    /* Server couldn't read message; handle it like a server side
>       exception that needs retries */
>    HANDSHAKE = -3
>    NOT_PUBLIC_API_WITH_TIMEOUT = -2,

I think that would certainly work better but I think ultimately the fix is to remove the ambiguity between the C++ API and the Server API by strongly typing the duration values as addressed in the mentioned tickets. All duration values can be cast to the internal API supported unit without any ambiguity. After which there is no need to create this pseudo message to change the unit of the ambiguous int value.

pivotal-jbarrett:
@gregt5259 please close this pull request since you opened #106.

gregt5259:
Closed by Ernie Burghardt request as duplicated #106

PR #107, opened 2017-07-06
GEODE-3019: Refactor Struct class

echobravopapa:
- Using std::string now

- this also addresses GEODE-3020, the map now used standard types

### DIFF
--- a/src/cppcache/include/geode/Struct.hpp
+++ b/src/cppcache/include/geode/Struct.hpp
@@ -76,7 +76,7 @@ class CPPCACHE_EXPORT Struct : public Serializable {
    * @returns A smart pointer to the field value.
    * @throws IllegalArgumentException if the field name is not found.
    */
-  const SerializablePtr operator[](const char* fieldName) const;
+  const SerializablePtr operator[](const std::string& fieldName) const;
 
   /**
    * Get the parent StructSet of this Struct.
@@ -137,8 +137,9 @@ class CPPCACHE_EXPORT Struct : public Serializable {
   /**
    * Returns the name of the field corresponding to the index number in the
    * Struct
+   * @throws std::out_of_range if index is not found
    */
-  virtual const char* getFieldName(const int32_t index) const;
+  virtual const std::string& getFieldName(const int32_t index) const;
 
   /**
    * always returns 0
@@ -152,9 +153,7 @@ class CPPCACHE_EXPORT Struct : public Serializable {
 
   Struct();
 
-  typedef std::unordered_map<CacheableStringPtr, CacheableInt32Ptr,
-                             dereference_hash<CacheableStringPtr>,
-                             dereference_equal_to<CacheableStringPtr>>
+  typedef std::unordered_map<std::string, int32_t>
       FieldNames;
   FieldNames m_fieldNames;
   std::vector<SerializablePtr> m_fieldValues;
--- a/src/cppcache/integration-test/testThinClientRemoteQuerySS.cpp
+++ b/src/cppcache/integration-test/testThinClientRemoteQuerySS.cpp
@@ -48,6 +48,7 @@ const char* locHostPort =
 bool isPoolConfig = false;  // To track if pool case is running
 const char* qRegionNames[] = {"Portfolios", "Positions", "Portfolios2",
                               "Portfolios3"};
+
 const char* checkNullString(const char* str) {
   return ((str == nullptr) ? "(null)" : str);
 }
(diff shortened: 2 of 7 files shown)


src/cppcache/include/geode/Struct.hpp
```diff
@@ -138,7 +138,7 @@ class CPPCACHE_EXPORT Struct : public Serializable {
    * Returns the name of the field corresponding to the index number in the
    * Struct
    */
-  virtual const char* getFieldName(const int32_t index) const;
+  virtual const std::string* getFieldName(const int32_t index) const;
```
dgkimura: Any reason why you're returning `std::string*`  instead of `std::string` or `std::string&`?

src/cppcache/include/geode/Struct.hpp
```diff
@@ -152,9 +152,7 @@ class CPPCACHE_EXPORT Struct : public Serializable {
 
   Struct();
 
-  typedef std::unordered_map<CacheableStringPtr, CacheableInt32Ptr,
-                             dereference_hash<CacheableStringPtr>,
-                             dereference_equal_to<CacheableStringPtr>>
```
dgkimura: Nice!

src/cppcache/integration-test/testThinClientRemoteQuerySS.cpp
```diff
@@ -56,59 +57,63 @@ const wchar_t* checkNullString(const wchar_t* str) {
...
 
+std::string checkNullString(const std::string* str) {
+    return ((str == nullptr) ? "(null)" : *str);
+}
+
 void _printFields(CacheablePtr field, Struct* ssptr, int32_t& fields) {
   if (auto portfolio = std::dynamic_pointer_cast<Portfolio>(field)) {
     printf("   pulled %s :- ID %d, pkid %s\n",
-           checkNullString(ssptr->getFieldName(fields)), portfolio->getID(),
+           checkNullString(ssptr->getFieldName(fields)).c_str(), portfolio->getID(),
```
dgkimura: If we can make `getFieldName` return `std::string` then I'd think we can just `+` concatenate the strings without having to do `checkNullString` and  `c_str()`.

src/cppcache/include/geode/Struct.hpp
echobravopapa: Yes.  To remain compatible with the existing usage of the method were `nullptr` is returned.

src/cppcache/integration-test/testThinClientRemoteQuerySS.cpp
echobravopapa: That is a good idea.  There are 3 versions of checkNullString and it's test code so we minimized changes here in the tests...

src/cppcache/include/geode/Struct.hpp
pivotal-jbarrett: I am too not a huge fan of returning `std::string*` but the current behavior is to return `nullptr` if the field does not exist at that index. 

@dgkimura would you be happier if this method threw [`out_of_range`](http://en.cppreference.com/w/cpp/error/out_of_range)? I think I might like that better than null.

src/cppcache/integration-test/testThinClientRemoteQuerySS.cpp
pivotal-jbarrett: I think I would just rather see the tests not log so much crap and if they do that they use a logger with stream functionality. 
```
if (log.debug) {
  log.debug << "Doing some logging. somevar=" << someVar << ", someOtherVar=" << someOtherVar;
}
```
😒

src/cppcache/include/geode/Struct.hpp
dgkimura: Yeah, if we consider when field does not exist at index a failure case, then I agree that throwing is better than returning `nullptr` and expecting someone else to handle it or blow up later.

echobravopapa:
@dgkimura @pivotal-jbarrett I squashed the latest so you won't get notified, which is a bummer, but please have a look when you have a minute.

src/cppcache/src/Struct.cpp
```diff
@@ -96,16 +99,16 @@ Serializable* Struct::fromData(DataInput& input) {
...
     return m_parent->getFieldName(index);
   } else {
     for (const auto& iter : m_fieldNames) {
-      if ((iter.second)->value() == index) return iter.first->asChar();
+      if (iter.second == index) return (iter.first);
     }
   }
 
-  return nullptr;
+  return std::string("");
```
pivotal-jbarrett: What about throwing [`std::out_of_range`](http://en.cppreference.com/w/cpp/error/out_of_range). I don't like the idea of checking for empty string.

src/cppcache/src/StructSetImpl.cpp
```diff
@@ -78,13 +78,11 @@ int32_t StructSetImpl::getFieldIndex(const char* fieldname) {
...
-  for (std::map<std::string, int32_t>::iterator iter =
-           m_fieldNameIndexMap.begin();
-       iter != m_fieldNameIndexMap.end(); ++iter) {
-    if (iter->second == index) return iter->first.c_str();
+const std::string& StructSetImpl::getFieldName(int32_t index) {
+  for (const auto& iter : m_fieldNameIndexMap) {
+    if (iter.second == index) return (iter.first);
   }
-  return nullptr;
+  return std::string("");
```
pivotal-jbarrett: `std::out_of_range`?

echobravopapa:
@dgkimura @pivotal-jbarrett added unit tests and throwing an exception for out of range now

src/cppcache/src/StructSetImpl.cpp
```diff
@@ -66,8 +66,8 @@ SelectResultsIterator StructSetImpl::getIterator() {
   return SelectResultsIterator(m_structVector, shared_from_this());
 }
 
-int32_t StructSetImpl::getFieldIndex(const char* fieldname) {
-  std::map<std::string, int32_t>::iterator iter =
+int32_t StructSetImpl::getFieldIndex(const std::string& fieldname) {
```
pivotal-jbarrett: should this method be a const method?

src/cppcache/src/StructSetImpl.cpp
```diff
@@ -66,8 +66,8 @@ SelectResultsIterator StructSetImpl::getIterator() {
...
+  const auto& iter =
       m_fieldNameIndexMap.find(fieldname);
   if (iter != m_fieldNameIndexMap.end()) {
     return iter->second;
```
pivotal-jbarrett: We should clear out commented sources.

src/cppcache/src/StructSetImpl.cpp
```diff
@@ -78,13 +78,12 @@ int32_t StructSetImpl::getFieldIndex(const char* fieldname) {
...
-           m_fieldNameIndexMap.begin();
-       iter != m_fieldNameIndexMap.end(); ++iter) {
-    if (iter->second == index) return iter->first.c_str();
+const std::string& StructSetImpl::getFieldName(int32_t index) {
+  for (const auto& iter : m_fieldNameIndexMap) {
+    if (iter.second == index) return (iter.first);
   }
-  return nullptr;
+
+  throw OutOfRangeException("Struct: fieldName not found.");
```
pivotal-jbarrett: why not `std::out_of_range`?

echobravopapa: we can do that, I was just using our existing exception

src/cppcache/src/StructSetImpl.cpp
echobravopapa: will do

src/cppcache/src/StructSetImpl.cpp
echobravopapa: I'll have a look

src/cppcache/src/StructSetImpl.cpp
pivotal-jbarrett: I am more keen myself on using `std::exception`s rather than our own.

Thoughts?

echobravopapa: I'm changing both of them, it is the direction the code is headed.

echobravopapa:
@dgkimura @pivotal-jbarrett I think this is all ready for final inspection

src/cppcache/src/StructSetImpl.cpp
```diff
@@ -66,25 +66,22 @@ SelectResultsIterator StructSetImpl::getIterator() {
...
-           m_fieldNameIndexMap.begin();
-       iter != m_fieldNameIndexMap.end(); ++iter) {
-    if (iter->second == index) return iter->first.c_str();
+const std::string& StructSetImpl::getFieldName(int32_t index) {
+  for (const auto& iter : m_fieldNameIndexMap) {
+    if (iter.second == index) return (iter.first);
   }
-  return nullptr;
+
+  throw std::out_of_range("Struct: fieldName not found.");
```
dgkimura: Would it add any value having the index in the error message?

src/cppcache/test/StructSetTest.cpp
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+#include <gtest/gtest.h>
+#include <stdexcept>
+
+#include <StructSetImpl.hpp>
+
+using namespace apache::geode::client;
+
+TEST(StructSetTest, Basic) {
```
dgkimura: Woot more unittests!

PR #110, opened 2017-07-07
GEODE-3135 Update Geode Client docs OpenSSL version

karensmolermiller:
@pivotal-jbarrett @joeymcallister @davebarnes97 
Please review.  I have reduced the locations which call out which version of OpenSSL to use, and reworded to specify version 1.0.2 (no references to 'or a later version').  I also eliminated instructions for installing OpenSSL; they provide those, and we should not.

### DIFF
--- a/docs/geode-native-docs/introduction/client-configurations.html.md.erb
+++ b/docs/geode-native-docs/introduction/client-configurations.html.md.erb
@@ -157,7 +157,7 @@ The following libraries are external dependencies of the client library, `libgfc
 
 If you plan on using SSL in your Geode client and server deployment, you will need to download and install OpenSSL.
 
-The client requires OpenSSL 1.0.1t or later. For Windows platforms, you can use either the regular or the OpenSSL "Light" version.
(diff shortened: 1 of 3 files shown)


docs/geode-native-docs/introduction/client-configurations.html.md.erb
```diff
@@ -157,7 +157,7 @@ The following libraries are external dependencies of the client library, `libgfc
 
 If you plan on using SSL in your Geode client and server deployment, you will need to download and install OpenSSL.
 
-The client requires OpenSSL 1.0.1t or later. For Windows platforms, you can use either the regular or the OpenSSL "Light" version.
+The client requires OpenSSL version 1.0.2. For Windows platforms, you can use either the regular or the OpenSSL "Light" version.
```
pivotal-jbarrett: I still feel like it would be better to save that "For Windows..." statement for a common area about where to get OpenSSL.

pivotal-jbarrett: Maybe just link to the sslclientserver.html.md.erb file below.

PR #114, opened 2017-08-07
GEODE-2729: Remove global variables

dgkimura:
- Remove CacheImpl::getInstance calls
- Converted PdxTypeRegistry clear to non-static method
- Adding Type Registry to cpp code as a public object.
- Convert static PoolManager method calls into instance method calls
- Remove PoolFactory/PoolManager global connectionPool singleton
- Allows multiple PoolFactories per Cache
- Remove default pool logic from Cache into PoolManager
- Add mutli-cache integration test

All c++ and .NET tests (unit and integration) passed on Windows, Linux, MacOS, and Solaris X86.


pivotal-jbarrett:
Please fix up commit messages

pivotal-jbarrett:
This was merged without auto-close comment.

PR #115, opened 2017-08-11
GEODE-3165: Reogranized sources relative to the root for better CMake

pivotal-jbarrett:
… IDE integration.

- Moved src/docs to docs/api.
- Moved src/* to root.
- Fixup paths in CMake files.


CMakeLists.txt
```diff
@@ -242,7 +242,7 @@ add_subdirectory(dhimpl)
 add_subdirectory(sqliteimpl)
 add_subdirectory(tests)
 add_subdirectory(templates/security)
-add_subdirectory(docs)
+add_subdirectory(docs/api)
```
dgkimura: What was the motivation to move the doc's `CMakeLists.txt` into api directory?

pivotal-jbarrett: Could not move /src/docs to /docs since /docs already exists. /src/docs are just the config to generate api docs. So /docs/api made sense.

pivotal-jbarrett:
@dgkimura which CMake find modules are you referring to?

CMakeLists.txt
dgkimura: Ah, I see. So we have 2 docs? That's a little confusing..

dgkimura:
@pivotal-jbarrett I was referring to `FindNativeClient.cmake` and `FindNativeClientCPPCache.cmake`.

CMakeLists.txt
pivotal-jbarrett: There are API docs and the manual.

pivotal-jbarrett:
@dgkimura Those files actually need to be redone in a way more closely aligned with other CMake find modules. I would leave them alone for now and fix later.

pivotal-jbarrett:
@dgkimura https://issues.apache.org/jira/browse/GEODE-3432

Anything else you don't like?

dgkimura:
@pivotal-jbarrett: Ship-it!

PR #125, opened 2017-09-13
GEODE-3611: Upgrade boost dependency to version 1.65

dgkimura:
Upgrade to boost 65 to leverage additional modules, specifically boost::stacktrace

### DIFF
--- a/dependencies/boost/CMakeLists.txt
+++ b/dependencies/boost/CMakeLists.txt
@@ -15,10 +15,10 @@
 
 project( boost )
 
-set( ${PROJECT_NAME}_VERSION 1.64.0 )
-set( ${PROJECT_NAME}_SHA265 0445c22a5ef3bd69f5dfb48354978421a85ab395254a26b1ffb0aa1bfd63a108 )
+set( ${PROJECT_NAME}_VERSION 1.65.1 )
+set( ${PROJECT_NAME}_SHA265 a13de2c8fbad635e6ba9c8f8714a0e6b4264b60a29b964b940a22554705b6b60 )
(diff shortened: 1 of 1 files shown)


cppcache/src/CMakeLists.txt
```diff
@@ -92,6 +92,7 @@ endif()
 
 target_link_libraries(_apache-geode INTERFACE
   ACE
+  boost
```
pivotal-jbarrett: Are you sure this is absolutely necessary right now? I think the exception bits are header only.

dgkimura: I think you're probably right.  I'll see if I can just add boost headers to include path.

pivotal-jbarrett: You may have to add a project dependency to make sure Boost is “built” before we build.

dependencies/boost/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ set( ${PROJECT_NAME}_INSTALL_DIR ${INSTALL_DIR} )
 add_library(${PROJECT_NAME} INTERFACE)
 target_include_directories(${PROJECT_NAME} INTERFACE
   $<BUILD_INTERFACE:${${PROJECT_NAME}_INSTALL_DIR}/include>
+  $<BUILD_INTERFACE:${${PROJECT_NAME}_INSTALL_DIR}/src>
```
pivotal-jbarrett: It seems odd that anything should be dependent on the src directory. My understanding was that if it's in src then it gets built into a lib, otherwise it gets copied to include for header-only libraries.

dgkimura:
Updated PR based on feedback. Also switched source to download from sourceforge and change to 1.65.1

PR #129, opened 2017-09-20
Fix testThinClientSecurityMultiUserTest

dgkimura:
Fix a bug in usage of shared pointer.  Workaround a test failure by reducing workload.

Testing: Passed multiple times on Windows and Mac.


cppcache/src/ProxyCache.cpp
```diff
@@ -63,8 +63,7 @@ void ProxyCache::close() {
       prqs->closeCqs(false);
     }
 
-    ProxyCachePtr pcp(this);
-    GuardUserAttribures gua(pcp);
+    GuardUserAttribures gua(shared_from_this());
```
pivotal-jbarrett: Good find!

PR #135, opened 2017-10-10
 GEODE-3765: Rethrow test client exceptions in testing framework

pivotal-jbarrett:
Fixes test framework.
Fixes issues with a chunk of global code still floating around.
Corrects tests that were failing but not reporting as a result of the test framework bug.


pivotal-jbarrett:
Paired with @dgkimura on this so we will just merge it forward.

PR #140, opened 2017-11-02
GEODE-3754: Replace public API function out parameters with return variables

dgkimura:
Unittests and integration-tests passed on Linux and Windows


pivotal-jbarrett:
@dgkimura I have fixed formatting for you. Will keep reviewing.

PR #143, opened 2017-11-06
GEODE-3420: Remove inline header files

dgkimura:
Removes the inline header files.  Uses template specializations to preserve existing interfaces.

Testing:
- Successfully built on Windows, Linux, and Mac
- Unit/integration tests passed on Mac and Linux


cppcache/include/geode/Cacheable.hpp
```diff
@@ -35,10 +35,10 @@ typedef SerializablePtr CacheablePtr;
 typedef Serializable Cacheable;
 
 template <typename TVALUE>
-inline CacheablePtr createValue(const std::shared_ptr<TVALUE>& value);
+inline CacheablePtr createValue(const std::shared_ptr<TVALUE>& value) {
```
pivotal-jbarrett: I know you are just moving this but I really don't see the utility in this. I feel like this is a hold over from some really old magic with the old SharedPtr. All we have here now is a redefinition of dynamic cast. If we can get rid of this then let's.

cppcache/include/geode/CacheableBuiltins.hpp
```diff
@@ -642,6 +642,16 @@ _GF_CACHEABLE_CONTAINER_TYPE_DEF_(HashSetOfCacheableKey,
  */
 _GF_CACHEABLE_CONTAINER_TYPE_(HashSetOfCacheableKey, CacheableLinkedHashSet);
 
+template <>
+inline CacheableKeyPtr CacheableKey::create(int value) {
```
pivotal-jbarrett: Why only these two specializations here? I would also prefer we say `int32_t` rather than `int` to be safe.

cppcache/include/geode/CacheableString.hpp
```diff
@@ -308,6 +308,31 @@ inline CacheablePtr createValueArr(const wchar_t* value) {
   return CacheableString::create(value);
 }
 
+template <typename TVALUE>
+inline CacheablePtr createValue(const TVALUE* value) {
+  return CacheableString::create(value);
+}
```
pivotal-jbarrett: Where did this come from? Why would any `TVAL*` cause the creation of a `CacheableString`?

dgkimura: This came from `Cacheable.inl` template function `createValue(const TVALUE* value)` which called `createValueArr` which is only defined in `CacheableString.hpp`.  So intent was to unroll that. I didn't move it to `Cacheable.hpp` because then `Cacheable.hpp` and `CacheableString.hpp` would have circular dependency.

cppcache/include/geode/Cacheable.hpp
dgkimura: Sounds good. I'll give it a try.

cppcache/include/geode/CacheableBuiltins.hpp
dgkimura: These were the only 2 that were needed to get compile/tests working.  Alright, I'll switch to `int32_t`.

cppcache/include/geode/Cacheable.hpp
dgkimura: Okay, I gave it a quick try.  Region header has some interesting template type resolutions.  I can do some type trait magic, but I'm not sure it's worth it.

Here's unchanged Region signatures:

    void put(const CacheableKeyPtr& key, const CacheablePtr& value...)

    template <class KEYTYPE, class VALUETYPE>
    void put(const KEYTYPE& key, const VALUETYPE& value... )

If we call `region.put(CacheableString::create("foobar"))` this matches templated put, which uses createValue template to reinterpret CacheableStringPtr into CacheablePtr.  Ideally, it seems like instead we'd want the call to match the non-templated signature.  In order to do that I can change the signatures to something like follows:

    template <class KEYTYPE, class VALUETYPE>
    inline void put(
      const typename std::enable_if<is_integral_or_char_ptr<KEYTYPE>::value>::type& key = 0,
      const typename std::enable_if<is_integral_or_char_ptr<VALUETYPE>::value>::type& value = 0,
      const SerializablePtr& arg = nullptr) {

And then we'd have to implement `is_integral_or_char_ptr`, which should be trivial.  Is this worth it?  Did I miss something?

pivotal-jbarrett: We can correct this in a different ticket. How about you open one to address the ambiguity here.

cppcache/include/geode/CacheableString.hpp
pivotal-jbarrett: I am just curious how it handles matching say `int*` since it lacks any conditional traits. I suppose our changes to `std::string` will address this.

pivotal-jbarrett:
I am cool with the changes here as is.

cppcache/include/geode/Cacheable.hpp
dgkimura: Opened ticket here: https://issues.apache.org/jira/browse/GEODE-3972

PR #145, opened 2017-11-07
GEODE-3421: Cleanup public API headers

dgkimura:
Move the following out of the public API:
- InternalCacheTransactionManager2PC
- Log.hpp
- utils.hpp

Testing:
- Built on Mac/Windows
- Unit/integration tests passed on Mac/Windows


cppcache/test/CacheableStringEqualityTest.cpp
```diff
@@ -21,9 +21,10 @@
 #include <unordered_set>
 #include <unordered_map>
 
-#include <geode/utils.hpp>
 #include <geode/CacheableString.hpp>
 
+#include "util/functional.hpp"
```
pivotal-jbarrett: While I like the rename, I think this one header should probably remain public. It provides the hash and equal_to for dereferenced `shared_ptr`objects. It could be quite helpful. Though I suppose we can move it public later if people are asking for it.

pivotal-jbarrett:
What about Assert.hpp? Seems like it shouldn't be in the public API?

dgkimura:
Ah, I think you're right about Assert.hpp.  I hadn't considered since it wasn't in the ticket.

cppcache/test/CacheableStringEqualityTest.cpp
dgkimura: Unless you have strong opinion, I think I am going to stick with your second approach to keep it private until people are asking for it.

PR #151, opened 2017-11-15
GEODE-3149: Removing unneeded file

mhansonp:
-HashFunction.hpp is no longer needed


mhansonp:
need to make some more changes.

PR #152, opened 2017-11-15
GEODE-3624: Update C++ client exceptions to implement std::exception interface

dgkimura:
Testing:
- Successfully built on Mac, Windows, Solaris X86, Solaris SPARC
- Unit/integration tests passed on Mac, Windows


cppcache/include/geode/Exception.hpp
```diff
@@ -42,87 +44,49 @@ class StackTrace;
...
   virtual const char* getName() const;
 
-  /**
-   * Throw polymorphically; this allows storing an exception object
-   * pointer and throwing it later.
-   */
-  virtual void raise() { throw *this; }
-
-  inline std::shared_ptr<Exception> getCause() const { return m_cause; }
+  virtual const char *what() const noexcept;
```
pivotal-jbarrett: `override`?

cppcache/include/geode/Exception.hpp
```diff
@@ -42,87 +44,49 @@ class StackTrace;
...
    * @param  msg1 message pointer, this is copied into the exception.
-   * @param  msg2 optional extra message pointer, appended to msg1.
-   * @param  forceTrace enables a stacktrace for this exception regardless of
-   * stacktrace-enabled system property.
-   * @param  cause optional cause of the exception which can be later
-   *               retrieved using <code>getCause</code>
    **/
-  Exception(const char* msg1, const char* msg2 = nullptr,
-            bool forceTrace = false, const std::shared_ptr<Exception>& cause = nullptr);
+  Exception(const char* msg1);
```
pivotal-jbarrett: Does `using exception::exception;` work here? If so you may consider that to avoid having to update classes later if std::exception adds constructors.

cppcache/include/geode/Exception.hpp
```diff
@@ -42,87 +44,49 @@ class StackTrace;
...
   /** Creates an exception.
    * @param  msg1 message pointer, this is copied into the exception.
-   * @param  msg2 optional extra message pointer, appended to msg1.
-   * @param  forceTrace enables a stacktrace for this exception regardless of
-   * stacktrace-enabled system property.
-   * @param  cause optional cause of the exception which can be later
-   *               retrieved using <code>getCause</code>
    **/
-  Exception(const char* msg1, const char* msg2 = nullptr,
-            bool forceTrace = false, const std::shared_ptr<Exception>& cause = nullptr);
```
pivotal-jbarrett: Concerned about the loss of `cause`. Can you provide rational for dropping it?

cppcache/include/geode/Exception.hpp
```diff
@@ -42,87 +44,49 @@ class StackTrace;
...
+  Exception(const char* msg1);
 
-  explicit Exception(const std::string& msg1);
+  Exception(const std::string& msg1);
```
pivotal-jbarrett: I thought rules suggested (warning on some compilers) that single parameter constructors should be `explicit` to avoid issues with conversion operations and the copy constructor.

cppcache/include/geode/Exception.hpp
```diff
@@ -42,87 +44,49 @@ class StackTrace;
...
-  inline std::shared_ptr<Exception> getCause() const { return m_cause; }
+  virtual const char *what() const noexcept;
 
  protected:
-  /** internal constructor used to clone this exception */
-  Exception(const std::shared_ptr<CacheableString>& message,
-            const std::shared_ptr<StackTrace>& stack,
-            const std::shared_ptr<Exception>& cause);
-
   static bool s_exceptionStackTraceEnabled;
```
pivotal-jbarrett: What about getting rid of this and always taking the stack trace. Given we are in an exception the microseconds it takes to get the stack is more invaluable than not.

cppcache/include/geode/ExceptionTypes.hpp
```diff
@@ -33,27 +33,20 @@ namespace client {
...
+    x(const char* msg1)                                                      \
+        : Exception(msg1) {}                                                 \
+    x(const std::string& msg1)                                               \
+        : Exception(msg1) {}                                                 \
+    virtual ~x() noexcept {}                                                 \
+    virtual const char* getName() const { return _exception_name_##x; }      \
+                                                                             \
+   private:                                                                  \
+    const x& operator=(const x&);                                            \
   }
```
pivotal-jbarrett: Any way we could templetize this?

cppcache/src/Exception.cpp
```diff
@@ -36,90 +34,29 @@ void Exception::setStackTraces(bool stackTraceEnabled) {
...
-
-#endif
-
-Exception::Exception(const std::shared_ptr<CacheableString>& message,
-                     const std::shared_ptr<StackTrace>& stack,
-                     const std::shared_ptr<Exception>& cause)
-    : m_message(message), m_stack(stack), m_cause(cause) {}
-
-Exception* Exception::clone() const {
-  return new Exception(m_message, m_stack, m_cause);
```
pivotal-jbarrett: Loving all these deletes!

cppcache/src/Exception.cpp
```diff
@@ -36,90 +34,29 @@ void Exception::setStackTraces(bool stackTraceEnabled) {
...
+const char *Exception::what() const noexcept {
+  return message.c_str();
+}
 
-Exception::~Exception() {}
+Exception::~Exception() noexcept {}
 
 const char _exception_name_Exception[] = "apache::geode::client::Exception";
 
 const char* Exception::getName() const { return _exception_name_Exception; }
```
pivotal-jbarrett: Is getName is ours so why not std::string?

cppcache/src/Exception.cpp
```diff
@@ -36,90 +34,29 @@ void Exception::setStackTraces(bool stackTraceEnabled) {
...
+  if (s_exceptionStackTraceEnabled/* || forceTrace*/) {
+    m_stack = std::unique_ptr<StackTrace>();
   }
-  m_message = CacheableString::createNoCopy(msg, static_cast<int32_t>(len));
 }
 
-Exception::Exception(const std::string& msg1) : Exception(msg1.c_str()) {}
+const char *Exception::what() const noexcept {
+  return message.c_str();
+}
```
pivotal-jbarrett: What about leaving `getMessage` and having it return the std::string version of `what`?

cppcache/src/Exception.cpp
```diff
@@ -36,90 +34,29 @@ void Exception::setStackTraces(bool stackTraceEnabled) {
...
-
-Exception::Exception(const std::shared_ptr<CacheableString>& message,
-                     const std::shared_ptr<StackTrace>& stack,
-                     const std::shared_ptr<Exception>& cause)
-    : m_message(message), m_stack(stack), m_cause(cause) {}
-
-Exception* Exception::clone() const {
-  return new Exception(m_message, m_stack, m_cause);
+std::string Exception::getStackTrace() const {
+  return m_stack ? m_stack->getString() : "  No stack available.\n";
```
pivotal-jbarrett: Are the spaces in this literal intentional?

mhansonp: I like the use of std::string here...

cppcache/include/geode/Exception.hpp
dgkimura: Primary rational was that it isn't part of `std::exception`.  I'll try playing around with `std::nested_exception`.

cppcache/include/geode/Exception.hpp
dgkimura: Thanks, I'll give that a try.

cppcache/include/geode/Exception.hpp
dgkimura: Ah, I didn't know that.  I'll bring it back.

cppcache/include/geode/Exception.hpp
dgkimura: Will do.

cppcache/include/geode/Exception.hpp
dgkimura: Sounds good to me.

cppcache/include/geode/ExceptionTypes.hpp
dgkimura: I'll give it a try.

cppcache/src/Exception.cpp
dgkimura: Yes.  No reason really.  I just figured the std::string story would cover it later.  ;)

cppcache/src/Exception.cpp
dgkimura: Hmm... probably some formatting thing.  Copy/paste so I'm not sure.

cppcache/src/ExceptionTypes.cpp
```diff
@@ -39,294 +40,338 @@ void GfErrTypeThrowException(const char* str, GfErrType err) {
   }
   switch (err) {
     case GF_NOTCON: {
-      NotConnectedException ex(
-          str, (exMsg != nullptr ? exMsg : ": not connected to Geode"));
+      std::stringstream ss;
+      ss << str << (exMsg != nullptr ? exMsg : ": not connected to Geode");
+      NotConnectedException ex(ss.str());
```
mhansonp: Nitpick here, but I think you can use a string and append the values you want here. I think it would be faster, though  not by a whole lot. Why go that way?

cppcache/src/Exception.cpp
pivotal-jbarrett: Way to pass the buck!

cppcache/src/ExceptionTypes.cpp
pivotal-jbarrett: Nitpick asside, I wonder if some duplicate code could be removed but moving some things:
```
std::stringstream ss;
ss << str;
switch ... {
  case GF_NOTCON
     ss << (exMsg ? exMsg : ": not connected to Geode")
...
```

cppcache/include/geode/Exception.hpp
dgkimura: Hmm... `std::exception` has only default and copy constructors so that's not going to work.

http://en.cppreference.com/w/cpp/error/exception/exception

cppcache/src/ExceptionTypes.cpp
dgkimura: Sounds good. I incorporated in next change set.

cppcache/include/geode/ExceptionTypes.hpp
```diff
@@ -59,265 +40,369 @@ class CacheableString;
 /**
  *@brief A geode assertion exception.
  **/
-_GF_EXCEPTION_DEF(AssertionException);
+class AssertionException : public Exception {
```
pivotal-jbarrett: OMG! So much better!

cppcache/include/geode/Exception.hpp
```diff
@@ -52,9 +52,9 @@ class CPPCACHE_EXPORT Exception : public std::exception {
   /** Creates an exception.
    * @param  msg1 message pointer, this is copied into the exception.
    **/
-  Exception(const char* msg1);
+  explicit Exception(const char* msg1);
```
pivotal-jbarrett: I think it is only really necessary when the first arg is a ref to avoid a conversation copy construct. Though it probably doesn't hurt here.

clicache/src/ExceptionTypes.cpp
```diff
@@ -112,6 +112,12 @@ namespace Apache
       System::Exception^ GeodeException::Get(const apache::geode::client::Exception& nativeEx)
       {
         Exception^ innerException = nullptr;
+        try
+        {
+          std::rethrow_if_nested(nativeEx);
+        } catch (const apache::geode::client::Exception& ex) {
+          innerException = GeodeException::Get(ex);
+        }
```
pivotal-jbarrett: Crazy but hey its the C++11 way right!

PR #154, opened 2017-11-16
GEODE-3977: Modifying the QueryService template

mhansonp:
-Modified the QueryService template to only take
 one parameter instead of two.

### DIFF
--- a/clicache/src/QueryService.hpp
+++ b/clicache/src/QueryService.hpp
@@ -22,6 +22,15 @@
 #include "begin_native.hpp"
 #include <geode/QueryService.hpp>
 #include "end_native.hpp"
+#include "Query.hpp"
+#include "Log.hpp"
+#include "CqAttributes.hpp"
+#include "CqQuery.hpp"
+#include "CqServiceStatistics.hpp"
+#include "impl/ManagedString.hpp"
+#include "ExceptionTypes.hpp"
+#include "impl/SafeConvert.hpp"
+
 
 
 
@@ -51,7 +60,7 @@ namespace Apache
       /// <summary>
       /// Provides a query service.
       /// </summary>
-      generic<class TKey, class TResult>
+      generic<class TResult>
       public ref class QueryService sealed
       {
       public:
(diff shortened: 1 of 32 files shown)


clicache/src/QueryService.hpp
```diff
@@ -51,7 +60,7 @@ namespace Apache
       /// <summary>
       /// Provides a query service.
       /// </summary>
-      generic<class TKey, class TResult>
+      generic<class TResult>
```
pivotal-jbarrett: The more I think about it does it even make sense for the class itself to be templated. Does the class itself depend in any way on the `TResult`. I think the answer is no if you look at each method. The bulk of the methods are factories for `Query<TResult>` and `CqQuery<TKey, TResult>`.  So should have to ask for a new `QueryService` every time I want to make new `Query` with a different `TResult`? That seems wrong to me.
```
var qsString = cache.getQueryService<string>();
var qString = qsString.createQuery("....");
var qsInt = cache.getQueryService<int>)();
var qInt = qsInt.createQuery("...");
```
vs.
```
var qs = cache.getQueryService();
var qString = qs.createQuery<string>("....");
var qInt = qs.createQuery<int>("...");
```

pivotal-jbarrett: Additionally methods like `CloseCqs` operates on all CQs, not just the ones that match the `TResult` of the class.

clicache/src/QueryService.hpp
```diff
@@ -66,65 +75,66 @@ namespace Apache
...
 
         /// @nativeclient
         /// <summary>
         /// Get all  <c>CqQuery</c> on this client.
         /// </summary>
         /// @endnativeclient
-  //generic<class TKey, class TResult>
-	array<CqQuery<TKey, TResult>^>^ GetCqs();
+        generic<class TKey>
+	      array<CqQuery<TKey, TResult>^>^ GetCqs();
```
pivotal-jbarrett: Formatting is getting out of whack here.

clicache/src/QueryService.hpp
mhansonp: I agree, so I will pull all of the templating out of QueryService.

PR #156, opened 2017-11-17
Feature/geode 3977

mhansonp:
- Removed template arguments from QueryService
- Made some changes to pick up the nitpicks from the initial review.

### DIFF
--- a/clicache/src/QueryService.hpp
+++ b/clicache/src/QueryService.hpp
@@ -75,7 +75,7 @@ namespace Apache
         /// </summary>
         /// @endnativeclient
         generic<class TKey, class TResult>
-          CqQuery<TKey, TResult>^ NewCq(String^ query, CqAttributes<TKey, TResult>^ cqAttr, bool isDurable);
+        CqQuery<TKey, TResult>^ NewCq(String^ query, CqAttributes<TKey, TResult>^ cqAttr, bool isDurable);
   
         /// @nativeclient
         /// <summary>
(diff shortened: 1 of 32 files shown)


clicache/src/QueryService.hpp
```diff
@@ -22,6 +22,15 @@
...
 #include <geode/QueryService.hpp>
 #include "end_native.hpp"
+#include "Query.hpp"
+#include "Log.hpp"
+#include "CqAttributes.hpp"
+#include "CqQuery.hpp"
+#include "CqServiceStatistics.hpp"
+#include "impl/ManagedString.hpp"
+#include "ExceptionTypes.hpp"
+#include "impl/SafeConvert.hpp"
```
dgkimura: Why are all these headers now needed?

clicache/src/QueryService.hpp
```diff
@@ -51,80 +60,80 @@ namespace Apache
...
-        //generic<class TKey, class TResult>
+        generic<class TKey, class TResult>
         CqQuery<TKey, TResult>^ NewCq( String^ name, String^ query, CqAttributes<TKey, TResult>^ cqAttr, bool isDurable );
         /// @nativeclient
         /// <summary>
         /// Close all  <c>CqQuery</c> on this client.
         /// </summary>
         /// @endnativeclient
-	void CloseCqs();
+	      void CloseCqs();
```
dgkimura: Nit - spacing seems a little off.  Mixed tabs/spaces. Same as some of other methods below.

clicache/src/QueryService.hpp
mhansonp: I *think they are, but I pulled their equivalents from the .cpp file, which they are not needed in.

clicache/src/QueryService.hpp
mhansonp: Went through and fixed the spacing...

PR #157, opened 2017-11-22
Feature/geode 3977

mhansonp:
Cleanup of QueryService variable declaration.


pivotal-jbarrett:
@mhansonp why is there a commit for GEODE-3571 on this pull?

mhansonp:
I accidentally put the wrong bug number in the checking. Let me see if I can change it.

PR #158, opened 2017-11-22
feature/GEODE-3571

mhansonp:
GEODE-3571: Move to the Builder and Fluent models.

### DIFF
--- a/clicache/integration-test/AttributesFactoryTestsN.cs
+++ b/clicache/integration-test/AttributesFactoryTestsN.cs
@@ -38,10 +38,7 @@ namespace Apache.Geode.Client.UnitTests
 
     protected override ClientBase[] GetClients()
     {
-      m_client1 = new UnitProcess();
-      m_client2 = new UnitProcess();
-
-      return new ClientBase[] { m_client1, m_client2 };
+      return new ClientBase[] {};
     }
 
 
--- a/clicache/integration-test/AttributesFactoryTestsN.cs
+++ b/clicache/integration-test/AttributesFactoryTestsN.cs
@@ -24,7 +24,8 @@ namespace Apache.Geode.Client.UnitTests
 {
   using NUnit.Framework;
   using Apache.Geode.DUnitFramework;
-  using Apache.Geode.Client;
+  
+ 
 
   [TestFixture]
   [Category("group1")]
@@ -32,12 +33,26 @@ namespace Apache.Geode.Client.UnitTests
   [Category("generics")]
   public class AttributesFactoryTests : UnitTests
   {
+
+    private UnitProcess m_client1, m_client2;
+
+    protected override ClientBase[] GetClients()
+    {
+      m_client1 = new UnitProcess();
+      m_client2 = new UnitProcess();
+
+      return new ClientBase[] { m_client1, m_client2 };
+    }
+
+
     [Test]
     public void fluentModeltest()
     {
-      AttributesFactory<TKey, TValue> af = new AttributesFactory<TKey, TValue>();
-      Apache.Geode.Client.RegionAttributes<TKey, TValue> rattrs = af.SetLruEntriesLimit(0).SetInitialCapacity(5).CreateRegionAttributes();
+      AttributesFactory<string, string> af = new AttributesFactory<string, string>();
+      Apache.Geode.Client.RegionAttributes<string, string> rattrs = af.SetLruEntriesLimit(0).SetInitialCapacity(5).CreateRegionAttributes();
       Assert.IsNotNull(rattrs);
+      Assert.True(rattrs.LruEntriesLimit == 2);
+      Assert.True(rattrs.InitialCapacity == 5);
     }
   }
 }
--- a/clicache/integration-test/AttributesFactoryTestsN.cs
+++ b/clicache/integration-test/AttributesFactoryTestsN.cs
@@ -15,94 +15,29 @@
  * limitations under the License.
  */
 
-//using System;
-//using System.Reflection;
+using System;
+using System.Reflection;
 
-//#pragma warning disable 618
+#pragma warning disable 618
 
-//namespace Apache.Geode.Client.UnitTests
-//{
-//  using NUnit.Framework;
-//  using Apache.Geode.DUnitFramework;
-// // using Apache.Geode.Client; 
-//  using Apache.Geode.Client;
-//  //using Region = Apache.Geode.Client.IRegion<Object, Object>;
+namespace Apache.Geode.Client.UnitTests
+{
+  using NUnit.Framework;
+  using Apache.Geode.DUnitFramework;
+  using Apache.Geode.Client;
 
-//  [TestFixture]
-//  [Category("group1")]
-//  [Category("unicast_only")]
-//  [Category("generics")]
-//  public class AttributesFactoryTests : UnitTests
-//  {
-//    protected override ClientBase[] GetClients()
-//    {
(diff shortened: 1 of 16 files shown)


clicache/integration-test/AttributesFactoryTestsN.cs
```diff
@@ -15,94 +15,29 @@
...
+  [Category("unicast_only")]
+  [Category("generics")]
+  public class AttributesFactoryTests : UnitTests
+  {
+    [Test]
+    public void fluentModeltest()
+    {
+      AttributesFactory<TKey, TValue> af = new AttributesFactory<TKey, TValue>();
+      Apache.Geode.Client.RegionAttributes<TKey, TValue> rattrs = af.SetLruEntriesLimit(0).SetInitialCapacity(5).CreateRegionAttributes();
+      Assert.IsNotNull(rattrs);
```
pivotal-jbarrett: I would suggest adding a bit more to this test. If the point is to assert that the return from each set call is the same object you called the setter an then I would put that in the test. All this test does is show that the last call creates a region attribute. The intermediate set calls could be returning different factory objects.

clicache/src/geode_defs.hpp
```diff
@@ -33,6 +33,11 @@
 // Disable native code generation warning
 #pragma warning(disable: 4793)
 
+// Disable Native Conditional Unique warning
+#pragma warning(disable: 4383)
+
+// C4103alignment changed after including header, may be due to missing #pragma pack(pop)
+#pragma warning(disable: 4103)
```
pivotal-jbarrett: Are these changes necessary to relevant to the immediate issue being addressed?

clicache/src/impl/CacheWriter.hpp
```diff
@@ -76,8 +76,7 @@ namespace Apache
             RegionEvent<TKey, TValue> gevent(ev->GetNative());
             return m_writer->BeforeRegionDestroy(%gevent);
           }
-          
-          virtual void Close(Apache::Geode::Client::Region<Object^, Object^>^ region) override
+          virtual void Close(Apache::Geode::Client::Region<Object^, Object^>^ region)
```
pivotal-jbarrett: Why did the override get dropped here?

cppcache/test/AttributesFactoryTest.cpp
```diff
@@ -23,36 +23,40 @@ using namespace apache::geode::client;
 
 TEST(AttributesFactoryTest, setEntryIdleTimeoutSeconds) {
   AttributesFactory attributesFactory;
-  attributesFactory.setEntryIdleTimeout(ExpirationAction::DESTROY,
-                                        std::chrono::seconds(10));
   std::shared_ptr<RegionAttributes> regionAttributes =
-      attributesFactory.createRegionAttributes();
+      attributesFactory
```
pivotal-jbarrett: Similar to the comments above I would add a test that makes sure the the return from each set call is a reference to the same object we called the set on.

cppcache/include/geode/Exception.hpp
```diff
@@ -19,6 +19,7 @@
 
 #ifndef GEODE_EXCEPTION_H_
 #define GEODE_EXCEPTION_H_
+#pragma warning(disable : 4275)
```
pivotal-jbarrett: What does this disable? Comment?

cppcache/src/AttributesFactory.cpp
```diff
@@ -38,92 +38,100 @@ AttributesFactory::AttributesFactory(
...
-
 std::unique_ptr<RegionAttributes> AttributesFactory::createRegionAttributes() {
   std::shared_ptr<RegionAttributes> res;
   validateAttributes(m_regionAttributes);
   return std::unique_ptr<RegionAttributes>(
       new RegionAttributes(m_regionAttributes));
 }
 
-void AttributesFactory::validateAttributes(RegionAttributes& attrs) {
+void  AttributesFactory::validateAttributes(RegionAttributes& attrs) {
```
pivotal-jbarrett: Formatting

cppcache/src/AttributesFactory.cpp
```diff
@@ -164,44 +172,52 @@ void AttributesFactory::validateAttributes(RegionAttributes& attrs) {
   }
 }
 
-void AttributesFactory::setLruEntriesLimit(const uint32_t entriesLimit) {
+AttributesFactory&  AttributesFactory::setLruEntriesLimit(const uint32_t entriesLimit) {
```
pivotal-jbarrett: Formatting

pivotal-jbarrett: Looks like all these have extra spaces between the return type and method name.

cppcache/src/PoolFactory.cpp
```diff
@@ -71,17 +71,18 @@ PoolFactory::PoolFactory(const Cache& cache)
 
 PoolFactory::~PoolFactory() {}
 
-void PoolFactory::setFreeConnectionTimeout(
+PoolFactory& PoolFactory::setFreeConnectionTimeout(
```
pivotal-jbarrett: Can we add a unit test for this class?

clicache/src/impl/CacheWriter.hpp
mhansonp: it was very wrong in terms of matching the actual prototype, and I was removing a warning. I actually didn't intend to check this in.

mhansonp: I know it is supposed to do that if someone in advertently changes the base function prototype, that is why I didn't intend to check that in.

cppcache/include/geode/Exception.hpp
mhansonp: Hmm. I am surprised, I didn't comment.... I will fix that.

clicache/integration-test/AttributesFactoryTestsN.cs
mhansonp: Good point.

clicache/src/geode_defs.hpp
mhansonp: These two warnings were causing over 1000 warnings, so I could not see the warnings I was generating. I took them out as they are CLI related, but not real issues.

clicache/integration-test/AttributesFactoryTestsN.cs
```diff
@@ -24,20 +24,35 @@ namespace Apache.Geode.Client.UnitTests
...
   [TestFixture]
   [Category("group1")]
   [Category("unicast_only")]
   [Category("generics")]
   public class AttributesFactoryTests : UnitTests
   {
+
+    private UnitProcess m_client1, m_client2;
+
+    protected override ClientBase[] GetClients()
```
pivotal-jbarrett: Why did this test require adding clients? As far as I can tell all the testing runs local to the test runner and are not invoked in remote clients.

cppcache/include/geode/Exception.hpp
```diff
@@ -19,6 +19,12 @@
 
 #ifndef GEODE_EXCEPTION_H_
 #define GEODE_EXCEPTION_H_
+
+/*
+warning C4275:
+non dll-interface class 'std::exception' used as base for dll-interface
+class 'apache::geode::client::Exception'
+*/
 #pragma warning(disable : 4275)
```
pivotal-jbarrett: Pretty sure this should be pushed and then popped so that it only covers this particular class.

PR #159, opened 2017-11-28
GEODE-3645: Update CacheFactory::create to return Cache by value

dgkimura:
Test:
- Built successfully on Mac, Linux, Windows, and Solaris SPARC
- Passed unit/integration tests on Max, Linux, Windows, and Solaris SPARC

### DIFF
--- a/clicache/src/CacheFactory.cpp
+++ b/clicache/src/CacheFactory.cpp
@@ -60,12 +60,10 @@ namespace Apache
 				bool pdxIgnoreUnreadFields = false;
         bool pdxReadSerialized = false;
 				bool appDomainEnable = false; 
-        std::shared_ptr<native::Cache> nativeCache = nullptr;
         _GF_MG_EXCEPTION_TRY2
           //msclr::lock lockInstance(m_singletonSync);
           DistributedSystem::acquireDisconnectLock();
-    
-          nativeCache = m_nativeptr->get()->create( );
+          auto nativeCache = std::make_shared<native::Cache>(m_nativeptr->get()->create( ));
 
           auto cache = Cache::Create( nativeCache );
           CacheResolver::Add(nativeCache.get(), cache);
--- a/clicache/src/DistributedSystem.cpp
+++ b/clicache/src/DistributedSystem.cpp
@@ -140,9 +140,9 @@ namespace Apache
 
         // this we are calling after all .NET initialization required in
         // each AppDomain
-        auto nativeptr = native::DistributedSystem::create(mg_name.CharPtr, cache->GetNative().get(),
+        auto nativeptr = native::DistributedSystem::create(mg_name.CharPtr,
                                                             config->GetNative());
-        nativeptr->connect();
+        nativeptr->connect(cache->GetNative().get());
 
         ManagedPostConnect(cache);
 
--- a/cppcache/integration-test/CacheHelper.cpp
+++ b/cppcache/integration-test/CacheHelper.cpp
@@ -86,7 +86,8 @@ CacheHelper::CacheHelper(const char* member_id,
     pp = Properties::create();
   }
 
-  cachePtr = CacheFactory::createCacheFactory(pp)->create();
+  auto cacheFactory = CacheFactory::createCacheFactory(pp);
+  cachePtr.reset(new Cache(cacheFactory->create()));
 
   m_doDisconnect = false;
 
@@ -115,7 +116,8 @@ CacheHelper::CacheHelper(const char* member_id, const char* cachexml,
     CacheHelper::createDuplicateXMLFile(newFile, tmpXmlFile);
     pp->insert("cache-xml-file", newFile.c_str());
   }
-  cachePtr = CacheFactory::createCacheFactory(pp)->create();
+  auto cacheFactory = CacheFactory::createCacheFactory(pp);
+  cachePtr.reset(new Cache(cacheFactory->create()));
 
   m_doDisconnect = false;
 }
@@ -127,7 +129,8 @@ CacheHelper::CacheHelper(const std::shared_ptr<Properties>& configPtr,
     pp = Properties::create();
   }
 
-  cachePtr = CacheFactory::createCacheFactory(pp)->create();
+  auto cacheFactory = CacheFactory::createCacheFactory(pp);
+  cachePtr.reset(new Cache(cacheFactory->create()));
 
   auto poolFactory = cachePtr->getPoolManager().createFactory();
   addServerLocatorEPs("localhost:40404", poolFactory);
@@ -156,9 +159,9 @@ CacheHelper::CacheHelper(const bool isThinclient,
   }
   try {
     LOG(" in cachehelper before createCacheFactory");
-    cachePtr = CacheFactory::createCacheFactory(pp)
-                   ->setAuthInitialize(authInitialize)
-                   ->create();
+    auto cacheFactory = CacheFactory::createCacheFactory(pp)
+                        ->setAuthInitialize(authInitialize);
+    cachePtr.reset(new Cache(cacheFactory->create()));
     m_doDisconnect = false;
   } catch (const Exception& excp) {
     LOG("Geode exception while creating cache, logged in following line");
@@ -177,7 +180,8 @@ CacheHelper::CacheHelper(const bool isThinclient,
   }
   try {
     LOG(" in cachehelper before createCacheFactory");
-    cachePtr = CacheFactory::createCacheFactory(pp)->create();
+    auto cacheFactory = CacheFactory::createCacheFactory(pp);
+    cachePtr.reset(new Cache(cacheFactory->create()));
     m_doDisconnect = false;
   } catch (const Exception& excp) {
     LOG("Geode exception while creating cache, logged in following line");
@@ -201,7 +205,7 @@ CacheHelper::CacheHelper(const bool isThinclient, bool pdxIgnoreUnreadFields,
     LOGINFO("pdxIgnoreUnreadFields = %d ", pdxIgnoreUnreadFields);
     cfPtr->setPdxReadSerialized(pdxReadSerialized);
     cfPtr->setPdxIgnoreUnreadFields(pdxIgnoreUnreadFields);
-    cachePtr = cfPtr->create();
+    cachePtr.reset(new Cache(cfPtr->create()));
     m_doDisconnect = false;
   } catch (const Exception& excp) {
     LOG("Geode exception while creating cache, logged in following line");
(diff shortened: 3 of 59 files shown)


cppcache/src/CacheFactory.cpp
```diff
@@ -87,76 +73,83 @@ CacheFactory::CacheFactory(const std::shared_ptr<Properties> dsProps) {
...
-   const auto& pdxTypeRegistry = cacheImpl->getPdxTypeRegistry();
-   const auto& memberListForVersionStamp =
-       std::ref(*(cacheImpl->getMemberListForVersionStamp()));
+  LOGFINE("CacheFactory called DistributedSystem::connect");
+  auto cache = create(DEFAULT_CACHE_NAME, nullptr);
 
-   serializationRegistry->addType2(
-       std::bind(TXCommitMessage::create, memberListForVersionStamp));
+  auto& cacheImpl = cache.m_cacheImpl;
+  const auto& serializationRegistry = cacheImpl->getSerializationRegistry();
```
pivotal-jbarrett: I have been using `auto&&` over explicit `const auto&` a lot lately and I like it.Thoughts?

```auto&& serializationRegistry = cacheImpl->getSerializationRegistry();```
If the method returns a const then the variable is const, otherwise it isn't. Maybe it ins't the best choice but would love to hear arguments.

File: cppcache/src/CacheFactory.cpp [hunk 2]
```diff
@@ -87,76 +73,83 @@ CacheFactory::CacheFactory(const std::shared_ptr<Properties> dsProps) {
...
+  }
+  if (name[0] == '\0') {
+    name = "NativeCache";
+  }
+
+  Cache cache = Cache(name, dsProp, ignorePdxUnreadFields,
+                    pdxReadSerialized, authInitialize);
+  cache.m_cacheImpl->setAttributes(attrs);
   try {
     const char* cacheXml =
```
pivotal-jbarrett: auto?

cppcache/src/CacheImpl.cpp
```diff
@@ -88,14 +88,11 @@ CacheImpl::CacheImpl(Cache* c, const std::string& name,
...
 
-  m_cacheStats = new CachePerfStats(m_distributedSystem.get()
-                                        ->getStatisticsManager()
-                                        ->getStatisticsFactory());
   m_expiryTaskManager->begin();
 
   m_initialized = true;
-
-  m_poolManager = std::unique_ptr<PoolManager>(new PoolManager(*m_implementee));
+  m_pdxTypeRegistry = std::make_shared<PdxTypeRegistry>(this);
```
pivotal-jbarrett: Is `this` not initialized enough in the initializer list to initialize `m_pdxTypeRegistry` there?

cppcache/src/CacheImpl.cpp
```diff
@@ -609,8 +606,7 @@ std::shared_ptr<RegionInternal> rptr = nullptr;
     rptr = tmp;
   } else {
     LOGINFO("Creating local region %s", name.c_str());
-    rptr = std::make_shared<LocalRegion>(name, this, rootRegion, attrs, csptr,
-                                         shared);
+    rptr.reset(new LocalRegion(name, this, rootRegion, attrs, csptr, shared));
```
pivotal-jbarrett: Prefer `make_shared`.

cppcache/src/DistributedSystem.cpp
```diff
@@ -248,6 +235,20 @@ void DistributedSystem::connect() {
         "DistributedSystem::connect: caught unknown exception");
   }
 
+  CacheImpl* cacheImpl = CacheRegionHelper::getCacheImpl(cache);
```
pivotal-jbarrett: Auto?

clicache/src/CacheFactory.cpp
```diff
@@ -60,12 +60,10 @@ namespace Apache
 				bool pdxIgnoreUnreadFields = false;
         bool pdxReadSerialized = false;
 				bool appDomainEnable = false; 
-        std::shared_ptr<native::Cache> nativeCache = nullptr;
         _GF_MG_EXCEPTION_TRY2
           //msclr::lock lockInstance(m_singletonSync);
           DistributedSystem::acquireDisconnectLock();
-    
-          nativeCache = m_nativeptr->get()->create( );
+          auto nativeCache = std::make_shared<native::Cache>(m_nativeptr->get()->create( ));
```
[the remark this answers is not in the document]
dgkimura: I gave this a quick try, but I might call it out of scope.  Problem is that `Cache^` passes it's `shared_ptr` around by `const &` in several places to both managed/native sides.  It feels like it might explode out this pull request.  If it's alright with you I'd like to address it outside this pull request.

cppcache/integration-test/CacheHelper.cpp
```diff
@@ -127,7 +129,8 @@ CacheHelper::CacheHelper(const std::shared_ptr<Properties>& configPtr,
     pp = Properties::create();
   }
 
-  cachePtr = CacheFactory::createCacheFactory(pp)->create();
+  auto cacheFactory = CacheFactory::createCacheFactory(pp);
+  cachePtr.reset(new Cache(cacheFactory->create()));
```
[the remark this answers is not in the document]
dgkimura: Sounds good.  I'll change in next revision.

cppcache/src/CacheFactory.cpp
dgkimura: Will do.

cppcache/src/CacheImpl.cpp
dgkimura: I changed after update to pass CacheImpl instead of Cache.

`this` might be initialized enough, but I didn't want to rely on that in case PdxTypeRegistry later changed to require more of the CacheImpl object before being constructed.

cppcache/src/CacheImpl.cpp
dgkimura: Will do.

cppcache/src/DistributedSystem.cpp
dgkimura: Will do

clicache/src/DistributedSystem.cpp
```diff
@@ -140,9 +140,9 @@ namespace Apache
 
         // this we are calling after all .NET initialization required in
         // each AppDomain
-        auto nativeptr = native::DistributedSystem::create(mg_name.CharPtr, cache->GetNative().get(),
+        auto nativeptr = native::DistributedSystem::create(mg_name.CharPtr,
                                                             config->GetNative());
-        nativeptr->connect();
+        nativeptr->connect(cache->GetNative().get());
```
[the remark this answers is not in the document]
dgkimura: This was split because of circular constructor dependency between CacheImpl and DistributedSystem.  DistributedSystem uses CacheImpl for StatisticsManager.  It worked before when DistributedSystem took a Cache, but we can't do that anymore since Cache may be moved.

cppcache/integration-test/QueryHelper.hpp
```diff
@@ -251,7 +251,7 @@ void QueryHelper::populatePositionPdxData(std::shared_ptr<Region>& rptr,
 void QueryHelper::populatePDXObject(std::shared_ptr<Region>& rptr) {
   // Register PdxType Object
 
-  CacheImpl* cacheImpl = CacheRegionHelper::getCacheImpl(rptr->getCache().get());
+  CacheImpl* cacheImpl = CacheRegionHelper::getCacheImpl(&rptr->getCache());
```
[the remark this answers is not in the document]
dgkimura: I had thought about CacheRegionHelper too, but I wasn't sure if it was worth it.

cppcache/src/CacheFactory.cpp
dgkimura: Are you sure it propagates const-ness?  Below compile under clang c++11/14.

```cpp
const int f() { return 42; }
int main() {
    auto&& i = f();
    return ++i;
}
```

cppcache/integration-test/QueryHelper.hpp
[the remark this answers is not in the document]
pivotal-jbarrett: Not necessary I guess, jut more readable in my opinion. Move along...

cppcache/src/CacheFactory.cpp
pivotal-jbarrett: In this case no, but in this it does.
```
class foo {
    public:
    std::string str;
    const std::string& getString() {return str;}
}

int main() {
    auto&& str = foo{}.getString();
    return str.append("bar");
}
``` 
In your case the `const int` is value return. I don't think there is any value in the const being applied to the lvalue since its a copy anyway.

cppcache/integration-test/CacheHelper.cpp
```diff
@@ -87,7 +87,7 @@ CacheHelper::CacheHelper(const char* member_id,
   }
 
   auto cacheFactory = CacheFactory::createCacheFactory(pp);
-  cachePtr.reset(new Cache(cacheFactory->create()));
+  cachePtr = std::make_shared<Cache>(cacheFactory->create());
```
pivotal-jbarrett: 👍

cppcache/src/CacheFactory.cpp
```diff
@@ -123,11 +123,11 @@ Cache CacheFactory::create(
     name = "NativeCache";
   }
 
-  Cache cache = Cache(name, dsProp, ignorePdxUnreadFields,
+  auto cache = Cache(name, dsProp, ignorePdxUnreadFields,
                     pdxReadSerialized, authInitialize);
   cache.m_cacheImpl->setAttributes(attrs);
   try {
-    const char* cacheXml =
+    const auto* cacheXml =
```
pivotal-jbarrett: Should be `const auto` not `const auto*`. I am actually shocked that isn't a warning or an error. Pointerness is auto deduced.

cppcache/src/DistributedSystem.cpp
```diff
@@ -235,7 +235,7 @@ void DistributedSystem::connect(Cache* cache) {
         "DistributedSystem::connect: caught unknown exception");
   }
 
-  CacheImpl* cacheImpl = CacheRegionHelper::getCacheImpl(cache);
+  auto* cacheImpl = CacheRegionHelper::getCacheImpl(cache);
```
pivotal-jbarrett: Same here about `auto*`.

cppcache/src/CacheFactory.cpp
dgkimura: That example helped. Looks like value of `const auto&` is keeping local scoped variable const.  I haven't really formulated an opinion on this yet and right now I can go either way.

cppcache/src/CacheFactory.cpp
dgkimura: Will do.

cppcache/src/DistributedSystem.cpp
dgkimura: Will do.

PR #160, opened 2017-11-29
GEODE-3288: Converts char* to std::string on public API.

pivotal-jbarrett:
Excludes DataInput/Output, Serializer and CacheableString.


clicache/integration-test/CacheHelperN.cs
```diff
@@ -1888,7 +1888,7 @@ public static int getBaseLocatorPort()
           extraLocatorArgs += sslArgs;
         }
 
-        string locatorArgs = LocatorStartArgs + " --name=" + serverName + startDir + extraLocatorArgs;
+        string locatorArgs = LocatorStartArgs + " --name=" + serverName + startDir + extraLocatorArgs + " --http-service-port=0";
```
pivotal-jbarrett: This change shaves about 10s off each test!

clicache/src/CMakeLists.txt
```diff
@@ -48,21 +49,23 @@ target_link_libraries(Apache.Geode
...
-set_target_properties(Apache.Geode PROPERTIES
+set_target_properties(${PROJECT_NAME} PROPERTIES
   OUTPUT_NAME ${PRODUCT_DLL_NAME}
   VS_DOTNET_TARGET_FRAMEWORK_VERSION "v4.5.2"
-  VS_DOTNET_REFERENCES "System;System.Xml")
+  VS_DOTNET_REFERENCES "System;System.Xml"
+)
+
+include(PrecompiledHeader)
+add_precompiled_header(${PROJECT_NAME} geode_includes.hpp FORCEINCLUDE)
```
pivotal-jbarrett: Added precompiled headers to work around and speed up issue including marshal_as headers.

cppcache/src/CacheImpl.cpp
```diff
@@ -332,29 +333,27 @@ void CacheImpl::close(bool keepalive) {
...
 void CacheImpl::validateRegionAttributes(
-    const char* name, const std::shared_ptr<RegionAttributes>& attrs) const {
-  RegionKind kind = getRegionKind(attrs);
-  std::string buffer = "Cache::createRegion: \"";
-  buffer += name;
-  buffer += "\" ";
-
+    const std::string& name,
+    const std::shared_ptr<RegionAttributes>& attrs) const {
+  auto&& kind = getRegionKind(attrs);
```
dgkimura: Intentional universal reference here?

cppcache/src/CacheableString.cpp
```diff
@@ -298,32 +305,6 @@ CacheableString::~CacheableString() {
   }
 }
 
-int32_t CacheableString::logString(char* buffer, int32_t maxLength) const {
```
dgkimura: Woot!

cppcache/src/ClientMetadataService.cpp
```diff
@@ -262,17 +262,12 @@ void ClientMetadataService::getBucketServerLocation(
...
-      const auto partition = fpResolver->getPartitionName(event);
-      if (partition == nullptr) {
-        throw IllegalStateException(
-            "partition name returned by Partition resolver is null.");
-      } else {
-        bucketId = cptr->assignFixedBucketId(partition, resolvekey);
-        if (bucketId == -1) {
-          return;
-        }
+      auto&& partition = fpResolver->getPartitionName(event);
```
dgkimura: Same comment as above about intent of universal refs here.

cppcache/src/DiskPolicyType.cpp
```diff
@@ -18,23 +18,29 @@
...
   }
   return DiskPolicyType::NONE;
 }
+
+const std::string& DiskPolicyType::fromOrdinal(const uint8_t ordinal) {
+  if (NONE <= ordinal && ordinal <= PERSIST) {
+    return names[ordinal];
+  }
+
+  return names[NONE];
```
dgkimura: Ugh.. I realize you didn't introduce this, but I'm not really a fan of using enum types as integers in c++.

cppcache/src/ExpirationAction.cpp
```diff
@@ -17,33 +17,30 @@
...
   return INVALID_ACTION;
 }
 
-const char* ExpirationAction::fromOrdinal(const int ordinal) {
+const std::string& ExpirationAction::fromOrdinal(const int ordinal) {
   if (INVALIDATE <= ordinal && ordinal <= LOCAL_DESTROY) {
     return names[ordinal];
   }
-  return nullptr;
+  return names[INVALID_ACTION];
```
dgkimura: Same as other comment regarding enums

cppcache/src/PoolFactory.cpp
```diff
@@ -271,24 +271,26 @@ std::shared_ptr<Pool> PoolFactory::create(const char* name) {
...
+                                                            copyAttrs, tccm);
       } else {
-        poolDM = std::make_shared<ThinClientPoolHADM>(name, copyAttrs, tccm);
+        poolDM =
+            std::make_shared<ThinClientPoolHADM>(name.c_str(), copyAttrs, tccm);
       }
     }
 
-    cacheImpl->getPoolManager().addPool(name,
+    cacheImpl->getPoolManager().addPool(std::move(name),
```
dgkimura: Why `move`?

clicache/integration-test/CacheHelperN.cs
dgkimura: Very cool!

cppcache/include/geode/CacheAttributes.hpp
```diff
@@ -80,17 +82,16 @@ class CPPCACHE_EXPORT CacheAttributes {
   /** Sets cache level endpoints list.
    *
    */
-  void setEndpoints(char* endpoints);
+  void setEndpoints(const std::string& endpoints);
+  void setEndpoints(std::string&& endpoints);
```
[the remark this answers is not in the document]
pivotal-jbarrett: I think I was experimenting here. I should back it out. We can always add it later.

```void setEndpoints(const std::string& endpoints);```
Does not result in a copy to make the call but since internally we keep it a copy is made. No optimization can be made but none is really necessary unless the param is an r-value. The method also fails to communicate that we are keeping a copy.

```void setEndpoints(std::string endpoints);```
Communicates we are keeping a copy as copy is don't on the call. Unoptimized the copy happens also at the assignment to the member variable, though most modern compilers I believe will optimize that second copy out.

```void setEndpoints(std::string&& endpoints);```
Communicates a copy is kept and optimizes the r-value to avoid extra copies.

The two methods implemented are consistent with what the container class templates generate to handle l-values and r-values optimally.

File: cppcache/include/geode/CacheableBuiltins.hpp [hunk 10]
```diff
@@ -40,12 +40,6 @@ namespace client {
...
 
-/** sprintf implementation. */
-extern int gf_sprintf(char* buffer, const char* fmt, ...);
-
-/** snprintf implementation. */
-extern int gf_snprintf(char* buffer, int32_t maxLength, const char* fmt, ...);
-
 /** Template CacheableKey class for primitive types. */
 template <typename TObj, int8_t TYPEID, const char* TYPENAME,
           const char* SPRINTFSYM, int32_t STRSIZE>
```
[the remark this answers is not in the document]
pivotal-jbarrett: I don't think we can, `std::string` is not `constexpr`. C++17 adds `std::string_view` which is `constexpr`.

cppcache/integration-test/QueryHelper.hpp
```diff
@@ -219,8 +219,9 @@ void QueryHelper::populatePortfolioPdxData(std::shared_ptr<Region>& rptr,
       auto keyport = CacheableKey::create(portname);
 
       rptr->put(keyport, port);
-      LOGINFO("populatePortfolioPdxData:: Put for iteration current = %d done",
-              current);
+      // LOGINFO("populatePortfolioPdxData:: Put for iteration current = %d
```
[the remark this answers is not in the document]
pivotal-jbarrett: Yes, it fills the logs. I should probably just delete but for diagnostics someone may want it.

cppcache/integration-test/QueryStrings.hpp
```diff
@@ -57,14 +57,11 @@ enum queryCategory {
 
 class QueryStrings {
  public:
-  QueryStrings(queryCategory pcategory, const char* pquery,
-               bool pisLargeResultset = false) {
-    size_t querylen = static_cast<int>(strlen(pquery));
-    if (querylen < MAX_QRY_LENGTH) memcpy(_query, pquery, querylen);
-    memset(&_query[querylen], '\0', 1);
```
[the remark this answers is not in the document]
pivotal-jbarrett: Yes!

cppcache/integration-test/TallyListener.hpp
```diff
@@ -122,8 +122,8 @@ class TallyListener : public CacheListener {
     if (m_callbackArg != nullptr) {
       auto callbkArg =
           std::dynamic_pointer_cast<CacheableKey>(event.getCallbackArgument());
-      if (strcmp(m_callbackArg->toString()->asChar(),
-                 callbkArg->toString()->asChar()) == 0) {
+      if (strcmp(m_callbackArg->toString().c_str(),
+                 callbkArg->toString().c_str()) == 0) {
```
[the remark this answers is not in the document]
pivotal-jbarrett: Nope! Was using global search and replace on internal code, especially test, to just get the API cleaned up. We can optimize internals later.

cppcache/src/CacheImpl.cpp
pivotal-jbarrett: Yes, been playing with them. Your thoughts here?

cppcache/src/DiskPolicyType.cpp
pivotal-jbarrett: I agree. Can you open a ticket and we can investigate alternatives?

cppcache/src/PoolFactory.cpp
pivotal-jbarrett: Why not? The method is `PoolFactory::create(std::string name)` so I am moving the already copied value into the member variable, rather than another copy. Saves the copy of the underlying `char[]`.

clicache/src/CMakeLists.txt
```diff
@@ -48,21 +49,23 @@ target_link_libraries(Apache.Geode
 
 string(REPLACE "/EHsc" "/EHa" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /clr /wd4947 /wd4251 /wd4635 /doc /we4488")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /clr /bigobj /wd4947 /wd4251 /wd4635 /doc /we4488")
```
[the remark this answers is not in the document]
pivotal-jbarrett: Good point, though if we remove it the project won't link, so context will be re-established. ;) Would you like a comment?

cppcache/src/PoolFactory.cpp
dgkimura: I feel like it makes it just a little easier to shoot ourselves in the foot (e.g. accidentally editing to use name after this line).  Also since `PoolFactory::create` probably won't be called in a tight loop a copy doesn't seem too terrible?

cppcache/integration-test/QueryHelper.hpp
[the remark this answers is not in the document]
dgkimura: Maybe `LOGDEBUG` is a good compromise?

clicache/src/CMakeLists.txt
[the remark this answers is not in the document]
dgkimura: I would find a comment to be helpful

cppcache/src/DiskPolicyType.cpp
dgkimura: Created ticket here: https://issues.apache.org/jira/browse/GEODE-4032

pivotal-jbarrett:
@dgkimura Can I get a final blessing?

PR #166, opened 2017-12-07
GEODE-3991: Update callbacks to take region by reference

dgkimura:
Testing:
- Built on Windows
- Passed unit/integration tests on Windows

### DIFF
--- a/clicache/src/Region.hpp
+++ b/clicache/src/Region.hpp
@@ -26,7 +26,7 @@
 
 #include "IRegion.hpp"
 #include "ISubscriptionService.hpp"
-#include "native_shared_ptr.hpp"
+#include "native_conditional_shared_ptr.hpp"
 
 namespace Apache
 {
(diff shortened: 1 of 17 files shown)


clicache/src/Region.hpp
```diff
@@ -26,7 +26,7 @@
 
 #include "IRegion.hpp"
 #include "ISubscriptionService.hpp"
-#include "native_shared_ptr.hpp"
+#include "native_conditional_shared_ptr.hpp"
```
pivotal-jbarrett: I wonder if we really have anywhere left that the CLI needs to have a `shared_ptr` to the `Region`. There should be no place where the `Region` would outlive the `Cache` so I think that any public API that gets the `Region` shouldn't have any ownership responsibilities.

cppcache/integration-test/TallyLoader.hpp
```diff
@@ -38,7 +38,7 @@ class TallyLoader : virtual public CacheLoader {
   virtual ~TallyLoader() {}
 
   std::shared_ptr<Cacheable> load(
-      const std::shared_ptr<Region>& rp,
+      Region& rp,
```
pivotal-jbarrett: 👍

cppcache/src/LocalRegion.cpp
```diff
@@ -3109,7 +3109,7 @@ void LocalRegion::invokeAfterAllEndPointDisconnected() {
   if (m_listener != nullptr) {
     int64_t sampleStartNanos = startStatOpTime();
     try {
-      m_listener->afterRegionDisconnected(shared_from_this());
```
pivotal-jbarrett: Can `shared_from_this` be removed from this class hierarchy?

PR #168, opened 2017-12-12
GEODE-4082: Changes objectSize to return size_t.

pivotal-jbarrett:


tests/cpp/security/Security.cpp
```diff
@@ -755,7 +755,7 @@ std::shared_ptr<Region> Security::getRegionPtr(const char *reg) {
   try {
     if (name.empty()) {  // just get a random root region
       auto rootRegionVector = m_cache->rootRegions();
-      int32_t size = rootRegionVector.size();
+      int32_t size = static_cast<int32_t>(rootRegionVector.size());
```
dgkimura: Could this be `auto` and do we need to `static_cast`?

pivotal-jbarrett: We need the cast but we could make it auto. I didn't want to bleed these changes into anything other than the objectSize/size functions.

PR #169, opened 2017-12-14
GEODE-4098, add mutex timeout 

igorbarc:
GEODE-4098, add mutex timeout  to avoid stuck in client

### DIFF
--- a/cppcache/src/ClientMetadataService.cpp
+++ b/cppcache/src/ClientMetadataService.cpp
@@ -26,7 +26,6 @@
 #include "ClientMetadataService.hpp"
 #include "ThinClientPoolDM.hpp"
 
-
 namespace apache {
 namespace geode {
 namespace client {
@@ -144,12 +143,11 @@ void ClientMetadataService::getClientPRMetadata(const char* regionFullPath) {
                                               reply.getFpaSet());
       if (m_bucketWaitTimeout > std::chrono::milliseconds::zero() &&
           reply.getNumBuckets() > 0) {
-
-        std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock, std::chrono::seconds(60));
-        if (!guard.owns_lock())
-        {
-            LOGERROR("Buckets statuses container read mutex timeout 60 seconds");
-            return;
+        std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock,
+                                                 std::chrono::seconds(60));
+        if (!guard.owns_lock()) {
+          LOGERROR("Buckets statuses container read mutex timeout 60 seconds");
+          return;
         }
         m_bucketStatus[regionFullPath] = new PRbuckets(reply.getNumBuckets());
       }
@@ -468,11 +466,12 @@ void ClientMetadataService::markPrimaryBucketForTimeout(
     std::shared_ptr<BucketServerLocation>& serverLocation, int8_t& version) {
   if (m_bucketWaitTimeout == std::chrono::milliseconds::zero()) return;
 
-
-  std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock, std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
-  if (!guard.owns_lock())
-  {
-    LOGERROR("Buckets statuses container write mutex timeout %d milliseconds", DEFAULT_MUTEX_TIMEOUT);
+  std::unique_lock<std::timed_mutex> guard(
+      m_timedBucketStatusLock,
+      std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
+  if (!guard.owns_lock()) {
+    LOGERROR("Buckets statuses container write mutex timeout %d milliseconds",
+             DEFAULT_MUTEX_TIMEOUT);
     return;
   }
 
@@ -819,10 +818,12 @@ void ClientMetadataService::markPrimaryBucketForTimeoutButLookSecondaryBucket(
     std::shared_ptr<BucketServerLocation>& serverLocation, int8_t& version) {
   if (m_bucketWaitTimeout == std::chrono::milliseconds::zero()) return;
 
-  std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock, std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
-  if (!guard.owns_lock())
-  {
-    LOGERROR("Buckets statuses container write mutex timeout %d milliseconds", DEFAULT_MUTEX_TIMEOUT);
+  std::unique_lock<std::timed_mutex> guard(
+      m_timedBucketStatusLock,
+      std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
+  if (!guard.owns_lock()) {
+    LOGERROR("Buckets statuses container write mutex timeout %d milliseconds",
+             DEFAULT_MUTEX_TIMEOUT);
     return;
   }
 
@@ -876,12 +877,14 @@ bool ClientMetadataService::isBucketMarkedForTimeout(const char* regionFullPath,
                                                      int32_t bucketid) {
   if (m_bucketWaitTimeout == std::chrono::milliseconds::zero()) return false;
 
-    std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock, std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
-    if (!guard.owns_lock())
-    {
-      LOGERROR("Buckets statuses container read mutex timeout %d milliseconds", DEFAULT_MUTEX_TIMEOUT);
-      return false ;
-    }
+  std::unique_lock<std::timed_mutex> guard(
+      m_timedBucketStatusLock,
+      std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
+  if (!guard.owns_lock()) {
+    LOGERROR("Buckets statuses container read mutex timeout %d milliseconds",
+             DEFAULT_MUTEX_TIMEOUT);
+    return false;
+  }
 
   const auto& bs = m_bucketStatus.find(regionFullPath);
   if (bs != m_bucketStatus.end()) {
--- a/cppcache/src/ClientMetadataService.cpp
+++ b/cppcache/src/ClientMetadataService.cpp
@@ -144,9 +144,10 @@ void ClientMetadataService::getClientPRMetadata(const char* regionFullPath) {
       if (m_bucketWaitTimeout > std::chrono::milliseconds::zero() &&
           reply.getNumBuckets() > 0) {
         std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock,
-                                                 std::chrono::seconds(60));
+                                                 CLOSE_SERVICE_MUTEX_TIMEOUT);
         if (!guard.owns_lock()) {
-          LOGERROR("Buckets statuses container read mutex timeout 60 seconds");
+          LOGERROR("Buckets statuses container read mutex timeout %lld seconds",
+                   CLOSE_SERVICE_MUTEX_TIMEOUT);
           return;
         }
         m_bucketStatus[regionFullPath] = new PRbuckets(reply.getNumBuckets());
@@ -466,11 +467,10 @@ void ClientMetadataService::markPrimaryBucketForTimeout(
     std::shared_ptr<BucketServerLocation>& serverLocation, int8_t& version) {
   if (m_bucketWaitTimeout == std::chrono::milliseconds::zero()) return;
 
-  std::unique_lock<std::timed_mutex> guard(
-      m_timedBucketStatusLock,
-      std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
+  std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock,
+                                           DEFAULT_MUTEX_TIMEOUT);
   if (!guard.owns_lock()) {
-    LOGERROR("Buckets statuses container write mutex timeout %d milliseconds",
+    LOGERROR("Buckets statuses container write mutex timeout %lld milliseconds",
              DEFAULT_MUTEX_TIMEOUT);
     return;
   }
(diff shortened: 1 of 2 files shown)


cppcache/src/ClientMetadataService.cpp
```diff
@@ -143,7 +144,13 @@ void ClientMetadataService::getClientPRMetadata(const char* regionFullPath) {
                                               reply.getFpaSet());
       if (m_bucketWaitTimeout > std::chrono::milliseconds::zero() &&
           reply.getNumBuckets() > 0) {
-        WriteGuard guard(m_PRbucketStatusLock);
+        //ReadGuard guard( m_PRbucketStatusLock );
+        std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock, std::chrono::milliseconds(1000));
```
pivotal-jbarrett: What is the rational for hard coded 1s (1000ms) timeout?

Consider using the value of chrono and not doing math yourself.
```std::chrono::seconds(1)```

Formatting rules should be obeyed. Use clang-format, a .clang-format file exists to apply style rules.

File: cppcache/src/ClientMetadataService.cpp [hunk 2]
```diff
@@ -143,7 +144,13 @@ void ClientMetadataService::getClientPRMetadata(const char* regionFullPath) {
                                               reply.getFpaSet());
       if (m_bucketWaitTimeout > std::chrono::milliseconds::zero() &&
           reply.getNumBuckets() > 0) {
-        WriteGuard guard(m_PRbucketStatusLock);
+        //ReadGuard guard( m_PRbucketStatusLock );
```
pivotal-jbarrett: Delete! Don't comment.

cppcache/src/ClientMetadataService.cpp
```diff
@@ -461,7 +468,13 @@ void ClientMetadataService::markPrimaryBucketForTimeout(
     std::shared_ptr<BucketServerLocation>& serverLocation, int8_t& version) {
   if (m_bucketWaitTimeout == std::chrono::milliseconds::zero()) return;
 
-  WriteGuard guard(m_PRbucketStatusLock);
+  //ReadGuard guard( m_PRbucketStatusLock );
+  std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock, std::chrono::milliseconds(1000));
```
pivotal-jbarrett: Same comments above. If the same timeout is expected to occur in multiple locations then create a `constexp` for the timeout.

cppcache/src/ClientMetadataService.cpp
```diff
@@ -806,7 +819,13 @@ void ClientMetadataService::markPrimaryBucketForTimeoutButLookSecondaryBucket(
...
```
pivotal-jbarrett: Same comment above.

cppcache/src/ClientMetadataService.cpp
```diff
@@ -25,6 +25,7 @@
 #include "TcrMessage.hpp"
 #include "ClientMetadataService.hpp"
 #include "ThinClientPoolDM.hpp"
+#include <chrono>
```
pivotal-jbarrett: Please group with other system headers.

cppcache/src/ClientMetadataService.hpp
```diff
@@ -222,10 +223,11 @@ class ClientMetadataService : public ACE_Task_Base,
   Pool* m_pool;
   Queue<std::string>* m_regionQueue;
 
-  ACE_RW_Thread_Mutex m_PRbucketStatusLock;
+  //ACE_RW_Thread_Mutex m_PRbucketStatusLock;
```
pivotal-jbarrett: Delete!

cppcache/src/ClientMetadataService.hpp
```diff
@@ -38,6 +38,7 @@
 #include "DistributedSystemImpl.hpp"
 #include "NonCopyable.hpp"
 #include "util/functional.hpp"
+#include <mutex>
```
pivotal-jbarrett: Group with other system headers.

pivotal-jbarrett:
I would also like to see a test for this change. Is there an issue you can reproduce in a test to exercise this change?

echobravopapa:
Also, the Travis build should succeed - it has a compilation error....

igorbarc:
@pivotal-jbarrett Thanks, for review, passed now with the the changes 

The issue( lock in client)  reproduced in geode 1.2.0 release ,  under high load from multiple clients, possibly related to buckets moving, rebalancing at the time, with  single-hop enabled for the relevant   region

gemfire::ReadGuard::ReadGuard(ACE_RW_Thread_Mutex&)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ReadWriteLock.hpp:33
gemfire::ClientMetadataService::isBucketMarkedForTimeout(char const*, int)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ClientMetadataService.cpp:1817
6) gemfire::ThinClientPoolDM::getConnectionFromQueueW(GfErrType*, std::set<gemfire::ServerLocation, std::less<gemfire::ServerLocation>, std::allocator<gemfire::ServerLocation> >&, bool, gemfire::TcrMessage&, signed char&, bool&, bool&, gemfire::SharedPtr<gemfire::BucketServerLocation> const&)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ThinClientPoolDM.hpp:247
7) gemfire::ThinClientPoolDM::sendSyncRequest(gemfire::TcrMessage&, gemfire::TcrMessage&, bool, bool, gemfire::SharedPtr<gemfire::BucketServerLocation> const&)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ThinClientPoolDM.cpp:1400
8) gemfire::ThinClientPoolDM::sendSyncRequest(gemfire::TcrMessage&, gemfire::TcrMessage&, bool, bool)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ThinClientPoolDM.cpp:1328
9) gemfire::ThinClientRegion::executeFunction(char const*, gemfire::SharedPtr<gemfire::Serializable> const&, gemfire::SharedPtr<gemfire::CacheableVector>, unsigned char, gemfire::SharedPtr<gemfire::ResultCollector>, int, unsigned int)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ThinClientRegion.cpp:2955
10) gemfire::ExecutionImpl::execute(char const*, bool, unsigned int, bool, bool, bool)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ExecutionImpl.cpp:226
11) gemfire::ExecutionImpl::execute(char const*, bool, unsigned int, bool, bool)/jphome/jp/ccjp/mb_ccjp/workspace/GEODE_Maven_Build/geode-client-native/src/cppcache/impl/ExecutionImpl.cpp:83


The issue do not seems to be reproducing with the fix

echobravopapa:
@igorbarc would you please create a test that your fix may be validated against as @pivotal-jbarrett requested, Thanks in advance EB

pivotal-jbarrett:
@igorbarc What is the expected behavior of these functions should they return without updating the member data? Under what circumstances is this mutex held indefinitely that making it timed is corrected? Strikes me that this fix is hiding the real issue.

Can you please also correct formatting.

igorbarc:
I believe the GEODE-4051 (uncaught timeout exception) was possible root cause server side  of indefinitely held mutex 
With above fix , we see  bucket not available timeouts responses  , instead of deadlocked threads in this flow 

Will update formatting shortly

igorbarc:
Fixed formatting with clang-format

cppcache/src/ClientMetadataService.cpp
```diff
@@ -858,7 +877,14 @@ bool ClientMetadataService::isBucketMarkedForTimeout(const char* regionFullPath,
                                                      int32_t bucketid) {
   if (m_bucketWaitTimeout == std::chrono::milliseconds::zero()) return false;
 
-  ReadGuard guard(m_PRbucketStatusLock);
+  std::unique_lock<std::timed_mutex> guard(
+      m_timedBucketStatusLock,
+      std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
```
pivotal-jbarrett: `DEFAULT_MUTEX_TIMEOUT` is already a duration so this code should just read:
```c++
std::unique_lock<std::timed_mutex> guard(
       m_timedBucketStatusLock,
       DEFAULT_MUTEX_TIMEOUT);
```

cppcache/src/ClientMetadataService.cpp
```diff
@@ -143,7 +143,12 @@ void ClientMetadataService::getClientPRMetadata(const char* regionFullPath) {
                                               reply.getFpaSet());
       if (m_bucketWaitTimeout > std::chrono::milliseconds::zero() &&
           reply.getNumBuckets() > 0) {
-        WriteGuard guard(m_PRbucketStatusLock);
+        std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock,
+                                                 std::chrono::seconds(60));
```
pivotal-jbarrett: Why 60 seconds here where all the other uses is 1 second and uses a `constexpr`?

mhansonp: Random numbers need explanation. Hence using a well titled define or comment. Just having 60 there helps no one.

cppcache/src/ClientMetadataService.cpp
```diff
@@ -461,7 +466,14 @@ void ClientMetadataService::markPrimaryBucketForTimeout(
     std::shared_ptr<BucketServerLocation>& serverLocation, int8_t& version) {
   if (m_bucketWaitTimeout == std::chrono::milliseconds::zero()) return;
 
-  WriteGuard guard(m_PRbucketStatusLock);
+  std::unique_lock<std::timed_mutex> guard(
+      m_timedBucketStatusLock,
+      std::chrono::milliseconds(DEFAULT_MUTEX_TIMEOUT));
+  if (!guard.owns_lock()) {
+    LOGERROR("Buckets statuses container write mutex timeout %d milliseconds",
+             DEFAULT_MUTEX_TIMEOUT);
```
mhansonp: %d should cause a warning. switch to %lld.

cppcache/src/ClientMetadataService.cpp
```diff
@@ -806,7 +818,14 @@ void ClientMetadataService::markPrimaryBucketForTimeoutButLookSecondaryBucket(
...
```
mhansonp: same here:  %d should cause a warning. switch to %lld.

cppcache/src/ClientMetadataService.cpp
igorbarc: Other uses  come in the context of  read/white requests , that has their own timeout,  beyond  which  as I understand the is no point to  continue waiting  for bucket.  I think now,  maybe it is better I will  use  read/write timeout constants here as basis  for mutex timeout , instead of defining new ones?  

Here specifically  though,  the flow is stopping the service for the pool, so I think the timeout can be higher , that is why defined 1 minute

igorbarc: Would it be ok to define constant like CLOSE_SERVICE_MUTEX_TIMEOUT for this ?

igorbarc:
@pivotal-jbarrett @mhansonp 
Thanks for review. Made requested changes

cppcache/src/ClientMetadataService.cpp
```diff
@@ -143,7 +143,13 @@ void ClientMetadataService::getClientPRMetadata(const char* regionFullPath) {
                                               reply.getFpaSet());
       if (m_bucketWaitTimeout > std::chrono::milliseconds::zero() &&
           reply.getNumBuckets() > 0) {
-        WriteGuard guard(m_PRbucketStatusLock);
+        std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock,
+                                                 CLOSE_SERVICE_MUTEX_TIMEOUT);
+        if (!guard.owns_lock()) {
+          LOGERROR("Buckets statuses container read mutex timeout %lld seconds",
```
pivotal-jbarrett: Sorry you were given incorrect advice here. `std::chrono::duration` should not be assumed to convert to a `uint64_t`. The internal representation of the duration value is system dependent. No explicit conversion is defined in the std C++ spec to any primitive integer type. No std C++ method exists for duration to string conversions, though a a few custom options exist for converting `std::chrono::duration` to a string. I prefer one that uses our utility class for converting to a string representation with unit. 
```c++
LOGERROR("Buckets statuses container read mutex timeout %s", 
    util::chrono::duration::to_string(CLOSE_SERVICE_MUTEX_TIMEOUT).c_str());
```
Or using the `std::string` overload of `LOGERROR`:
```c++
LOGERROR("Buckets statuses container read mutex timeout " + 
    util::chrono::duration::to_string(CLOSE_SERVICE_MUTEX_TIMEOUT));
```
You may need to include "util/chrono/duration.hpp".

The above will produce log message like this:
```
Buckets statuses container read mutex timeout 60s.
```

igorbarc:
@pivotal-jbarrett Thanks, made requested changes

cppcache/src/ClientMetadataService.cpp
```diff
@@ -143,7 +144,14 @@ void ClientMetadataService::getClientPRMetadata(const char* regionFullPath) {
                                               reply.getFpaSet());
       if (m_bucketWaitTimeout > std::chrono::milliseconds::zero() &&
           reply.getNumBuckets() > 0) {
-        WriteGuard guard(m_PRbucketStatusLock);
+        std::unique_lock<std::timed_mutex> guard(m_timedBucketStatusLock,
```
pivotal-jbarrett: On a more detailed review I am confused as to what the name CLOSE_SERVICE_MUTEX_TIMEOUT means in this context. There is not closing or service in near context to this use. Please explain in more detail what this symbol means or perhaps rename to something more meaningful.

igorbarc: As I understand this function getClientPRMetadata , only called by svc() when it is called in  stop ClientMetadataService flow ( line 93)

pivotal-jbarrett: The `getClientPRMetadata() method is called in the `svc()` while in the running loop. This means it is called repeatedly while the `ClientMetadataService` is running. So again I question why it has a different timeout value than any of the others.

Looking at all the code around this I can completely understand the confusion. It is very poorly implemented. From what I have deciphered the `svc()` loop is waiting for a region name to arrive on the queue by waiting on the semaphore at the top of the loop. Then it takes the next name off the queue followed by all the other names that happen to match, effectively coalescing duplicates (with major logic error). After coalescing then it executes the `getClientPRMetadata()` for that region and goes back to the top of the loop. This continues until the queue return `nullptr` indicating it is empty at which point it waits on the semaphore again. This outer loop runs for the life of the pool basically.

pivotal-jbarrett:
Closing for no activity. Please reopen if you want to follow up on this.

PR #172, opened 2017-12-21
GEODE-4143: Improves integration test stability.

pivotal-jbarrett:
- Fixes test hangs on Windows.
- Fixes crashes on exception.
- Disables http service in locator startup.


cppcache/integration-test/CacheHelper.cpp
```diff
@@ -1848,16 +1851,17 @@ int CacheHelper::getRandomAvailablePort() {
 }
 
 std::string CacheHelper::unitTestOutputFile() {
-  char currWDPath[512];
```
dgkimura: Looks brittle.  If only we had C++17, then we'd also have `std::filesystem`... [sigh]

clicache/integration-test/CacheHelperN.cs
```diff
@@ -1990,7 +1990,7 @@ static int getLocatorPort(int num)
         {
           extraLocatorArgs = locatorPort;
         }
-        string locatorArgs = LocatorStartArgs + " --name=" + serverName + startDir + extraLocatorArgs;
+        string locatorArgs = LocatorStartArgs + " --name=" + serverName + startDir + extraLocatorArgs + " --http-service-port=0";
```
dgkimura: How did this improve stability?

cppcache/integration-test/CacheHelper.cpp
echobravopapa: Someday @dgkimura someday

clicache/integration-test/CacheHelperN.cs
pivotal-jbarrett: This line was not consistent with the other line that starts the locator. It also shaves 10-15s off the locator startup, lowers resource consumption, and prevents undeletable long directory names.

cppcache/integration-test/CacheHelper.cpp
pivotal-jbarrett: We could use boost::filesystem here. I just wanted to make it less nasty than it was to move forward since this is framework code.

PR #174, opened 2017-12-26
 GEODE-4071: Stabilizes CLI integration tests, ThinClientFunctionExecutionTestsN

pivotal-jbarrett:
There are a couple of seemingly unrelated issues here but they relate. The unused code was just unused code and cleaning up make sense. The static lock around DS was left over form the app domain refcounting and since this test was creating and destroying lots of caches it made sense to eliminate it as a cause. Ultimately it came down to this test being long running and the windows stats collector being very buggy. Eventually the stats collector would cause a segfault and the test would either hang or fail.

### DIFF
--- a/clicache/integration-test/CacheHelperN.cs
+++ b/clicache/integration-test/CacheHelperN.cs
@@ -2168,7 +2168,14 @@ namespace Apache.Geode.Client.UnitTests
         outSr.Close();
         if (!started)
         {
-          javaProc.Kill();
+          try
+          {
+            javaProc.Kill();
+          }
+          catch
+          {
+            //ignore
+          }
         }
         Assert.IsTrue(started, "Timed out waiting for " +
           "Java cacheserver to start.{0}Please check the server logs.",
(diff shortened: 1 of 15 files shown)


rhoughton-pivot:
So test splitting for good sanitary tests (do one thing) and runtime. Removing the lock for performance? In what scenario would a client be using a cache in a concurrent way? I guess I don't have context about the code to say "yeah, you don't need the lock there, so go ahead and remove it for performance"

pivotal-jbarrett:
@rhoughton-pivot The lock isn't being removed for performance but rather because it is unused. It used to synchronize a reference counter between .NET AppDomains and the C++ singleton DistributedSystem. Now that there is a one to one mapping and no more singletons the ref counting is gone and the lock is useless. I found the lock while looking at an area of code that blew up debugging this test. One could argue the change be moved to another ticket. I don't really care either way.

rhoughton-pivot:
@pivotal-jbarrett I understand now. I think a second ticket is overkill for removal of dead code.

PR #175, opened 2017-12-27
Geode 4171

igodwin:
Fixes to the BUILDING.md.
- Removed src from cmake command, and added GEODE_ROOT build option
- Added Apache Geode binaries to Required Tools section


igodwin:
Oops this was against master.

PR #178, opened 2018-01-03
GEODE-3998: Links Cache::Name to DistributeSystem::Name

pivotal-jbarrett:


cppcache/src/CacheFactory.cpp
```diff
@@ -109,39 +132,6 @@ Cache CacheFactory::create() const {
...
-      cache.close();
-    }
-    throw apache::geode::client::UnknownException(
-        "Exception thrown in CacheFactory::create");
-  }
-
-  return cache;
-}
-
 std::shared_ptr<CacheFactory> CacheFactory::set(const std::string& name,
```
echobravopapa: If Cache name is linked to DS name, doesn't this need to flow through to the DS...?

pivotal-jbarrett: Yes, but I am not sure what you are referring to here?

echobravopapa: CacheFactory::set() takes a name parameter, which is inserted into the dsProperties... is that guaranteed to update the name the DS uses?

pivotal-jbarrett: This method takes a name/value pair to update ds properties. It is not explicitly setting the "ds name". 

The fact that this is the worst way to handle configuration is outside the scope of this issue.

echobravopapa: fair enough...

PR #182, opened 2018-01-09
GEODE-4188: Moves headers and templates internal.

pivotal-jbarrett:
- Moves chrono::duration extensions internal.
- Moves functional extensions internal.


cppcache/include/geode/Properties.hpp
```diff
@@ -97,7 +98,7 @@ class CPPCACHE_EXPORT Properties : public Serializable {
   template <class _Rep, class _Period>
   void insert(std::string key,
               const std::chrono::duration<_Rep, _Period>& value) {
-    insert(key, util::chrono::duration::to_string(value));
+    insert(key, to_string(value));
```
echobravopapa: duration is no longer needed here?

PR #187, opened 2018-01-11
GEODE-4282: Fixes createValue/Key functions.

pivotal-jbarrett:


pivotal-jbarrett:
For some reason this doesn't compile on GCC despite compiling just fine on Clang. I am investigating. Will re-open when resolved.

PR #188, opened 2018-01-17
GEODE-3391: Update Transaction id to pass by references

dgkimura:
Updates _CacheTransactionManager.hpp_ to use type ```TransactionId&``` instead of ```std::shared_ptr<TransactionId>```.

Notes: Important to note this change to ref forces behavior change because previous code returned nullptr when suspending transaction if transaction isn't currently in flight ([example](https://github.com/apache/geode-native/compare/develop...dgkimura:feature/GEODE-3391-transactionidref?expand=1#diff-1d700ac09f0ace4175f618859ead7f34L329)).  Now we instead throw.  If we wanted to preserve previous behavior, we would likely need to use raw/smart pointers instead of references.

Testing: All stable integration tests and unittests passed on Mac and Windows.  Note that all but 1 of transaction tests are currently marked flaky...  I will try to tackle fixing these tests in separate pull-request.

### DIFF
--- a/clicache/integration-test/ThinClientCSTXN.cs
+++ b/clicache/integration-test/ThinClientCSTXN.cs
@@ -409,7 +409,9 @@ namespace Apache.Geode.Client.UnitTests
         resumeEx = true;
       }
       Assert.AreEqual(resumeEx, true, "The transaction should not be resumed");
-      Assert.AreEqual(CacheHelper.CSTXManager.Suspend(), null, "The transaction should not be suspended");
+
+      // The transaction should not be suspended
+      // Assert.Throws<Exception>( delegate { CacheHelper.CSTXManager.Suspend(); });
     }
 
     public void CallOp()
--- a/clicache/src/Cache.cpp
+++ b/clicache/src/Cache.cpp
@@ -99,7 +99,7 @@ namespace Apache
         {
           auto nativeptr = std::dynamic_pointer_cast<InternalCacheTransactionManager2PC>(
             m_nativeptr->get()->getCacheTransactionManager());
-          return Apache::Geode::Client::CacheTransactionManager::Create(nativeptr);
+          return Apache::Geode::Client::CacheTransactionManager::Create(nativeptr.get());
         }
         finally
         {
--- a/clicache/src/CacheTransactionManager.cpp
+++ b/clicache/src/CacheTransactionManager.cpp
@@ -35,14 +35,7 @@ namespace Apache
       {
         _GF_MG_EXCEPTION_TRY2
 
-          try
-          {
-            m_nativeptr->get()->begin( );
-          }
-          finally
-          {
-            GC::KeepAlive(m_nativeptr);
-          }
+          m_nativeptr->begin( );
 
         _GF_MG_EXCEPTION_CATCH_ALL2
       }
@@ -51,14 +44,7 @@ namespace Apache
       {
         _GF_MG_EXCEPTION_TRY2
 
-          try
-          {
-            m_nativeptr->get()->prepare( );
-          }
-          finally
-          {
-            GC::KeepAlive(m_nativeptr);
-          }
+          m_nativeptr->prepare( );
 
         _GF_MG_EXCEPTION_CATCH_ALL2
       }
@@ -66,28 +52,18 @@ namespace Apache
       void CacheTransactionManager::Commit( )
       {
         _GF_MG_EXCEPTION_TRY2
-          try
-          {
-            m_nativeptr->get()->commit( );
-          }
-          finally
-          {
-            GC::KeepAlive(m_nativeptr);
-          }
+
+            m_nativeptr->commit( );
+
         _GF_MG_EXCEPTION_CATCH_ALL2
       }
 
       void CacheTransactionManager::Rollback( )
       {
         _GF_MG_EXCEPTION_TRY2
-          try
-          {
-            m_nativeptr->get()->rollback( );
-          }
-          finally
-          {
-            GC::KeepAlive(m_nativeptr);
-          }
+
+            m_nativeptr->rollback( );
+
         _GF_MG_EXCEPTION_CATCH_ALL2
       }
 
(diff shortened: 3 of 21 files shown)


clicache/integration-test/ThinClientCSTXN.cs
```diff
@@ -409,7 +409,9 @@ public void SuspendResumeCommit()
         resumeEx = true;
       }
       Assert.AreEqual(resumeEx, true, "The transaction should not be resumed");
-      Assert.AreEqual(CacheHelper.CSTXManager.Suspend(), null, "The transaction should not be suspended");
+
+      // The transaction should not be suspended
```
pivotal-jbarrett: Was your intention to add a test for the exception here?

clicache/src/Cache.cpp
```diff
@@ -99,7 +99,7 @@ namespace Apache
         {
           auto nativeptr = std::dynamic_pointer_cast<InternalCacheTransactionManager2PC>(
             m_nativeptr->get()->getCacheTransactionManager());
```
pivotal-jbarrett: I thought this was now supposed to return a ref not a shared_ptr?

clicache/src/CacheTransactionManager.cpp
```diff
@@ -35,14 +35,7 @@ namespace Apache
...
 
-          try
-          {
-            m_nativeptr->get()->begin( );
-          }
-          finally
-          {
-            GC::KeepAlive(m_nativeptr);
-          }
+          m_nativeptr->begin( );
```
pivotal-jbarrett: Do you still need the keepalive here? This CLI object can collect while still in the native begin method. Does the CLI object own and destroy the native object?

cppcache/include/geode/CacheTransactionManager.hpp
```diff
@@ -173,15 +173,15 @@ class _GEODE_EXPORT CacheTransactionManager {
    * @since 3.6.2
    * @see #isSuspended(TransactionId)
    */
-  virtual bool exists(std::shared_ptr<TransactionId> transactionId) = 0;
+  virtual bool exists(TransactionId& transactionId) = 0;
```
pivotal-jbarrett: Could TransactionId be a value type? My thinking is what if someone needed to suspend the tx, keep a copy of the id, maybe pass to another thread, and resume. It was easy to do with shared pointer, harder to do with ref, easy to do as value.

cppcache/src/CacheTransactionManagerImpl.cpp
```diff
@@ -321,12 +321,12 @@ ThinClientPoolDM* CacheTransactionManagerImpl::getDM() {
...
 
 Cache* CacheTransactionManagerImpl::getCache() { return m_cache->getCache(); }
-std::shared_ptr<TransactionId> CacheTransactionManagerImpl::suspend() {
+TransactionId& CacheTransactionManagerImpl::suspend() {
   // get the current state of the thread
   TXState* txState = TSSTXStateWrapper::s_geodeTSSTXState->getTXState();
   if (txState == nullptr) {
     LOGFINE("Transaction not in progress. Returning nullptr transaction Id.");
-    return nullptr;
+    throw TransactionException("Transaction not in progress.");
```
pivotal-jbarrett: 👍 Love this so much more than `nullptr`.

cppcache/src/CacheTransactionManagerImpl.cpp
```diff
@@ -364,16 +364,16 @@ std::shared_ptr<TransactionId> CacheTransactionManagerImpl::suspend() {
...
   // add the transaction state to the list of suspended transactions
-  addSuspendedTx(txState->getTransactionId()->getId(), txState);
+  addSuspendedTx(txState->getTransactionId().getId(), txState);
 
   // set the current transaction state as null
   TSSTXStateWrapper::s_geodeTSSTXState->setTXState(nullptr);
 
   // return the transaction ID
-  return txState->getTransactionId();
+  return static_cast<TransactionId&>(txState->getTransactionId());
```
pivotal-jbarrett: What's with this cast?

cppcache/include/geode/CacheTransactionManager.hpp
dgkimura: I agree that would be nicer.  And originally that was my plan too.  However, I think problem I ran into was that `TransactionId` is an abstraction and actual class being passed is `TXId`.  If you'd like, I can investigate meaning behind this abstraction and whether it can be dropped and then make change to value type accordingly.  Thoughts?

clicache/src/Cache.cpp
dgkimura: That seems reasonable to me.  But, that's a change to `Cache` interface and I assumed scope of this change was only `CacheTransactionManager`.

cppcache/src/CacheTransactionManagerImpl.cpp
dgkimura: Type being returned is a `TXId`.  Similar to above comment, it may be worth investigating if abstraction can be removed.  Is that something that should be tackled in a separate GEODE ticket or shall I give it a quick stab as part of GEODE-3391?

clicache/integration-test/ThinClientCSTXN.cs
dgkimura: Yup, thanks for catching.

cppcache/include/geode/CacheTransactionManager.hpp
pivotal-jbarrett: Let's investigate separate from this ticket.

cppcache/src/CacheTransactionManagerImpl.cpp
pivotal-jbarrett: Let's investigate separate from this ticket.

cppcache/integration-test/CMakeLists.txt
```diff
@@ -183,8 +183,6 @@ set_property(TEST testThinClientPutAllPRSingleHop PROPERTY LABELS FLAKY)
 set_property(TEST testThinClientPutAllWithCallBackArgWithoutConcurrency PROPERTY LABELS FLAKY)
 set_property(TEST testThinClientSecurityCQAuthorizationMU PROPERTY LABELS FLAKY)
 set_property(TEST testThinClientTXFailover PROPERTY LABELS FLAKY)
-set_property(TEST testThinClientTransactionsWithSticky PROPERTY LABELS FLAKY)
```
pivotal-jbarrett: Yay for fixed tests!

PR #189, opened 2018-01-22
GEODE-4282: Remove createKey/Value templates.

pivotal-jbarrett:
- Use Serializable::create to specialize for creating cacheable objects.
- Use CacheableKey::create to specialize for creating key objects.


cppcache/include/geode/CacheableString.hpp
```diff
@@ -167,46 +166,86 @@ class _GEODE_EXPORT CacheableString : public CacheableKey {
   static bool isAscii(const std::string& str);
 };
 
-inline std::shared_ptr<CacheableKey> createKeyArr(const std::string& value) {
+template <>
+inline std::shared_ptr<CacheableKey> CacheableKey::create(const char* value) {
```
dgkimura: It looks like the only difference between these template specializations is the argument.  Would it make sense to create a type_trait for `is_char` and do some `enable_if` logic here?

cppcache/include/geode/Cacheable.hpp
```diff
@@ -34,12 +34,6 @@ namespace client {
 
 typedef Serializable Cacheable;
```
dgkimura: Is there any point to keeping this file around now?

cppcache/test/SerializableCreateTests.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include <gtest/gtest.h>
+
+#include <geode/Serializable.hpp>
+#include <geode/CacheableString.hpp>
+#include <geode/CacheableBuiltins.hpp>
+#include <geode/CacheableDate.hpp>
+
+using namespace apache::geode::client;
+
+TEST(SerializableCreateTests, forArrayOf_constchar) {
```
dgkimura: Yay!

cppcache/include/geode/Cacheable.hpp
pivotal-jbarrett: We could totally move this into Serializable.hpp at this point. I'll write up another ticket to investigate the best solution. 

https://issues.apache.org/jira/browse/GEODE-4356

cppcache/include/geode/CacheableString.hpp
pivotal-jbarrett: It would. I spent some time on it and could find the right combination. Do you want to have a crack at it. Changing it later won't be an issue since it will still expose all the same string types.

cppcache/include/geode/Region.hpp
```diff
@@ -380,22 +380,22 @@ class _GEODE_EXPORT Region : public std::enable_shared_from_this<Region> {
...
   template <class KEYTYPE>
   inline void put(const KEYTYPE& key, const std::shared_ptr<Cacheable>& value,
                   const std::shared_ptr<Serializable>& arg = nullptr) {
-    put(createKey(key), value, arg);
+    put(CacheableKey::create(key), value, arg);
   }
 
   /** Convenience method allowing value to be a const char* */
   template <class VALUETYPE>
   inline void put(const std::shared_ptr<CacheableKey>& key,
```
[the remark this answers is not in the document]
pivotal-jbarrett: Let's investigate this a little more. I am not excited at all by the templates that were here and tried to clean it up by removing the 4 namespaced functions that created cacheables. Because of matching rules they don't really match what you would expect. This particular function you commented on only matches if the second argument is actually of type `std::shared_ptr<Cacheable>` and will not match any of the derivatives of `Cacheable`. Almost all the calls end up matching the `KEYTYPE`, `VALUETYPE` template.

PR #190, opened 2018-01-23
GEODE-4314: Removes std::enable_shared_from_this.

pivotal-jbarrett:
- Adds std::enable_shared_from_this to PdxSerializable to satisfy some
  strange serialization behavior. See GEODE-4359

### DIFF
--- a/cppcache/src/PdxHelper.cpp
+++ b/cppcache/src/PdxHelper.cpp
@@ -14,12 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/*
- * PdxHelper.cpp
- *
- *  Created on: Dec 13, 2011
- *      Author: npatel
- */
 
 #include <geode/Cache.hpp>
 #include <geode/DataInput.hpp>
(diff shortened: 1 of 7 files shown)


cppcache/src/PdxHelper.cpp
```diff
@@ -53,10 +47,8 @@ PdxHelper::~PdxHelper() {}
 
 void PdxHelper::serializePdx(DataOutput& output,
                              const PdxSerializable& pdxObject) {
-  serializePdx(
-      output,
-      std::static_pointer_cast<PdxSerializable>(
-          std::const_pointer_cast<Serializable>(pdxObject.shared_from_this())));
+  serializePdx(output, std::const_pointer_cast<PdxSerializable>(
+                           pdxObject.shared_from_this()));
```
dgkimura: Hmm.. something feels awkward about this.  Looks like we assume `pdxObject` is a smart pointer allocated on the heap w/control block.  If we want to increment ref count in this call would it make more sense to pass a shared_ptr as argument instead of explicitly calling `shared_from_this()`?

pivotal-jbarrett: Completely agree! I’d say we should pair up on GEODE-4359 and dig in

PR #197, opened 2018-01-29
 GEODE-4416: Converts CqState to enum class.

pivotal-jbarrett:


cppcache/src/TcrMessage.hpp
```diff
@@ -50,12 +50,15 @@
...
 namespace geode {
 namespace client {
+
 class TcrMessage;
 class ThinClientRegion;
 class ThinClientBaseDM;
 class TcrMessageHelper;
 class TcrConnection;
 class TcrMessagePing;
+enum class CqState;
```
dgkimura: Why does this need to be forward declared?

cppcache/src/TcrMessage.cpp
```diff
@@ -2332,7 +2332,7 @@ TcrMessageExecuteCq::TcrMessageExecuteCq(std::unique_ptr<DataOutput> dataOutput,
   writeStringPart(str1);  // cqName
   writeStringPart(str2);  // query string
 
-  writeIntPart(state);  // cq state
+  writeIntPart(static_cast<int32_t>(state));  // cq state
```
dgkimura: Almost seems there should be a `writeCqStatePart(CqState state)` or a `writePart(CqState state)`?

cppcache/src/TcrMessage.hpp
pivotal-jbarrett: This header does include, nor does it need to include the enum. I suppose it wouldn't hurt, but it is one less include.

cppcache/src/TcrMessage.cpp
pivotal-jbarrett: It looks like `enum` automatically converts to it underlying type, in this case `int` but `enum class` you have to explicitly cast to its underlying type to get the same result. Nothing changed at this call spot other than adding the explicit cast.

dgkimura: I feel like a big part of `enum class` is type safety.  By using `static_cast` we're overriding that safety.  It's not a big deal, but this smells a little fishy.

pivotal-jbarrett: http://en.cppreference.com/w/cpp/language/enum

It looks like the ambiguity and some of the nastiness of enum melts away if you explicitly define the integral type the enum extends. I will see what this looks like by extending int32_t explicitly. Though at this particular call we want to guarantee that a 32 bit value is serialized.

pivotal-jbarrett: So explicitly making this `enum` extend `int32_t` does not remove the need to cast to `int32_t` here since, inline with the documentation, scoped `enum`s to do not implicitly convert to integral types but may be explicitly converted using `static_cast`.

dgkimura: If you think it makes sense, I'm fine with keeping the `static_cast`.  It was just fishy to me that we're treating the `enum class` like an `int` and calling `writeIntPart`, but it really isn't an `int` and shouldn't be treated as one.

PR #198, opened 2018-01-29
 GEODE-4413: Converts CqOperation to enum class.

pivotal-jbarrett:


clicache/src/CqOperation.hpp
```diff
@@ -15,34 +15,22 @@
...
-
       /// <summary>
-      /// Enumerated type for CqOperationType
+      /// Enumerated type for CqOperation
       /// </summary>
-      public enum class CqOperationType
+      public enum class CqOperation
       {
-	OP_TYPE_INVALID = -1,
+	      OP_TYPE_INVALID = -1,
```
dgkimura: Oops

pivotal-jbarrett: Oops, good find.

PR #200, opened 2018-01-30
Geode 4335: Custom serialization example

mhansonp:


mhansonp:
request merge to the wrong base...

PR #201, opened 2018-01-30
Geode 4335: Custom serializer example

mhansonp:


mhansonp:
Changed PR title.

mhansonp:
I may have misunderstood, I can accept that.  So we were supposed to implement a PdxSerializable, not PdxSerializer.... Not a big deal. We will fix it.

igodwin:
PdxSerializable example pushed.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+class Order : public PdxSerializable {
+ public:
+  Order() {
+    orderID = (unsigned long)0;
+    quantity = (unsigned int)0;
+    name = "";
+  };
+
+  Order(unsigned long inOrderID, const std::string& inName,
```
pivotal-jbarrett: Additionally consider adding another constructor that accepts an rvalue. This allows the caller to avoid copy construction on std::string.
```
Order(unit32_t orderID, std::string&& name, uint32_t quantity)
    : orderID(orderID), name(std::move(name)), quantity(quantity) {}
```

pivotal-jbarrett: Used C++11 types: `uint32_t`, etc.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+        unsigned int inQuantity)
+      : orderID(inOrderID), name(inName), quantity(inQuantity){};
+
+  virtual ~Order(){};
+
+  unsigned long getOrderID() { return orderID; }
```
pivotal-jbarrett: `uint32_t`

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+  const std::string& getName() const { return name; }
+  unsigned int getQuantity() { return quantity; }
+
+  virtual void fromData(PdxReader& pdxReader) override {
+    try {
+      orderID = (unsigned long)pdxReader.readLong(orderid_key);
```
pivotal-jbarrett: Do not use C-style casts in C++11 source. Use `static_cast<uint32_t>()`.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+      name = pdxReader.readString(name_key);
+      quantity = (unsigned int)pdxReader.readInt(quantity_key);
+    } catch (std::exception& e) {
+      std::cout << "Caught exception: " << e.what() << std::endl;
+      return;
```
pivotal-jbarrett: Bad programming practice to eat exceptions. This example is meant to show best practices.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+    return;
+  }
+
+  static PdxSerializable* createDeserializable() { return new Order(); }
+
+  virtual std::string toString() const override {
+    std::stringstream ss("");
+    ss << " OrderID: " << orderID << " Product Name: " << name
+       << " Quantity: " << quantity;
+    return ss.str();
```
pivotal-jbarrett: This string is pretty simple, you may want to avoid the overhead of `std::stringstream`:
```
return std::string("Order ID: ") + std::to_string(orderId) + " Product Name: " + name + " Quantity: " + std::string(quantity);
```

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  virtual size_t objectSize() const override {
+    auto objectSize = sizeof(Order);
+    return objectSize;
```
pivotal-jbarrett: Does not account for the memory allocated by `name`.
```
objectSize += name.capacity();
```

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  const std::string& getClassName() const override {
+    static std::string className = "com.example.Order";
```
pivotal-jbarrett: static const

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+    return className;
+  }
+
+ private:
+  unsigned long orderID;
+  std::string name;
+  unsigned int quantity;
+
+  const std::string orderid_key = "orderid";
```
pivotal-jbarrett: Constant values should be ALL CAPS and static.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+int main(int argc, char** argv) {
+
+  auto cacheFactory = CacheFactory();
+  cacheFactory.set("log-level", "none");
+  auto cache = cacheFactory.create();
+
+  auto poolFactory = cache.getPoolManager().createFactory();
+  poolFactory->addLocator("localhost", 10334);
+  auto pool = poolFactory->create("pool");
+  auto regionFactory = cache.createRegionFactory(PROXY);
```
pivotal-jbarrett: Put some whitespace after pool setup.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+  auto region = regionFactory.setPoolName("pool").create("custom_orders");
+
+  cache.getTypeRegistry().registerPdxType(Order::createDeserializable);
+
+  std::cout << "Create orders" << std::endl;
+  auto order1 = std::shared_ptr<Order>(new Order(1, "product x", 23));
+  auto order2 = std::shared_ptr<Order>(new Order(2, "product y", 37));
+
+  std::cout << "Storing orders in the region" << std::endl;
+  auto keyptr = CacheableKey::create("Customer1");
```
pivotal-jbarrett: You don't need to do this. Use the same example below using the raw string.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+  region->put(keyptr, order1);
+  region->put("Customer2", order2);
+
+  std::cout << "Getting the orders from the region" << std::endl;
+  auto keyptr = CacheableKey::create("Customer1");
+  auto order = region->get(keyptr);
```
pivotal-jbarrett: Use:
```
auto order = region->get("Customer1");
```

examples/cpp/customserializer/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
pivotal-jbarrett: Leave customer serializer example out of this pull and ticket.

examples/cpp/customserializable/main.cpp
dgkimura: Another option is universal reference and std::forward.  Then you only need 1 constructor.  

```cpp
template <typename T>
Order(uint32_t orderId, T&& name, uint32_t quantity)
  : orderId(orderid), name(std::forward(name), quantity(quantity) {}
```

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+#include <geode/CacheFactory.hpp>
+#include <geode/PoolManager.hpp>
+#include <geode/PdxWrapper.hpp>
+
+using namespace apache::geode::client;
+
+class Order : public PdxSerializable {
+ public:
+  Order() {
```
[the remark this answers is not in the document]
dgkimura: +1
No point in calling default constructor and then turning around and calling assignment operator.

examples/cpp/customserializable/main.cpp
dgkimura: +1

examples/cpp/customserializable/main.cpp
dgkimura: +1

examples/cpp/customserializable/main.cpp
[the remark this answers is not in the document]
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: We ended up introducing the rvalue reference constructor to keep things somewhat simple in the example.

mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.  this one was interesting - couldn't find great info when we searched for the reasoning but may be searching wrong keywords?

examples/cpp/customserializable/main.cpp
mcmellawatt: done

examples/cpp/customserializable/main.cpp
mcmellawatt: is this to be consistent with a prior styling choice?  google style guide says to prefix consts (including static consts) with a 'k' followed by mixed case e.g. kOrderIdKey.

examples/cpp/customserializable/main.cpp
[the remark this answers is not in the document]
mcmellawatt: done.  we also changed our class members to match google style recommendations with an underscore suffix.

https://google.github.io/styleguide/cppguide.html#Variable_Names

examples/cpp/customserializable/main.cpp
mcmellawatt: so these are class members and we ran into a compiler error when making them static.

`/opt/local/geode-native/examples/cpp/customserializable/main.cpp:75:28: error: static data member of type 'const std::string' (aka 'const basic_string<char, char_traits<char>, allocator<char> >') must be initialized out of line

static const std::string ORDER_ID_KEY_ = "order_id";`

We did a search and found this, which makes it seem like without a separate hpp file that this would end up being awkward anyways:

https://stackoverflow.com/questions/11693690/const-array-declaration-in-c-header-file

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializer/CMakeLists.txt
mcmellawatt: will do.

mcmellawatt:
Requested changes added

examples/cpp/customserializable/main.cpp
pivotal-jbarrett: Yes... C++ is fun. You will need the header to declare your class and a source file to define it. This is all best practice. I should have caught that your class was defined in your main source file.

examples/cpp/customserializable/main.cpp
pivotal-jbarrett: `sizeof(type)` gives you the size of the `type`. So in this case you will get the size of 2 `uint32_t` and `std::string`. The `sizeof(std::string)` is is roughly the size of the `size_t` and 'char*`. To get the length of the memory allocated for the `char*` we have to ask `std::string` for its current allocation, `capacity`.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,111 @@
...
+    return objectSize;
+  }
+
+  const std::string& getClassName() const override {
+    static const std::string class_name = "com.example.Order";
+    return class_name;
+  }
+
+ private:
+  uint32_t order_id_;
```
pivotal-jbarrett: What is up with suffixing member variables with `_`?

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,111 @@
...
+  std::string name_;
+  uint16_t quantity_;
+
+  const std::string ORDER_ID_KEY_ = "order_id";
```
pivotal-jbarrett: Make static. Will need to create an Order.hpp/cpp file to be complete.

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,111 @@
...
+  poolFactory->addLocator("localhost", 10334);
+  auto pool = poolFactory->create("pool");
+
+  auto regionFactory = cache.createRegionFactory(PROXY);
+  auto region = regionFactory.setPoolName("pool").create("custom_orders");
+
+  cache.getTypeRegistry().registerPdxType(Order::createDeserializable);
+
+  std::cout << "Create orders" << std::endl;
+  auto order1 = std::shared_ptr<Order>(new Order(1, "product x", 23));
```
pivotal-jbarrett: Use:
```c++
auto order1 = std::make_shared<Order>(1, "product x", 23);
```

examples/cpp/customserializable/main.cpp
pivotal-jbarrett: I like @dgkimura solution even better!

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <geode/PoolManager.hpp>
+#include <geode/PdxWrapper.hpp>
+
+using namespace apache::geode::client;
+
+class Order : public PdxSerializable {
+ public:
+  Order() : order_id_(0), name_(), quantity_(0) {};
+
+  Order(uint32_t order_id, std::string&& name, uint16_t quantity)
```
dgkimura: By doing it this way you're making it illegal to pass name by lvalue.  Is that intentional?

examples/cpp/customserializable/main.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <geode/CacheFactory.hpp>
+#include <geode/PoolManager.hpp>
+#include <geode/PdxWrapper.hpp>
+
+using namespace apache::geode::client;
+
+class Order : public PdxSerializable {
+ public:
+  Order() : order_id_(0), name_(), quantity_(0) {};
```
dgkimura: Any reason not to use _delegating constructor_ here?  Then if the constructor ever has extra logic, it only needs to exist in one place.

pivotal-jbarrett: 👍

examples/cpp/customserializable/main.cpp
pivotal-jbarrett: 👍 good eye, I totally missed that they got rid of the l-value constructor.

mhansonp: I did that because it wasn't being used in the example and to keep unused code to a minimum, but this shouldn't matter anyways after we switch to the universal reference constructor.

mcmellawatt: I did that because it wasn't being used in the example and to keep unused code to a minimum, but this shouldn't matter anyways after we switch to the universal reference constructor.

dgkimura: Perhaps this is a side topic, but how much C++11 knowledge do we expect of our users for barrier to entry?  Given that we are a Java first product I would think not much.  If that's the case then I would vote to use your original constructor even though it invokes an extra copy for a few reasons:

1) I think reference semantics is confusing for a novice C++11 developer and using them could detract from the core example we are trying to convey here.
2) If our user is a more advanced C++11 developer then they should already know how to do this themselves.
3) I think there is _sometimes_ a use case for invoking an extra copy.  Perhaps for readability and code that is not executed in a tight loop it may actually be the right way to go.

examples/cpp/customserializable/main.cpp
mcmellawatt: yeah agreed its weird, but I vote for just sticking to the style guide whenever possibility for consistency.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
mcmellawatt: done.

examples/cpp/customserializable/main.cpp
pivotal-jbarrett: Good points David. 👍

mcmellawatt: Thanks!  Ended up going back with the original constructor for simplicity as suggested.

mcmellawatt:
This is ready for review after Travis finishes.

PR #202, opened 2018-01-31
GEODE-4426: Refactor PdxFieldTypes as enum class.

pivotal-jbarrett:
- Removes unused readCollection, which had an undefined PDX field type.

### DIFF
--- a/cppcache/include/geode/PdxFieldTypes.hpp
+++ b/cppcache/include/geode/PdxFieldTypes.hpp
@@ -20,39 +20,36 @@
 #ifndef GEODE_PDXFIELDTYPES_H_
 #define GEODE_PDXFIELDTYPES_H_
 
-#include "internal/geode_base.hpp"
-
 namespace apache {
 namespace geode {
 namespace client {
 
-class _GEODE_EXPORT PdxFieldTypes {
- public:
-  enum PdxFieldType {
-    BOOLEAN,
-    BYTE,
-    CHAR,
-    SHORT,
-    INT,
-    LONG,
-    FLOAT,
-    DOUBLE,
-    DATE,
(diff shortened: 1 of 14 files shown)


cppcache/include/geode/PdxFieldTypes.hpp
```diff
@@ -20,39 +20,36 @@
...
-    SHORT_ARRAY,
-    INT_ARRAY,
-    LONG_ARRAY,
-    FLOAT_ARRAY,
-    DOUBLE_ARRAY,
-    STRING_ARRAY,
-    OBJECT_ARRAY,
-    ARRAY_OF_BYTE_ARRAYS
-  };
+enum class PdxFieldTypes : int8_t {
```
dgkimura: Hmm... I'm not sure how I feel about this approach.  Now the only real benefit of `enum class` is scoping?  I would actually prefer to *not* inherit and instead call `static_cast` like you did in previous PR.  That way at least the programmer must write `static_cast` if he or she wants to treat it as an `int8_t`.  And at that point all bets are off and assumably programmer knows what he or she is doing.

pivotal-jbarrett: The `static_cast` will always be required to cast to and integral type even with the type specified. What specifying the type allows is us to initialize it from an integral type without explicit `static_cast`.

I felt that making it `int8_t` kept it consistent with the original value which was stored as the integral value `int8_t`.

dgkimura: Interesting.  I didn't know about this feature.  In that case I revoke my previous comments.  :)  I had thought it was creating an "is-a" relationship and you'd lose type safety.

PR #203, opened 2018-01-31
 GEODE-4448: Refactors ExpirationAction to enum class.

pivotal-jbarrett:


tests/cpp/fwklib/RegionHelper.hpp
```diff
@@ -85,10 +83,25 @@ class RegionHelper {
     return sString;
   }
 
+  static std::string to_string(ExpirationAction expirationAction) {
```
dgkimura: Yes! I'm glad to see this approach to `to_string(...)` instead of putting it on the class itself.

PR #204, opened 2018-02-05
 GEODE-4572: Refactor createDeserialized with std::shared_ptr.

pivotal-jbarrett:


clicache/src/impl/DelegateWrapper.hpp
```diff
@@ -80,16 +80,18 @@ namespace Apache
...
         /// delegate and wrapping inside a <c>ManagedCacheableKey</c> object.
         /// </returns>
-        apache::geode::client::Serializable* NativeDelegateGeneric( )
+        std::shared_ptr<apache::geode::client::Serializable> NativeDelegateGeneric( )
         {
           auto tempObj = m_delegate( );
           if(auto tempDelta = dynamic_cast<IGeodeDelta^>(tempObj))
           {
-            return new apache::geode::client::ManagedCacheableDeltaGeneric(tempDelta);
+            return std::shared_ptr<apache::geode::client::ManagedCacheableDeltaGeneric>(
```
dgkimura: Did `std::make_shared` not work here?

quickstart/cpp/queryobjects/PortfolioPdx.hpp
```diff
@@ -90,7 +90,9 @@ class TESTOBJECT_EXPORT PortfolioPdx : public PdxSerializable {
 
   int8_t* getArrayZeroSize() { return arrayZeroSize; }
 
-  static PdxSerializable* createDeserializable() { return new PortfolioPdx(); }
+  static std::shared_ptr<PdxSerializable> createDeserializable() {
+    return new PortfolioPdx();
```
dgkimura: Hmm.. `std::make_shared`?

quickstart/cpp/queryobjects/PositionPdx.hpp
```diff
@@ -90,7 +90,9 @@ class TESTOBJECT_EXPORT PositionPdx
 
   int32_t getSharesOutstanding() { return sharesOutstanding; }
 
-  static PdxSerializable* createDeserializable() { return new PositionPdx(); }
+  static std::shared_ptr<PdxSerializable> createDeserializable() {
+    return new PositionPdx();
```
dgkimura: `std::make_shared`?

tests/cpp/testobject/PdxClassV2.hpp
```diff
@@ -584,7 +594,7 @@ class TESTOBJECT_EXPORT TestEqualsV1 : public PdxSerializable {
     return "PdxTests::TestEquals";
   }
 
-  static PdxSerializable* createDeserializable() {
+  static std::shared_ptr<PdxSerializable> createDeserializable() {
     return new TestEqualsV1();
```
dgkimura: `std::make_shared`?

clicache/src/Serializable.cpp
```diff
@@ -544,7 +544,7 @@ namespace Apache
 
         _GF_MG_EXCEPTION_TRY2
           CacheImpl *cacheImpl = CacheRegionHelper::getCacheImpl(cache->GetNative().get());
-          cacheImpl->getSerializationRegistry()->addType((native::Serializable*(*)())System::Runtime::InteropServices::Marshal::GetFunctionPointerForDelegate(nativeDelegate).ToPointer());
+          cacheImpl->getSerializationRegistry()->addType((std::shared_ptr<native::Serializable>(*)())System::Runtime::InteropServices::Marshal::GetFunctionPointerForDelegate(nativeDelegate).ToPointer());
```
dgkimura: Could we do a c++ style cast to `TypeFactoryMethod`?  There's a couple more you touched in this file where we might be able to make the change.

pivotal-jbarrett: I don't think so, TypeFactoryMethod is defined in terms of std::function but maybe.

clicache/src/impl/DelegateWrapper.hpp
pivotal-jbarrett: No, common problem we ran into in C++/CLR. If a parameter to the constructor is a GC reference or pointer the `std::forward` in the `std::shared_ptr` constructor fails.

quickstart/cpp/queryobjects/PortfolioPdx.hpp
pivotal-jbarrett: Good eye, must not be compiled as part of the build because that is definitely a compiler failure.

PR #207, opened 2018-02-06
Feature/geode 4424

mmartell:
Refactor PoolManager::createFactory to return a PoolFactory instead of a pointer to a PoolFactory.


mmartell:
Incorrect pull request

PR #208, opened 2018-02-06
GEODE-4424: Refactor PoolManager::createFactory to return value

mmartell:
Testing: Unit/integration tests passed on Mac and Windows

### DIFF
--- a/clicache/src/Cache.cpp
+++ b/clicache/src/Cache.cpp
@@ -339,7 +339,8 @@ namespace Apache
 
         PoolFactory^ Cache::GetPoolFactory()
         {
-          return PoolFactory::Create(m_nativeptr->get_shared_ptr()->getPoolManager().createFactory());
+          return PoolFactory::Create(std::make_shared<native::PoolFactory>(
+            m_nativeptr->get_shared_ptr()->getPoolManager().createFactory()));
         }
 
         PoolManager^ Cache::GetPoolManager()
--- a/clicache/src/Cache.cpp
+++ b/clicache/src/Cache.cpp
@@ -339,8 +339,8 @@ namespace Apache
 
         PoolFactory^ Cache::GetPoolFactory()
         {
-          return PoolFactory::Create(std::make_shared<native::PoolFactory>(
-            m_nativeptr->get_shared_ptr()->getPoolManager().createFactory()));
+          return PoolFactory::Create(std::unique_ptr<native::PoolFactory>(new native::PoolFactory(
+            m_nativeptr->get_shared_ptr()->getPoolManager().createFactory())));
         }
 
         PoolManager^ Cache::GetPoolManager()
--- a/clicache/src/PoolManager.cpp
+++ b/clicache/src/PoolManager.cpp
@@ -37,7 +37,8 @@ namespace Apache
 
       PoolFactory^ PoolManager::CreateFactory()
       {
-        return PoolFactory::Create(m_nativeref.createFactory());
+        return PoolFactory::Create(std::make_shared<native::PoolFactory>(
+          m_nativeref.createFactory()));
       }
 
       const Dictionary<String^, Pool^>^ PoolManager::GetAll()
--- a/clicache/src/PoolManager.cpp
+++ b/clicache/src/PoolManager.cpp
@@ -37,8 +37,8 @@ namespace Apache
 
       PoolFactory^ PoolManager::CreateFactory()
       {
-        return PoolFactory::Create(std::make_shared<native::PoolFactory>(
-          m_nativeref.createFactory()));
+        return PoolFactory::Create(std::unique_ptr<native::PoolFactory>(
+          new native::PoolFactory(m_nativeref.createFactory())));
       }
 
       const Dictionary<String^, Pool^>^ PoolManager::GetAll()
--- a/cppcache/include/geode/PoolFactory.hpp
+++ b/cppcache/include/geode/PoolFactory.hpp
@@ -533,9 +533,10 @@ class _GEODE_EXPORT PoolFactory {
 
   ~PoolFactory();
 
+  PoolFactory(const PoolFactory&);
+
  private:
   PoolFactory(const Cache& cache);
-  PoolFactory(const PoolFactory&);
   PoolFactory& addCheck(const std::string& host, int port);
   std::shared_ptr<PoolAttributes> m_attrs;
   bool m_isSubscriptionRedundancy;
@@ -544,6 +545,7 @@ class _GEODE_EXPORT PoolFactory {
 
   friend class Cache;
   friend class PoolManager;
+  friend class PoolManagerImpl;
   friend class CacheFactory;
   friend class CacheXmlCreation;
 
--- a/cppcache/include/geode/PoolFactory.hpp
+++ b/cppcache/include/geode/PoolFactory.hpp
@@ -531,9 +531,9 @@ class _GEODE_EXPORT PoolFactory {
    */
   PoolFactory& setPRSingleHopEnabled(bool enabled);
 
-  ~PoolFactory();
+  ~PoolFactory() = default;
 
-  PoolFactory(const PoolFactory&);
+  PoolFactory(const PoolFactory&) = default;
 
  private:
   PoolFactory(const Cache& cache);
--- a/cppcache/integration-test/CacheHelper.cpp
+++ b/cppcache/integration-test/CacheHelper.cpp
@@ -134,7 +134,7 @@ CacheHelper::CacheHelper(const std::shared_ptr<Properties>& configPtr,
   auto cacheFactory = CacheFactory(pp);
   cachePtr = std::make_shared<Cache>(cacheFactory.create());
 
-  auto poolFactory = cachePtr->getPoolManager().createFactory();
+  auto poolFactory = std::make_shared<PoolFactory>(cachePtr->getPoolManager().createFactory());
   addServerLocatorEPs("localhost:40404", poolFactory);
   poolFactory->create("__CACHE_HELPER_POOL__");
 
@@ -233,7 +233,7 @@ CacheHelper::CacheHelper(const bool isthinClient, const char* poolName,
     auto cacheFac = CacheFactory(pp);
     cachePtr = std::make_shared<Cache>(cacheFac.create());
 
-    auto poolFactory = cachePtr->getPoolManager().createFactory();
+    auto poolFactory = std::make_shared<PoolFactory>(cachePtr->getPoolManager().createFactory());
 
     poolFactory->setPRSingleHopEnabled(prSingleHop);
     poolFactory->setThreadLocalConnections(threadLocal);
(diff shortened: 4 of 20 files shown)


clicache/src/Cache.cpp
```diff
@@ -339,7 +339,8 @@ namespace Apache
 
         PoolFactory^ Cache::GetPoolFactory()
         {
-          return PoolFactory::Create(m_nativeptr->get_shared_ptr()->getPoolManager().createFactory());
+          return PoolFactory::Create(std::make_shared<native::PoolFactory>(
```
pivotal-jbarrett: The .NET factory does not need to keep a shared pointer to the C++ factory. It is a 1:1 relationship now so it should keep a unique_ptr. See other factories that have gone through this change for examples.

clicache/src/PoolManager.cpp
```diff
@@ -37,7 +37,8 @@ namespace Apache
 
       PoolFactory^ PoolManager::CreateFactory()
       {
-        return PoolFactory::Create(m_nativeref.createFactory());
+        return PoolFactory::Create(std::make_shared<native::PoolFactory>(
```
pivotal-jbarrett: Same, use unique_ptr.

cppcache/include/geode/PoolFactory.hpp
```diff
@@ -533,9 +533,10 @@ class _GEODE_EXPORT PoolFactory {
 
   ~PoolFactory();
 
+  PoolFactory(const PoolFactory&);
```
pivotal-jbarrett: If the copy is trivial, as it should be just default, inline and include I header.

cppcache/include/geode/PoolFactory.hpp
```diff
@@ -544,6 +545,7 @@ class _GEODE_EXPORT PoolFactory {
 
   friend class Cache;
   friend class PoolManager;
+  friend class PoolManagerImpl;
```
pivotal-jbarrett: Why this friend?

cppcache/integration-test/CacheHelper.cpp
```diff
@@ -134,7 +134,7 @@ CacheHelper::CacheHelper(const std::shared_ptr<Properties>& configPtr,
   auto cacheFactory = CacheFactory(pp);
   cachePtr = std::make_shared<Cache>(cacheFactory.create());
 
-  auto poolFactory = cachePtr->getPoolManager().createFactory();
+  auto poolFactory = std::make_shared<PoolFactory>(cachePtr->getPoolManager().createFactory());
```
pivotal-jbarrett: Is the pool factory assigned to a member variable later? If not then we don’t need the shared_ptr.

cppcache/integration-test/testPoolFactory.cpp
```diff
@@ -29,7 +29,7 @@ BEGIN_TEST(POOLFACTORY)
 {
   auto cacheFactory = CacheFactory();
   auto cache = cacheFactory.create();
-  auto poolFactory = cache.getPoolManager().createFactory();
+  auto poolFactory = std::make_shared<PoolFactory>(cache.getPoolManager().createFactory());
```
pivotal-jbarrett: If this is only a local variable then you don’t need to make it shared_ptr.

cppcache/src/PoolFactory.cpp
```diff
@@ -69,6 +69,12 @@ PoolFactory::PoolFactory(const Cache& cache)
       m_addedServerOrLocator(false),
       m_cache(cache) {}
 
+PoolFactory::PoolFactory(const PoolFactory& factory)
```
pivotal-jbarrett: This looks like default impl, use that instead.

clicache/src/Cache.cpp
dgkimura: Thanks. Agreed that unique_ptr is better here.

cppcache/include/geode/PoolFactory.hpp
dgkimura: Yup. Done.

cppcache/include/geode/PoolFactory.hpp
dgkimura: `PoolManagerImpl::createFactory` calls into `PoolFactory` constructor and `PoolFactory` constructor is private.  I've wondered is if these constructors even need to be private anymore?  I assumed it used to be to protect the original singleton design.  Perhaps this should be a separate ticket to investigate.

cppcache/integration-test/CacheHelper.cpp
dgkimura: It is passed around later on through method calls. We had kept it as shared_ptr to minimize our code changes.  But your requested change isn't that bad so it has been updated in next revision.

cppcache/include/geode/PoolFactory.hpp
pivotal-jbarrett: Ok. It is more likely private to require the use of the factory method. Just odd to expose internal classes in a public header like this. Should be ok for now.

cppcache/integration-test/testPoolFactory.cpp
```diff
@@ -30,9 +30,8 @@ BEGIN_TEST(POOLFACTORY)
   auto cacheFactory = CacheFactory();
   auto cache = cacheFactory.create();
   auto poolFactory = cache.getPoolManager().createFactory();
-  ASSERT(poolFactory != nullptr, "poolFactory was a nullptr");
-  auto& testPoolFactory = poolFactory->setSubscriptionEnabled(true);
-  bool test = poolFactory.get() == std::addressof<PoolFactory>(testPoolFactory);
+  auto& testPoolFactory = poolFactory.setSubscriptionEnabled(true);
+  bool test = &poolFactory == std::addressof<PoolFactory>(testPoolFactory);
```
pivotal-jbarrett: Do you think this test is really valuable anymore? Maybe PoolFactory should implement `operator==()` and we can compare them this way?

cppcache/integration-test/testThinClientCqDurable.cpp
```diff
@@ -192,7 +192,7 @@ void RunDurableCqClient() {
   // Create a Geode Cache Programmatically.
   auto cacheFactory = CacheFactory(pp);
   auto cachePtr = std::make_shared<Cache>(cacheFactory.create());
-  auto poolFactory = cachePtr->getPoolManager().createFactory();
+  auto poolFactory = std::make_shared<PoolFactory>(cachePtr->getPoolManager().createFactory());
```
pivotal-jbarrett: Is `shared_ptr` necessary?

cppcache/integration-test/testThinClientPoolAttrTest.cpp
```diff
@@ -69,7 +69,7 @@ class putThread : public ACE_Task_Base {
 };
 
 void doAttrTestingAndCreatePool(const char* poolName) {
-  auto poolFacPtr = getHelper()->getCache()->getPoolManager().createFactory();
+  auto poolFacPtr = std::make_shared<PoolFactory>(getHelper()->getCache()->getPoolManager().createFactory());
```
pivotal-jbarrett: Is `shared_ptr` necessary?

cppcache/integration-test/testThinClientPoolAttrTest.cpp
```diff
@@ -204,7 +204,7 @@ DUNIT_TASK(CLIENT1, StartC1)
    doAttrTestingAndCreatePool(poolName);
 
    // Do PoolCreation testing , create another pool with same name
-   auto poolFacPtr = getHelper()->getCache()->getPoolManager().createFactory();
+   auto poolFacPtr = std::make_shared<PoolFactory>(getHelper()->getCache()->getPoolManager().createFactory());
```
pivotal-jbarrett: Is `shared_ptr` necessary?

cppcache/integration-test/testThinClientSecurityMultiUserTest.cpp
```diff
@@ -47,7 +47,7 @@ END_TASK_DEFINITION
 DUNIT_TASK_DEFINITION(CLIENT1, PerformSecureOperationsWithUserCredentials)
   {
     auto cache = CacheFactory().create();
-    auto poolFactory = cache.getPoolManager().createFactory();
+    auto poolFactory = std::make_shared<PoolFactory>(cache.getPoolManager().createFactory());
```
pivotal-jbarrett: Is `shared_ptr` necessary?

cppcache/src/CacheXmlParser.cpp
```diff
@@ -597,7 +597,7 @@ void CacheXmlParser::startPool(const xmlChar** atts) {
         "A pool cannot be created without a name";
     throw CacheXmlException(s.c_str());
   }
-  auto factory = m_cache->getPoolManager().createFactory();
+  auto factory = std::make_shared<PoolFactory>(m_cache->getPoolManager().createFactory());
```
pivotal-jbarrett: Is `shared_ptr` necessary?

tests/cpp/fwklib/FrameworkTest.cpp
```diff
@@ -338,7 +338,7 @@ void FrameworkTest::localDestroyRegion(std::shared_ptr<Region>& region) {
 void FrameworkTest::parseEndPoints(int32_t ep, std::string label,
                                    bool isServer) {
   std::string poolName = "_Test_Pool";
-  auto pfPtr = m_cache->getPoolManager().createFactory();
+  auto pfPtr = std::make_shared<PoolFactory>(m_cache->getPoolManager().createFactory());
```
pivotal-jbarrett: Is `shared_ptr` necessary?

tests/cpp/fwklib/FwkObjects.cpp
```diff
@@ -573,7 +573,7 @@ FwkPool::FwkPool(const DOMNode* node) : m_locators(false), m_servers(false) {
   // Init Factory
   m_cache = std::make_shared<Cache>(CacheFactory().create());
   m_poolManager = &m_cache->getPoolManager();
-  m_poolFactory = m_poolManager->createFactory();
+  m_poolFactory = std::make_shared<PoolFactory>(m_poolManager->createFactory());
```
pivotal-jbarrett: Is `shared_ptr` necessary?

cppcache/integration-test/testPoolFactory.cpp
dgkimura: I had wondered this as well.  My interpretation is that this test is trying to assert that the fluent interface is returning the exact same object which is probably not how one would usually expect `operator==()` to work.  I am fine deleting this test if you agree that it has minimal value.

cppcache/src/CacheXmlParser.cpp
dgkimura: Yes.  In this case `shared_ptr` is providing abstraction to push objects onto a `std::stack<std::shared_ptr<void>>` _stack.

cppcache/integration-test/testThinClientCqDurable.cpp
dgkimura: Not necessary. Just minimal change to get the code working.  Probably best to keep test code up-to-date so I'll see about making suggested changes to remove `shared_ptr` in this context in integration-tests.

cppcache/integration-test/testPoolFactory.cpp
pivotal-jbarrett: Oh.. I misread what it was doing. Yes leave it.
👍

PR #209, opened 2018-02-06
Feature/geode 4507 - Custom PdxSerializer Example

moleske:


examples/cpp/customserializer/Order.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+  ~Order() = default;
+
+  inline uint32_t getOrderID() { return order_id_; }
+
+  inline const std::string &getName() const { return name_; }
+
+  inline uint16_t getQuantity() { return quantity_; }
+
+  size_t getSize() const;
```
pivotal-jbarrett: Why not line these last two methods?

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,93 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "OrderSerializer.hpp"
+#include "Order.hpp"
+
+namespace customserializer {
+
+void *OrderSerializer::fromData(const std::string &className,
```
pivotal-jbarrett: Preference is */& is stuck to type and not name for all new sources. 
```c++
void* OrderSerializer::fromData(const std::string& className,
                                PdxReader& pdxReader) {
```

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,93 @@
...
+                                PdxReader &pdxReader) {
+  if (className != Order::CLASS_NAME_) {
+    return nullptr;
+  }
+
+  try {
+    auto myOrder = new Order((uint32_t)pdxReader.readLong(ORDER_ID_KEY_),
```
pivotal-jbarrett: Maybe a cleaner example doesn't use unsigned since the server has to store all integral types as signed.

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,93 @@
...
+                             pdxReader.readString(NAME_KEY_),
+                             (uint16_t)pdxReader.readInt(QUANTITY_KEY_));
+    return myOrder;
+  } catch (std::exception &e) {
```
pivotal-jbarrett: format

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+void OrderSerializer::deallocate(void *testObject,
+                                 const std::string &className) {
+  if (className == Order::CLASS_NAME_) {
+    Order *order = reinterpret_cast<Order *>(testObject);
+    delete order;
+  }
+}
+
+size_t OrderSerializer::objectSize(const void *testObject,
```
pivotal-jbarrett: Object size is really an advanced feature for local caches with LRU expiration. Let's leave it out of this example. Remove this and remove the getSize from the Order class. Default implementation of this method return 0.

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,93 @@
...
+  if (className == Order::CLASS_NAME_) {
+    return objectSize;
+  }
+  return nullptr;
+}
+
+void OrderSerializer::deallocate(void *testObject,
+                                 const std::string &className) {
+  if (className == Order::CLASS_NAME_) {
+    Order *order = reinterpret_cast<Order *>(testObject);
```
pivotal-jbarrett: Use auto.

examples/cpp/customserializer/Order.cpp
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Order.hpp"
+
+namespace customserializer {
+
+const std::string Order::CLASS_NAME_ = "com.example.Order";
```
pivotal-jbarrett: It is unlikely that your existing domain object, that you can't modify, would have this constant value. Should move this to your serializer where all code related to serializing this unmodified domain object should live.

examples/cpp/customserializer/Order.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+size_t Order::getSize() const {
+  size_t size = name_.length();
```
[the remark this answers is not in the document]
pivotal-jbarrett: All that said, see comment below about removing this method.

examples/cpp/customserializer/main.cpp
```diff
@@ -0,0 +1,64 @@
...
+  auto pool = poolFactory->create("pool");
+  auto regionFactory = cache.createRegionFactory(RegionShortcut::PROXY);
+  auto region = regionFactory.setPoolName("pool").create("custom_orders");
+
+  std::shared_ptr<PdxSerializer> orderSer =
+      std::shared_ptr<PdxSerializer>(new OrderSerializer());
+  cache.getTypeRegistry().registerPdxSerializer(orderSer);
+
+  auto order1 = new Order(1, "product x", 42);
+  std::shared_ptr<PdxWrapper> pdxobj1(
```
pivotal-jbarrett: Use `std::make_shared`:
```c++
auto pdxobj1 = std::make_shared<PdxWrapper>(order1, Order::CLASS_NAME_, orderSer);
```

examples/cpp/customserializer/main.cpp
```diff
@@ -0,0 +1,64 @@
...
+  cacheFactory.set("log-level", "none");
+  auto cache = cacheFactory.create();
+  auto poolFactory = cache.getPoolManager().createFactory();
+
+  poolFactory->addLocator("localhost", 10334);
+  auto pool = poolFactory->create("pool");
+  auto regionFactory = cache.createRegionFactory(RegionShortcut::PROXY);
+  auto region = regionFactory.setPoolName("pool").create("custom_orders");
+
+  std::shared_ptr<PdxSerializer> orderSer =
```
pivotal-jbarrett: Use `std::make_shared`.

examples/cpp/customserializer/main.cpp
```diff
@@ -0,0 +1,64 @@
...
+      std::shared_ptr<PdxSerializer>(new OrderSerializer());
+  cache.getTypeRegistry().registerPdxSerializer(orderSer);
+
+  auto order1 = new Order(1, "product x", 42);
+  std::shared_ptr<PdxWrapper> pdxobj1(
+      new PdxWrapper(order1, Order::CLASS_NAME_, orderSer));
+  auto order2 = new Order(2, "product y", 37);
+  std::shared_ptr<PdxWrapper> pdxobj2(
```
pivotal-jbarrett: Use `std::make_shared`.

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+std::shared_ptr<void> OrderSerializer::fromData(const std::string& className,
+                                                PdxReader& pdxReader) {
+  if (className != OrderSerializer::CLASS_NAME_) {
+    return nullptr;
+  }
+
+  try {
+    auto myOrder =
+        std::make_shared<Order>((uint32_t)pdxReader.readLong(ORDER_ID_KEY_),
```
pivotal-jbarrett: nit: While it is nice that you can do this, I wonder if having an explicit variable for each parameter would be more readable?

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,65 @@
...
+    pdxWriter.writeString(NAME_KEY_, myOrder->getName());
+    pdxWriter.writeInt(QUANTITY_KEY_, myOrder->getQuantity());
+  } catch (std::exception& e) {
+    std::cout << "Caught exception: " << e.what() << std::endl;
+    return false;
+  }
+  return true;
+}
+
+const std::string OrderSerializer::ORDER_ID_KEY_ = "orderid";
```
pivotal-jbarrett: I would group the collection of const initialization together. There is one at the top of this file too.

examples/cpp/customserializer/OrderSerializer.cpp
```diff
@@ -0,0 +1,65 @@
...
+const std::string OrderSerializer::NAME_KEY_ = "name";
+const std::string OrderSerializer::QUANTITY_KEY_ = "quantity";
+}  // namespace customserializer
```
pivotal-jbarrett: Some compilers warn about missing trailing line terminator.

examples/cpp/customserializer/main.cpp
```diff
@@ -0,0 +1,64 @@
...
+  auto cache = cacheFactory.create();
+  auto poolFactory = cache.getPoolManager().createFactory();
+
+  poolFactory->addLocator("localhost", 10334);
+  auto pool = poolFactory->create("pool");
+  auto regionFactory = cache.createRegionFactory(RegionShortcut::PROXY);
+  auto region = regionFactory.setPoolName("pool").create("custom_orders");
+
+  auto orderSer = std::make_shared<OrderSerializer>();
+  cache.getTypeRegistry().registerPdxSerializer(orderSer);
```
pivotal-jbarrett: I don't think we actually need to register a serializer when we use the pdx wrapper model. If you do then it seems rather redundant to have to specify it on the `PdxWrapper`. We should test this and write up a ticket to address this if you must register the serializer and pass it to the wrapper.

moleske: It is required, which is odd.  We should get a ticket for it.

PR #217, opened 2018-02-11
GEODE-4449: Replace Cacheable array macro-templates with plain template

dgkimura:
Testing: Unit/integration tests passed on Windows

### DIFF
--- a/cppcache/include/geode/CacheableBuiltins.hpp
+++ b/cppcache/include/geode/CacheableBuiltins.hpp
@@ -37,9 +37,6 @@ namespace apache {
 namespace geode {
 namespace client {
 
-template <typename Target, int8_t TYPEID>
-class CacheableArrayType;
-
 /** Template CacheableKey class for primitive types. */
 template <typename TObj, int8_t TYPEID, const char* TYPENAME,
           const char* SPRINTFSYM, int32_t STRSIZE>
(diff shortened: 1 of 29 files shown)


cppcache/include/geode/CacheableBuiltins.hpp
```diff
@@ -528,34 +346,63 @@ class _GEODE_EXPORT CacheableArray : public Cacheable {
...
+  }
 
-  inline const std::vector<T> value() const { return m_value; }
-  inline int32_t length() const { return m_value.size(); }
   static std::shared_ptr<Serializable> createDeserializable() {
     return std::make_shared<CacheableArray<T, GeodeTypeId>>();
   }
-  inline static std::shared_ptr<CacheableArray<T, GeodeTypeId>> create() {
+
+  static std::shared_ptr<CacheableArray<T, GeodeTypeId>> create() {
```
pivotal-jbarrett: Any reason for dropping the inline?

cppcache/integration-test/BuiltinCacheableWrappers.hpp
```diff
@@ -840,14 +840,14 @@ class CacheableBytesWrapper : public CacheableWrapper {
     auto randArr =
         CacheableHelper::randomArray<uint8_t>(maxSize, UCHAR_MAX);
     m_cacheableObject = CacheableBytes::create(
-        reinterpret_cast<const int8_t*>(randArr.data()), maxSize);
+        std::vector<int8_t>(randArr.data(), randArr.data() + maxSize));
```
pivotal-jbarrett: I wonder if since the intent is to have a random `CacheableBytes` (`int8_t`) then why not have the initial `randomArray` call return a `vector<int8_t>`?

cppcache/include/geode/CacheableBuiltins.hpp
```diff
@@ -490,32 +322,18 @@ _GEODE_CACHEABLE_KEY_TYPE_(char16_t, CacheableCharacter, 3);
 
 // Instantiations for array built-in Cacheables
 
-_GEODE_CACHEABLE_ARRAY_TYPE_DEF_(int8_t, CacheableBytes);
```
pivotal-jbarrett: Do we not need to add explicit instantiation statements for these types now? How does this effect documentation with Doxygen?

cppcache/include/geode/CacheableBuiltins.hpp
dgkimura: No reason really.  I just didn't think explicit `inline` added much value here.

cppcache/integration-test/BuiltinCacheableWrappers.hpp
dgkimura: Yes, that would be better.  I'll give it a quick stab.

cppcache/include/geode/CacheableBuiltins.hpp
dgkimura: Good question.  I'm not sure.

PR #224, opened 2018-02-16
GEODE-4697: Delete FunctionExecutionTimeOut.java and related tests

moleske:
- remove unused FunctionExcuteTimeOut

Signed-off-by: Ryan McMahon <rmcmahon@pivotal.io>


moleske:
updated PR name to be clearer now since we can't just do the rename

PR #225, opened 2018-02-21
GEODE-4341: .net example PutGetRemove added

austxcodemonkey:


examples/dotnet/Apache.Geode.Examples.PutGetRemove/Apache.Geode.Examples.PutGetRemove.csproj
```diff
@@ -23,8 +23,8 @@ limitations under the License.
     <ProjectGuid>{15ACFB5B-0D2C-4C38-9236-7AF01874A84F}</ProjectGuid>
```
echobravopapa: Is this project CMake generated?

echobravopapa: if so, the .csproj does not need to be checked in...

austxcodemonkey:
Requested changes made.  Since removing the AuthInitialize .csproj would leave us with no way to build it, we added it to the cmake setup as well, made it build and verified it runs correctly.

PR #226, opened 2018-02-23
Docs restructuring related to GEODE-4729

addisonhuddy:
Changes related to GEODE0-4729
- removal of quickstart documentation
- removal of example code
- addition of example page in "Getting Started" that points to /examples


echobravopapa:
Looks like a good start... @karensmolermiller @davebarnes97 @joeymcallister could one/many of you have a look from a bookbinder standpoint... Thx

docs/geode-native-book/Gemfile.lock
```diff
@@ -202,4 +202,4 @@ DEPENDENCIES
   libv8 (= 3.16.14.7)
 
 BUNDLED WITH
-   1.16.1
+   1.11.2
```
pivotal-jbarrett: Why is this being downgraded?

addisonhuddy: I had an older version of bundler installed, this will get updated in future docs stories.

davebarnes97:
Downloaded the changes, built the manual using bookbinder. Build succeeded.
VERY COOL to see someone else working in doc land! Two observations:
1. I found (and corrected) a badly-formatted link (which you inherited, not your fault) in the cpp-caching-api.html file.
2. Also caching-api.html: You deleted all the links to subsections from this top-level page, but left the subsections in place, along with their entries in the left-hand subnav. I notice this same change was not made in the parallel section for the .NET API. So, there's an inconsistency here. If your intention was to delete all this material, then you should also delete the files themselves and the left-hand nav entries that point to them. If the sections are to be included, then it's customary to provide links in the subnav and on the section's top-level page.

PR #228, opened 2018-02-28
GEODE-4411 Normalize AttributesFactory to match other factories

moleske:
- `createRegionAttributes` renamed to `create` and returns value
- `AttributesFactory` renamed to `RegionAttributesFactory`
- `AttributesFactory` constructor takes value of `RegionAttributes`


clicache/integration-test/AttributesMutatorTestsN.cs
```diff
@@ -142,10 +142,10 @@ public void StepOneCallbacks()
       //m_reg3Loader1 = new TallyLoader<string, string>();
       m_reg3Loader1 = new ThinClientTallyLoader();
       m_reg3Writer1 = new TallyWriter<string, string>();
-      AttributesFactory<string, string> af = new AttributesFactory<string, string>();
+      RegionAttributesFactory<string, string> regionAttributesFactory = new RegionAttributesFactory<string, string>();
```
pivotal-jbarrett: Please use (convert to) `var` in C# code where applicable.

```C#
var regionAttributesFactory = new RegionAttributesFactory<string, string>();
```

cppcache/integration-test/CacheHelper.cpp
```diff
@@ -883,14 +880,14 @@ std::shared_ptr<Region> CacheHelper::createPooledRegionStickySingleHop(
...
-  AttributesFactory af;
-  af.setCachingEnabled(caching);
+  RegionAttributesFactory regionAttributeFactory;
+  regionAttributeFactory.setCachingEnabled(caching);
   if (listener != nullptr) {
-    af.setCacheListener(listener);
+    regionAttributeFactory.setCacheListener(listener);
   }
-  std::shared_ptr<RegionAttributes> rattrsPtr = af.createRegionAttributes();
+  RegionAttributes regionAttributes = regionAttributeFactory.create();
```
pivotal-jbarrett: Please use `auto` where applicable in C++ code.
```C++
auto regionAttributes = regionAttributeFactory.create();
```

moleske:
Yeah there was a lot of search and replace.  Good points for `var` and `auto`, we'll take another pass and try to nab some of those since that wasn't at the top of mind with the other changes.

cppcache/include/geode/Region.hpp
```diff
@@ -114,7 +114,7 @@ class _GEODE_EXPORT Region : public std::enable_shared_from_this<Region> {
 
   /** Return the RegionAttributes for this region.
    */
-  virtual std::shared_ptr<RegionAttributes> getAttributes() const = 0;
+  virtual RegionAttributes getAttributes() const = 0;
```
pivotal-jbarrett: `RegionAttributes` are immutable so it might make a lot more sense to return a `const RegionAttributes&` here. It would fix the issue you addressed in a test change where you were taking the ref to the region name off an r-value RegionAttributes.

PR #229, opened 2018-03-01
Updates contributing instructions with refactoring guides.

pivotal-jbarrett:
- Fixes formatting and code block highlights.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,7 @@ This will take ~ 2 hours, YMMV... you can up the jobs to 4 and run in parallel, 
 $ cd build/cppcache/integration-test
 $ ctest -R <test_name> -C <Debug|Release>
 ```
+.NET integration tests can be executed similarly from `build/clicache/integration-test`.
 
 ## Style
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,10 +16,6 @@ Before submitting a pull request the unit and integration tests must all pass. W
 ```bash
 $ cd <clone>
 $ cd build
-```
-The following steps will be updated once the "run-unit-tests" target is fixed.
-
-```bash
 $ cd cppcache/test/<Debug|Release|if needed>
 $ ./gfcppcache_unittests
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,6 +55,7 @@ $ clang-format -i --style=file <PATH_TO_FILES>
 
 ### Code
 When writing new or refactoring old code please make the following changes.
+
  * Prefer the use of [`auto`](http://en.cppreference.com/w/cpp/language/auto) C++ or [`var`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/var) in C# where applicable.
    ```c++
    std::vector<Cars> cars = lot.getCars();
@@ -63,6 +64,7 @@ When writing new or refactoring old code please make the following changes.
    ```c++
    auto cars = lot.getCars();
    ```
+
  * Prefer [range for](http://en.cppreference.com/w/cpp/language/range-for) loops over traditional for loops where applicable.
    ```c++
    for (std::vector<Car>::iterator i = cars.begin(); i != cars.end(), ++i) {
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 This document assumes you have followed the [Apache Geode Code contribution instructions](https://cwiki.apache.org/confluence/display/GEODE/Code+contributions)
 
 ## Building the code
-    see BUILDING.md
+see BUILDING.md
```
moleske: if you use `[BUILDING.md](BUILDING.md)` then a nice link appears for clicking through

CONTRIBUTING.md
```diff
@@ -11,34 +11,112 @@ This document assumes you have followed the [Apache Geode Code contribution inst
...
+      Derived();
+      ~Derived() override;
+      virtualMethod() override;
+      pureVirtualMethod() override;
+    }
+    ```
 
-    $ clang-format -i --style=file <PATH_TO_FILES>
 
 # System Requirements
```
moleske: Is this section a duplicate of the **Building** section above?

CONTRIBUTING.md
```diff
@@ -11,34 +11,112 @@ This document assumes you have followed the [Apache Geode Code contribution inst
...
+```
 
-   Which is equivalent to running ctest directly:
+Which is equivalent to running ctest directly:
 
-    $ cd build/cppcache/integration-test
-    $ ctest --timeout 2000 -L STABLE -C <Debug|Release> -R . -j1
-   This will take ~ 2 hours, YMMV... you can up the jobs to 4 and run in parallel, but you may end up with test failures that will need to be re-run sequentially.  Like so:
+```bash
+$ cd build/cppcache/integration-test
```
moleske: Is it worth calling out the `clicache/integration-test` explicitly since we should also expect those tests to pass?

moleske:
Some more things to consider adding
1. use `std::dynamic_pointer_cast` over `dynamic_cast`
  - `dynamic_cast<MyCar*>(car.get());` to `std::dynamic_pointer_cast<MyCar>(car);`
2. don't use `c_str()`
3. `LOG` can just use strings (no buffer)

CONTRIBUTING.md
echobravopapa: Should work if running on Windows, otherwise is a no-op

CONTRIBUTING.md
```diff
@@ -11,34 +11,112 @@ This document assumes you have followed the [Apache Geode Code contribution inst
...
+   ```
+  * Fix bad variable names. Variable names should be expressive.
+    ```c++
+    double i = car.getFuelLevel();
+    ```
+    should be changed to
+    ```c++
+    auto fuelLevel = car.getFuelLevel();
+    ```
+  * Use [`override`]() on all method overrides.
```
echobravopapa: maybe add "see [override](http://en.cppreference.com/w/cpp/language/override) at end of line

CONTRIBUTING.md
```diff
@@ -11,34 +11,112 @@ This document assumes you have followed the [Apache Geode Code contribution inst
...
-   Before submitting a pull request the unit and integration tests must all pass. We are using CTest, (Please see [the CTest documentation](https://cmake.org/Wiki/CMake/Testing_With_CTest) for further information.)
+Before submitting a pull request the unit and integration tests must all pass. We are using CTest, (Please see [the CTest documentation](https://cmake.org/Wiki/CMake/Testing_With_CTest) for further information.)
 ### Running unit tests
-    $ cd <clone>
-    $ cd build
+```bash
+$ cd <clone>
+$ cd build
+```
+The following steps will be updated once the "run-unit-tests" target is fixed.
```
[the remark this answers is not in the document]
pivotal-jbarrett: What is wrong with it?

pivotal-jbarrett:
@moleske @echobravopapa you both should have access to edit this file directly from this pull request if you want.

CONTRIBUTING.md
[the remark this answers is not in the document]
echobravopapa: don't know.... the document says the steps will be updated when "run-unit-tests" is fixed...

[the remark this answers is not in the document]
pivotal-jbarrett: Yeah beats me!

PR #230, opened 2018-03-05
GEODE-4408: Move CacheableKeys.hpp to internal include dir

austxcodemonkey:
Signed-off-by: Echo Bravo <eburghardt@pivotal.io>

### DIFF
--- a/cppcache/include/geode/CacheableBuiltins.hpp
+++ b/cppcache/include/geode/CacheableBuiltins.hpp
@@ -92,7 +92,7 @@ class CacheableKeyType : public CacheableKey {
 
   /** Return the hashcode for this key. */
   virtual int32_t hashcode() const override {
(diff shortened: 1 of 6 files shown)


cppcache/include/geode/CacheableBuiltins.hpp
```diff
@@ -30,7 +30,7 @@
 #include "Serializable.hpp"
 #include "CacheableKey.hpp"
 #include "Serializer.hpp"
-#include "CacheableKeys.hpp"
+#include "internal/CacheableKeys.hpp"
 #include "CacheableString.hpp"
 
 namespace apache {
```
pivotal-jbarrett: Needs to be moved into the internal namespace too.

echobravopapa:
@pivotal-jbarrett how about now? ;)

PR #233, opened 2018-03-07
GEODE-4427: Moving CliCallbackMethod to DistributedSystemImpl

austxcodemonkey:


mcmellawatt:
Fixed the build issue, needed some changes in the c++/cli.

clicache/src/DistributedSystem.cpp
```diff
@@ -493,7 +493,7 @@ namespace Apache
           &CliCallbackDelegate::Callback);
 
         native::DistributedSystemImpl::registerCliCallback(System::Threading::Thread::GetDomainID(),
-                                                                          (native::CliCallbackMethod)System::Runtime::InteropServices::
+                                                                          (void (*)(apache::geode::client::Cache &cache))System::Runtime::InteropServices::
```
dgkimura: Given that you remove a C-style function pointer on C++ side (which I think is appropriate), is there any reason to perpetuate the C-style function pointer here? If you must, then why not use a C++ style cast? Furthermore, is it possible to replace CliCallbackDelegate and cliCallback with some std::functional magic now that that's what will be used in the C++ side?

pivotal-jbarrett: The cast is required because the delegate function pointer given by this call is a `void*`. It won’t convert to anything that matches `std::function` will accept.

dgkimura: Do we need to use delegates?  Was that required because of appdomain fun?  If so, is it still relevant in a world without globals?

pivotal-jbarrett: Sadly we still need this delegate to call back from the native side into the .NET side for cleanup when the cachenis closed. It could probably be refactored but its outside this scope.

PR #235, opened 2018-03-09
Geode 4421: Convert Execution factory to value type

igodwin:


cppcache/include/geode/Execution.hpp
```diff
@@ -17,100 +17,122 @@
 
 #pragma once
 
-#ifndef GEODE_EXECUTION_H_
-#define GEODE_EXECUTION_H_
+#ifndef GEODE_EXECUTIONIMPL_H_
```
igodwin: Should drop the "IMPL" on this. There are a couple more occurrences, too.

cppcache/include/geode/Execution.hpp
```diff
@@ -17,100 +17,122 @@
...
+#include <ace/Guard_T.h>
 
-/**
- * @file
- */
+#include <geode/CacheableBuiltins.hpp>
+#include <geode/ResultCollector.hpp>
+#include <geode/Region.hpp>
+
+#include "../../src/ProxyCache.hpp"
```
dgkimura: Include path looks suspicious to me...

cppcache/include/geode/Execution.hpp
```diff
@@ -17,100 +17,122 @@
...
+        m_region(nullptr),
+        m_allServer(allServer),
+        m_pool(pool),
+        m_proxyCache(proxyCache) {}
+
+  Execution withFilter(std::shared_ptr<CacheableVector> routingObj);
+
+  Execution withArgs(std::shared_ptr<Cacheable> args);
+
+  Execution withCollector(std::shared_ptr<ResultCollector> rs);
```
dgkimura: I'm glad to see more of these changes happening

cppcache/include/geode/Execution.hpp
```diff
@@ -17,100 +17,122 @@
...
-#include <memory>
-#include <chrono>
-#include <string>
+#include <map>
 
-#include "internal/geode_globals.hpp"
-#include "CacheableBuiltins.hpp"
-#include "ResultCollector.hpp"
+#include <ace/Condition_Recursive_Thread_Mutex.h>
+#include <ace/Guard_T.h>
```
dgkimura: I don't think we want to leak out this ACE dependency in a public header.  However, if I recall correctly you may be able to replace these usages with `std::lock_guard` and `std::mutex` or `std::recursive_mutex`.

pivotal-jbarrett: Yeah and then the C++/CLI will fail to compile because `std::mutex` is forbidden. 

This attemp will need to be refactored into a modern pimpl pattern.

PR #240, opened 2018-03-16
 GEODE-4850: Update minimum CMake version to 3.10

pivotal-jbarrett:


docker/Dockerfile
```diff
@@ -17,18 +17,22 @@
 FROM ubuntu
 LABEL maintainer Apache Geode <dev@geode.apache.org>
 
-RUN \
-    apt-get update \
-    && apt-get install -y build-essential cmake doxygen git graphviz openjdk-8-jdk wget zlib1g-dev \
-    && rm -rf /var/lib/apt/lists/*
+RUN apt-get update
```
metatype: You want these all on one line to reduce the number of layers in your image.

PR #246, opened 2018-03-21
GEODE-4912: Return value for DistributedSystem::create

moleske:


clicache/src/DistributedSystem.hpp
```diff
@@ -122,7 +122,7 @@ namespace Apache
         /// </returns>
         static DistributedSystem^ Create(native::DistributedSystem* nativeptr);
 
-        DistributedSystem(std::unique_ptr<native::DistributedSystem> nativeptr);
+        DistributedSystem(native::DistributedSystem nativeDistributedSystem);
```
pivotal-jbarrett: This results in a copy of the DistributedSystem. We don't want to make a copy, in this case we would move into here and then into a `std::unique_ptr` allocated heap instance.

cppcache/include/geode/DistributedSystem.hpp
```diff
@@ -108,16 +114,13 @@ class _GEODE_EXPORT DistributedSystem {
  private:
   std::string m_name;
   bool m_connected;
-  std::unique_ptr<statistics::StatisticsManager> m_statisticsManager;
-  std::unique_ptr<SystemProperties> m_sysProps;
+  std::shared_ptr<statistics::StatisticsManager> m_statisticsManager;
```
pivotal-jbarrett: Why are these suddenly `std::shared_ptr`?

cppcache/include/geode/DistributedSystem.hpp
```diff
@@ -56,15 +56,21 @@ class _GEODE_EXPORT DistributedSystem {
    * @brief public methods
    */
  public:
+  DistributedSystem(const DistributedSystem&) = default;
```
pivotal-jbarrett: Copy construct and assignment should be deleted.

cppcache/src/CacheImpl.cpp
```diff
@@ -46,15 +45,15 @@
 
 using namespace apache::geode::client;
 
-CacheImpl::CacheImpl(Cache* c, std::unique_ptr<DistributedSystem> sys,
-                     bool iPUF, bool readPdxSerialized,
+CacheImpl::CacheImpl(Cache* c, DistributedSystem distributedSystem, bool iPUF,
```
pivotal-jbarrett: You have changed this method from taking ownership of `DistributedSystem` to making a copy. This method should be taking an r-value and moving ownership to the `CacheImpl`.

cppcache/src/CacheImpl.cpp
```diff
@@ -46,15 +45,15 @@
...
+                     bool readPdxSerialized,
                      const std::shared_ptr<AuthInitialize>& authInitialize)
     : m_defaultPool(nullptr),
       m_ignorePdxUnreadFields(iPUF),
       m_readPdxSerialized(readPdxSerialized),
       m_closed(false),
       m_initialized(false),
-      m_distributedSystem(std::move(sys)),
+      m_distributedSystem(distributedSystem),
```
pivotal-jbarrett: `std::move`

cppcache/src/CacheImpl.hpp
```diff
@@ -140,7 +142,7 @@ class _GEODE_EXPORT CacheImpl : private NonCopyable, private NonAssignable {
    * Returns the distributed system that this cache was
    * {@link CacheFactory::create created} with.
    */
-  DistributedSystem& getDistributedSystem() const;
+  DistributedSystem& getDistributedSystem();
```
pivotal-jbarrett: Why drop the `const` here?

cppcache/src/CacheImpl.hpp
```diff
@@ -213,7 +214,7 @@ class _GEODE_EXPORT CacheImpl : private NonCopyable, private NonAssignable {
   /**
    * @brief constructors
    */
-  CacheImpl(Cache* c, std::unique_ptr<DistributedSystem> sys,
+  CacheImpl(Cache* c, DistributedSystem distributedSystem,
```
pivotal-jbarrett: See comment about impl needing to move ownership.

cppcache/src/DistributedSystem.cpp
```diff
@@ -59,11 +58,10 @@ DistributedSystem::DistributedSystem(
     DiffieHellman::initOpenSSLFuncPtrs();
   }
 }
-DistributedSystem::~DistributedSystem() {}
 
 void DistributedSystem::logSystemInformation() const {
-  std::string gfcpp = CppCacheLibrary::getProductDir();
-  LOGCONFIG("Using Geode Native Client Product Directory: %s", gfcpp.c_str());
+  std::string productDir = CppCacheLibrary::getProductDir();
```
pivotal-jbarrett: `auto`

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -46,10 +45,8 @@ void DistributedSystemImpl::disconnect() {
 void DistributedSystemImpl::CallCliCallBack(Cache& cache) {
   ACE_Guard<ACE_Recursive_Thread_Mutex> disconnectGuard(m_cliCallbackLock);
   if (m_isCliCallbackSet == true) {
-    for (std::map<int, CliCallbackMethod>::iterator iter =
-             m_cliCallbackMap.begin();
-         iter != m_cliCallbackMap.end(); ++iter) {
-      (*iter).second(cache);
+    for (auto& iter : m_cliCallbackMap) {
```
pivotal-jbarrett: `const auto &`?

cppcache/src/CacheImpl.hpp
austxcodemonkey: The Cache class calls connect() and disconnect() on the returned value from getDistributedSystem(), which are non-const methods.  We put the const back in, and referred to the cacheImpl distributedSystem directly - let us know which way you prefer, or what option 3 might be.

cppcache/include/geode/DistributedSystem.hpp
moleske: Short answer: `shared_ptr` is what would get it to build with our other changes.  Longer answer: when m_statisticsManager is a `unique_ptr` now, it breaks down somewhere in <memory> because some code path in std calls sizeof(StatisticsManager), which is unknown at compile time because the class is a forward reference.  Suppose this _could be_ solved with a custom move ctor, but is that the right thing to do?

The other looks like it was oversight, changed it back.

cppcache/src/CacheImpl.hpp
pivotal-jbarrett: Nothing should be calling DS connection/disconnect directly anymore.

clicache/src/DistributedSystem.cpp
```diff
@@ -448,11 +448,11 @@ namespace Apache
       {
         try
         {
-          return marshal_as<String^>(m_nativeptr->get()->getName().c_str());
+          return marshal_as<String^>(m_nativeDistributedSystem->get()->getName().c_str());
```
pivotal-jbarrett: Remove the `c_str()` call.

clicache/src/DistributedSystem.cpp
```diff
@@ -470,14 +470,15 @@ namespace Apache
         return instance;
       }
 
-      DistributedSystem::DistributedSystem(std::unique_ptr<native::DistributedSystem> nativeptr)
+      DistributedSystem::DistributedSystem(native::DistributedSystem nativeDistributedSystem)
```
pivotal-jbarrett: We don't want copies. Copy constructor should be deleted. This should be r-value reference.

cppcache/src/statistics/StatisticsManager.hpp
```diff
@@ -71,6 +71,8 @@ class StatisticsManager {
                     CacheImpl* cache, const char* durableClientId,
                     const std::chrono::seconds durableTimeout,
                     int64_t statFileLimit = 0, int64_t statDiskSpaceLimit = 0);
+  StatisticsManager(const StatisticsManager& statisticsManager) = default;
```
pivotal-jbarrett: Why are we adding copy constructors? There should only be one `StatisticsManager` per `Cache`. We don't ever want it to copy.

cppcache/src/DistributedSystemImpl.hpp
```diff
@@ -84,16 +84,17 @@ class _GEODE_EXPORT DistributedSystemImpl {
...
 
+  std::string m_name;
+  DistributedSystem m_implementee;
   DiffieHellman m_dh;
 
   /**
    * @brief constructors
    */
-  DistributedSystemImpl(const char* name, DistributedSystem* implementee);
+  DistributedSystemImpl(std::string name, DistributedSystem implementee);
```
pivotal-jbarrett: This is going to copy the `DistributedSystem`.  You need to keep a pointer here. `DistributedSystemImpl` is owned by `DistributedSystem`, the pointer back to `DistributedSystem` is non-owning reference. See `Cache` and `CacheImpl` pimpl model.

pivotal-jbarrett:
@moleske @austxcodemonkey please rebase.

PR #251, opened 2018-03-27
Remove back references to cache

moleske:


pivotal-jbarrett:
Looks like travis was just not happy with how long the compile was taking.

PR #252, opened 2018-03-29
GEODE-4946: Treat warnings as error on MSVC

pivotal-jbarrett:
- Removes all warnings.
- Treat all new warnings as errors.


mmartell:
This is huge! The plethora of warnings in this code has always made it difficult to sort out the serious issues. This is of particular importance to a big project like this as it's being migrated to C++11.

PR #253, opened 2018-03-29
GEODE-4968: Parameterized product names in Find module

mcmellawatt:


CMakeLists.txt
```diff
@@ -248,12 +248,14 @@ add_subdirectory(sqliteimpl)
...
 add_subdirectory(templates/security)
 add_subdirectory(docs/api)
+add_subdirectory(examples/cmake)
 if (${BUILD_CLI})
   add_subdirectory(clicache)
   add_subdirectory(plugins/SQLiteCLI)
 endif()
 
-install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/examples/ DESTINATION examples)
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/examples/ DESTINATION examples
```
pivotal-jbarrett: You don’t want to install then .in files. This should be the configured files in the binary dir now.

examples/.gitignore
```diff
@@ -0,0 +1,3 @@
+cpp/**/CMakeLists.txt
```
pivotal-jbarrett: Should not need these ignored. CMake files should be generated into the binary for.

examples/cmake/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+string(REPLACE " " "" PRODUCT_NAME_NOSPACE ${PRODUCT_NAME})
+
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/FindGeodeNative.cmake.in ${CMAKE_CURRENT_SOURCE_DIR}/Find${PRODUCT_NAME_NOSPACE}.cmake @ONLY)
+
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/../cpp/customserializable/CMakeLists.txt.in ${CMAKE_CURRENT_SOURCE_DIR}/../cpp/customserializable/CMakeLists.txt @ONLY)
```
pivotal-jbarrett: These should not be in this CMake file. They should also be generating into the binary directory. You NEVER modify files in the source directory.

PR #258, opened 2018-04-03
GEODE-4916: Fixes compilation errors when including in empty VC++ project.

pivotal-jbarrett:
- Convert to CMake GenerateExportHeader.
- Remove use of sprintf which caused error on Windows.


pivotal-jbarrett:
Found some issues.

tests/cli/PkcsWrapper/CMakeLists.txt
```diff
@@ -36,20 +36,23 @@ set_target_properties(${PROJECT_NAME} PROPERTIES
   VS_DOTNET_REFERENCES "System;System.Xml"
 )
 
-target_link_libraries(${PROJECT_NAME}
+target_link_libraries(PkcsWrapper
   PUBLIC 
-    Apache.Geode
+    # Apache.Geode #- Causes include of .lib
```
mcmellawatt: Same comment.  We need this despite the whole line being commented out?

tests/cli/QueryHelper/CMakeLists.txt
```diff
@@ -31,28 +31,31 @@ set_target_properties(${PROJECT_NAME} PROPERTIES
...
   PRIVATE
     $<BUILD_INTERFACE:$<TARGET_PROPERTY:Apache.Geode,SOURCE_DIR>>
     $<BUILD_INTERFACE:$<TARGET_PROPERTY:testThinClientRemoteQueryRS,SOURCE_DIR>>
 )
 
-target_link_libraries(${PROJECT_NAME}
+target_link_libraries(QueryWrapper
   PUBLIC 
-    Apache.Geode
+    # Apache.Geode #- Causes include of .lib
```
mcmellawatt: Same comment as above

clicache/test/CMakeLists.txt
```diff
@@ -32,32 +32,35 @@ add_library(${PROJECT_NAME} MODULE
...
   	$<TARGET_PROPERTY:apache-geode,SOURCE_DIR>
   	$<TARGET_PROPERTY:apache-geode,INCLUDE_DIRECTORIES>
   	$<TARGET_PROPERTY:Apache.Geode,SOURCE_DIR>
 )
 
-target_link_libraries(${PROJECT_NAME}
+target_link_libraries(Apache.Geode.Test
   PUBLIC
-    Apache.Geode
+    # Apache.Geode #- Causes include of .lib
```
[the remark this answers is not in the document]
pivotal-jbarrett: It's a good question. I left the comment there because the "normal" use case is that this would be here but because CMake doesn't really understand the a C++/CLI can create a .DLL without a .LIB file it creates a dependency that fails. It is not the normal way to specify library dependencies by using the target dependency function used later in the file. Someone that later came back to this file might rightly believe the file was in error and "correct" it only to wonder why it doesn't work.

cppcache/shared/CMakeLists.txt
```diff
@@ -59,13 +70,13 @@ if (USE_PCH)
...
+  set_target_properties(apache-geode PROPERTIES
     COTIRE_ENABLE_PRECOMPILED_HEADER TRUE
     COTIRE_ADD_UNITY_BUILD TRUE
     COTIRE_PREFIX_HEADER_IGNORE_PATH "${_COTIRE_PREFIX_HEADER_IGNORE_PATH}"
     COTIRE_PREFIX_HEADER_INCLUDE_PATH "${CMAKE_BINARY_DIR}/dependencies;${CMAKE_CURRENT_SOURCE_DIR}/../include")
 
   set_source_files_properties(cppcache/version.h PROPERTIES GENERATED TRUE)
 
-  cotire(${PROJECT_NAME})
+  cotire(apache-geode)
```
echobravopapa: just saw this cotire speed up

pivotal-jbarrett: Dude that was so 4 weeks ago!

PR #260, opened 2018-04-04
 GEODE-4967: Treat warnings as error on Clang

pivotal-jbarrett:
-Wall
-Werror
-Wextra w/ exclusions
-Wpedantic
-Xeverything w/ exclusions


mmartell:
Whew! Lots of work here. Indeed, does build the apache-geode project and associated test projects **warning free!**

One question: Looks like we're using the default warning level on Windows (/W3). Are there plans to fix the errors reported by W4 (or perhaps even /Wall). This is the recommended level for finding bugs. I did a build with /W4 and there's a lot of warnings, but not that many warning types.

pivotal-jbarrett:
Many of the MSVC++ warnings at level 4 were in “system headers” but the compiler lacks the ability to define system headers. There is an experimental feature in the 2017 version of the compiler. We would have to add lots of pragma warning push/pops to our includes to work around it for the 2015 compiler.

mmartell:
I just looked at the C4244 errors (truncation). Although there's a lot of
them (307), easy to fix using a handful of casts.

PR #266, opened 2018-04-06
GEODE-4994: Fix windows compilation

mcmellawatt:
- Missing inttypes.h header added
- Remaining references to 'ret' in log messages removed

### DIFF
--- a/cppcache/integration-test/testXmlCacheCreationWithPools.cpp
+++ b/cppcache/integration-test/testXmlCacheCreationWithPools.cpp
@@ -16,11 +16,12 @@
  */
 
 #include <inttypes.h>
-#include <geode/internal/chrono/duration.hpp>
 #include <cstdint>
 #include <iostream>
 #include <string>
 #include <vector>
+
+#include <geode/internal/chrono/duration.hpp>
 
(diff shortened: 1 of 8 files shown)


cppcache/integration-test/testXmlCacheCreationWithPools.cpp
```diff
@@ -15,14 +15,14 @@
...
 
-#include <string>
-#include <iostream>
-#include <vector>
-#include <cstdint>
-
 #include "fw_dunit.hpp"
 
 #include <geode/internal/chrono/duration.hpp>
+#include <cstdint>
```
pivotal-jbarrett: Move system headers to the top. Inttypes.h is a C header not C++.

mcmellawatt: Just to clarify, using Inttypes.h is fine its just the ordering which isn't right?

mcmellawatt: Pushed a fix for ordering - if you had other concerns we can fix it in a subsequent commit.

PR #268, opened 2018-04-10
GEODE-4637: Allow only GEODE_HOME, remove GEODE

mmartell:


BUILDING.md
```diff
@@ -6,7 +6,7 @@
 * [Doxygen 8.11](http://www.stack.nl/~dimitri/doxygen/download.html) *(for building source documentation)*
 * [Apache Geode](http://geode.apache.org/releases/) binaries installed or available to link against
 
-   Building requires access to an installation of Geode. By default the value of `GEODE_ROOT` or `GEODE` is used during CMake configuration if either of those shell variables is exported. To explicitly specify the location in which Geode is installed, add `-DGEODE_ROOT=/path/to/geode` to the _initial_ `cmake` execution command.
+   Building requires access to an installation of Geode. By default the value of `GEODE_ROOT` or `GEODE_HOME` is used during CMake configuration if either of those shell variables is exported. To explicitly specify the location in which Geode is installed, add `-DGEODE_ROOT=/path/to/geode` to the _initial_ `cmake` execution command.
```
pivotal-jbarrett: This shouldn't be correct. See note blow. 

`GEODE_ROOT` should only be a CMake variable defined at configure time via `-DGEODE_ROOT` and only needs to be defined if `GEODE_HOME` environment is not set or Geode is not installed in a well known location.
`GEODE_HOME` is an environment variable that may be set if Geode is not installed in a well known location.

cmake/FindGeode.cmake
```diff
@@ -26,8 +26,6 @@ else()
     file(TO_CMAKE_PATH "$ENV{GEODE_ROOT}" _ENV_GEODE_ROOT)
```
pivotal-jbarrett: Should not be looking for an environment variable `GEODE_ROOT`. This is a CMake variable the should override `GEODE_HOME` environment variable or any well known location of Geode.

echobravopapa:
@pivotal-jbarrett have a look, requested changes made, will squash/merge if approved

PR #274, opened 2018-04-13
GEODE-5066: Accessing cacheImpl directly to avoid access exception

mcmellawatt:


cppcache/src/statistics/StatArchiveWriter.cpp
```diff
@@ -40,19 +40,19 @@ using std::chrono::system_clock;
...
+StatDataOutput::StatDataOutput(CacheImpl *cacheImpl)
     : bytesWritten(0), m_fp(nullptr), closed(false) {
-  dataBuffer = cache->getCache()->createDataOutput();
+  dataBuffer = cacheImpl->createDataOutput();
 }
 
-StatDataOutput::StatDataOutput(std::string filename, CacheImpl *cache) {
+StatDataOutput::StatDataOutput(std::string filename, CacheImpl *cacheImpl) {
   if (filename.length() == 0) {
     std::string s("undefined archive file name");
```
moleske: Would be nice to not use `s` as the variable name, maybe `exceptionMessage`?

cppcache/src/statistics/StatArchiveWriter.cpp
```diff
@@ -40,19 +40,19 @@ using std::chrono::system_clock;
...
     throw IllegalArgumentException(s.c_str());
```
moleske: Don't need `c_str()`.

pivotal-jbarrett: Actually we don't need that variable at all just.
```
throw IllegalArgumentException("undefined archive file name");
```

PR #275, opened 2018-04-13
 GEODE-4728 - Update defaultSystem/geode.properties

moleske:
@davebarnes97 and I stumbled upon these when looking at some documentation

### DIFF
--- a/defaultSystem/geode.properties
+++ b/defaultSystem/geode.properties
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# Default Gemfire C++ distributed system properties
+# Default Geode Native distributed system properties
 # Copy to current directory and uncomment to override defaults.
 #
 ## Debugging support, enables stacktraces in apache::geode::client::Exception.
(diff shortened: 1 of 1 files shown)


defaultSystem/geode.properties
```diff
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# Default Gemfire C++ distributed system properties
+# Default C++ distributed system properties
```
pivotal-jbarrett: These properties are also loaded by the .NET client. I wonder if we can give them a better name.

defaultSystem/geode.properties
```diff
@@ -22,13 +22,13 @@
...
 #
-## Cache region configurtion
+## Cache region configuration
 #
 #cache-xml-file=cache.xml
 #
 ## Log file config
 #
-#log-file=gemfire_cpp.log
+#log-file=geode_cpp.log
```
pivotal-jbarrett: Same here.

defaultSystem/geode.properties
pivotal-jbarrett: Geode Native probably. This file shouldn't get confused with the one for the Java client/server.

pivotal-jbarrett:
@moleske please use the GEODE number in the PR title and at least the first commit message.

moleske:
My bad, was on GitHub when we saw it and forgot to check Jira before blindly updating comments

PR #278, opened 2018-04-17
GEODE-5018: Fixing PIMPL pattern for Cache

mcmellawatt:


cppcache/include/geode/Cache.hpp
```diff
@@ -76,7 +78,8 @@ class APACHE_GEODE_EXPORT Cache : public GeodeCache {
    * @param regionShortcut
    *        To create the region specific type, @see RegionShortcut
    */
-  virtual RegionFactory createRegionFactory(RegionShortcut regionShortcut);
```
pivotal-jbarrett: Region shortcut is an enum type. The size of the ref is actually more expensive than the enum value.

cppcache/integration-test/testExpiration.cpp
```diff
@@ -378,8 +395,9 @@ BEGIN_TEST(TEST_EXPIRATION)
     //////////////
     RegionAttributes attrs_18;
     // ettl = 0, eit = 0, rttl = 0, reit = 0
-    attrs_18 = setRegionAttributesTimeouts(std::chrono::seconds(6), std::chrono::seconds(3),
```
pivotal-jbarrett: Seems like a lot of formatting changes. Are you using clang-format?

mcmellawatt: Yes I ran all the source I touched through clang-format.

cppcache/include/geode/Cache.hpp
pivotal-jbarrett: That’s fine, it’s a small enum. Though I would expect a forward to work. See docs for `enum class`.

cppcache/integration-test/testExpiration.cpp
pivotal-jbarrett: What version of clang-format?

cppcache/include/geode/Cache.hpp
mcmellawatt: You are right - forward declaration does work, CLion was giving me a false error, but it compiles fine.  I'll watch out for that sort of thing in the future.

PR #282, opened 2018-04-20
GEODE-4854: CqQuery::getCqAttributesMutator returns a value type

igodwin:


clicache/src/CqAttributesMutator.hpp
```diff
@@ -77,40 +75,40 @@ namespace Apache
 
 
         /// <summary>
-	/// Initialize with an array of listeners
+	      /// Initialize with an array of listeners
```
echobravopapa: formatting...

PR #285, opened 2018-04-23
GEODE-4839: Use std::shared_ptr to hold persistent info object.

pivotal-jbarrett:


cppcache/src/LRUAction.cpp
```diff
@@ -67,8 +67,8 @@ bool LRUOverFlowToDiskAction::evict(
         "OverflowAction: destroyed entry added to "
         "LRU list");
   }
-  LRUEntryProperties& lruProps = mePtr->getLRUProperties();
-  void* persistenceInfo = lruProps.getPersistenceInfo();
+  auto&& lruProps = mePtr->getLRUProperties();
```
echobravopapa: How did you arrive at using `universal reference` here and below?  Don't see any `std::forward` in play...

pivotal-jbarrett: Universal references don't require `std::forward`. By using `auto&&` I don't need to concern myself with the return type or it's constness. In this particular line its `const auto&`.

PR #288, opened 2018-04-26
 GEODE-4977: Make query results behave like standard containers

pivotal-jbarrett:


cppcache/integration-test-2/StructTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+                    .create("region");
+
+  return region;
+}
+
+/**
+ * Example test using 2 servers and waiting for async tasks to synchronize using
+ * furtures.
+ */
+TEST(StructTest, queryResultForRange) {
```
mcmellawatt: Are you just trying to test that the queryresults iterator and new struct iterator work?  If so, it seems like a lot of this test could be reworked/deleted.  See my additional comments below.

cppcache/integration-test-2/StructTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    auto key = -1;
+    for (auto&& column : *rowStruct) {
+      // Expect to read: key:int, value:string
+      if (auto columnValue = std::dynamic_pointer_cast<CacheableInt32>(column)) {
+        key = columnValue->value();
+        EXPECT_NE(values.end(), values.find(key));
+        break;
+      }
+      auto columnValue = std::dynamic_pointer_cast<CacheableString>(column);
```
mcmellawatt: Code below this line is not executed because all `column` types in this test are CacheableInt32.

cppcache/integration-test-2/StructTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+    auto rowStruct = std::dynamic_pointer_cast<Struct>(row);
+    ASSERT_NE(nullptr, rowStruct);
+
+    auto key = -1;
+    for (auto&& column : *rowStruct) {
+      // Expect to read: key:int, value:string
+      if (auto columnValue = std::dynamic_pointer_cast<CacheableInt32>(column)) {
+        key = columnValue->value();
+        EXPECT_NE(values.end(), values.find(key));
+        break;
```
mcmellawatt: We always break because the type is always CacheableInt32.  Is this desirable?

cppcache/integration-test-2/StructTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+  for (auto&& row : *queryResult) {
+    auto rowStruct = std::dynamic_pointer_cast<Struct>(row);
+    ASSERT_NE(nullptr, rowStruct);
+
+    auto key = -1;
+    for (auto&& column : *rowStruct) {
+      // Expect to read: key:int, value:string
+      if (auto columnValue = std::dynamic_pointer_cast<CacheableInt32>(column)) {
+        key = columnValue->value();
+        EXPECT_NE(values.end(), values.find(key));
```
mcmellawatt: What is this asserting?  end() and find() return hash_iterator objects - may just be my misunderstanding but the value of this assertion is not clear.

cppcache/include/geode/Struct.hpp
```diff
@@ -154,17 +140,15 @@ class APACHE_GEODE_EXPORT Struct : public Serializable {
   }
 
  private:
+  Struct() = default;
```
mcmellawatt: Is this used by friend classes?  If not we could make it public and delete...

cppcache/integration-test/QueryHelper.hpp
```diff
@@ -315,76 +315,58 @@ void QueryHelper::getPDXObject(std::shared_ptr<Region>& rptr) {
...
-  }
-
-  auto ssptr = std::static_pointer_cast<StructSet>(structSet);
+  if (auto ssptr = std::static_pointer_cast<StructSet>(structSet)) {
+    size_t foundRows = 0;
+    for (auto&& ser : *ssptr) {
+      foundRows++;
 
-  size_t foundRows = 0;
+      auto siptr = std::dynamic_pointer_cast<Struct>(ser);
```
mcmellawatt: variable naming

cppcache/integration-test/QueryHelper.hpp
```diff
@@ -315,76 +315,58 @@ void QueryHelper::getPDXObject(std::shared_ptr<Region>& rptr) {
...
-  if (!std::dynamic_pointer_cast<StructSet>(structSet)) {
-    if (expectedRows == 0 && expectedFields == 0) {
-      return true;  // quite possible we got a null set back.
-    }
-    printf("we have structSet itself nullptr \n");
-    return false;
-  }
-
-  auto ssptr = std::static_pointer_cast<StructSet>(structSet);
+  if (auto ssptr = std::static_pointer_cast<StructSet>(structSet)) {
```
mcmellawatt: variable name could be better perhaps

cppcache/integration-test/QueryHelper.hpp
```diff
@@ -315,76 +315,58 @@ void QueryHelper::getPDXObject(std::shared_ptr<Region>& rptr) {
...
 
-  size_t foundRows = 0;
+      auto siptr = std::dynamic_pointer_cast<Struct>(ser);
 
-  for (SelectResults::Iterator iter = ssptr->begin(); iter != ssptr->end();
-       iter++) {
-    auto ser = (*iter);
-    foundRows++;
+      if (siptr == nullptr) {
+        printf("siptr is nullptr \n\n");
```
mcmellawatt: cout?

cppcache/integration-test/QueryHelper.hpp
```diff
@@ -315,76 +315,58 @@ void QueryHelper::getPDXObject(std::shared_ptr<Region>& rptr) {
...
-    }
-
-    if (foundFields != expectedFields) {
-      char buffer[1024] = {'\0'};
-      sprintf(buffer, "found fields %d, expected fields %d \n", foundFields,
-              expectedFields);
-      LOG(buffer);
-      return false;
+    // lets log and return in case of error only situation
+    char buffer[1024] = {'\0'};
```
mcmellawatt: std::string and ostringstream?

cppcache/integration-test/QueryHelper.hpp
```diff
@@ -315,76 +315,58 @@ void QueryHelper::getPDXObject(std::shared_ptr<Region>& rptr) {
...
+    sprintf(buffer, "found rows %zd, expected rows %zd\n", foundRows,
+            expectedRows);
+    LOG(buffer);
+  } else {
+    if (expectedRows == 0 && expectedFields == 0) {
```
mcmellawatt: Won't we have already returned from the `if (foundRows == expectedRows)` check above?

clicache/integration-test/ThinClientRegionQueryTests.cs
```diff
@@ -182,7 +182,7 @@ public void StepThreeRQ()
 
         ISelectResults<object> results = region.Query<object>(qrystr.Query);
 
-        if (results.Size != QueryStatics.RegionQueryRowCounts[qryIdx])
+        if (results.Size != (ulong) QueryStatics.RegionQueryRowCounts[qryIdx])
```
mcmellawatt: Prefer 'as' keyword to c style cast in c#.

clicache/integration-test/ThinClientSecurityAuthzTestBaseN.cs
```diff
@@ -318,8 +318,8 @@ private bool CheckFlags(OpFlags flags, OpFlags checkFlag)
               {
                 Assert.AreEqual(numOps, queryResults.Size);
               }
-              //CacheableHashSet querySet = new CacheableHashSet(queryResults.Size);
-              List<string> querySet = new List<string>(queryResults.Size);
+
+              var querySet = new List<string>((int) queryResults.Size);
```
mcmellawatt: Prefer 'as' to c style casts in c#

tests/cli/NewFwkLib/CacheServer.cs
```diff
@@ -3533,7 +3533,7 @@ public void DoVerifyQueryResult()
       startTime = DateTime.Now;
       results = cq.ExecuteWithInitialResults(QueryResponseTimeout);
       elapsedTime = DateTime.Now - startTime;
-      if ((resultSize - localInvalidate) != results.Size)
+      if ((resultSize - localInvalidate) != (int) results.Size)
```
mcmellawatt: as

tests/cli/NewFwkLib/QueryTest/QueryTests.cs
```diff
@@ -560,7 +560,7 @@ private bool ReadQueryString(ref String query)
       bool isCq = GetBoolValue("cq");
       do
       {
-        int resultSize = GetUIntValue(ResultSize);  // set the query result size in xml
+        var resultSize = (ulong) GetUIntValue(ResultSize);  // set the query result size in xml
```
mcmellawatt: as

tests/cli/QueryHelper/QueryHelperN.cs
```diff
@@ -344,7 +341,7 @@ public virtual bool VerifyRS(ISelectResults<object> resultset, int expectedRows)
         }
 
         int foundFields = 0;
-        for (uint cols = 0; cols < si.Length; cols++)
+        for (int cols = 0; cols < si.Length; cols++)
```
mcmellawatt: can we do `var cols` here?

cppcache/include/geode/Struct.hpp
pivotal-jbarrett: Pure virtual class that can only be instantiated by it's sub-classes. Making it private doesn't really buy us anything but that's what was there before.

clicache/integration-test/ThinClientRegionQueryTests.cs
pivotal-jbarrett: Keeping consistent with other casts in C# sources. Also, my understanding of `as` is it would result in an auto boxing of value type returned by call and then be unboxed to compare with the value type of `result.Size`. Since `QueryStatics.RegionQueryRowCounts[]` contains non-nullable values using the `as` keyword is less desirable.

cppcache/integration-test/QueryHelper.hpp
pivotal-jbarrett: Not changing existing test code like that.

cppcache/integration-test/QueryHelper.hpp
pivotal-jbarrett: Probably but I am not super concerned with legacy test code at this point.

cppcache/integration-test/QueryHelper.hpp
pivotal-jbarrett: Yes, but I am not going to refactor existing legacy test code to modern C++.

clicache/integration-test/ThinClientRegionQueryTests.cs
pivotal-jbarrett: `as type` is to `dynamic_cast<type>` as `(type)` is to `static_cast<type>`... ish.

cppcache/integration-test-2/StructTest.cpp
mcmellawatt: It looks like the problem here is the `break` which causes the test to not check the value e.g. "one", "two", "three".

clicache/integration-test/ThinClientRegionQueryTests.cs
mcmellawatt: The real advantage of `as` over c-style cast is that a c-style cast throws an InvalidCastException while `as` will result in a null value which can then be checked gracefully.  It is nicer to see a test failure due to an ASSERT_NOT_NULL (not sure thats the actual assert method or not) than an InvalidCastException.

```
var resultsSize = QueryStatics.RegionQueryRowCounts[qryIdx] as ulong;

ASSERT_NOT_NULL(resultsSize, "Results size was not a ulong");
....
```

pivotal-jbarrett: The `resultSize` is the expected value not the actual value under test. The type just needs to be nudged to fix differences between APIs in C++ / C#.

mcmellawatt: Subtle difference - the next line would be `if (results.Size != resultsSize)` per my example above, but again probably doesn't buy us much here.

cppcache/integration-test/QueryHelper.hpp
mcmellawatt: Ack, that's what I get for eyeballing it :).

PR #289, opened 2018-04-26
GEODE-4854: Fix build error on Windows

austxcodemonkey:


clicache/src/CqAttributesMutator.hpp
```diff
@@ -105,7 +105,8 @@ namespace Apache
         /// <param name="nativeptr">The native object pointer</param>
         inline CqAttributesMutator<TKey, TResult>(native::CqAttributesMutator* nativeptr)
         {
-            m_nativeptr = gcnew native_unique_ptr<native::CqAttributesMutator>(nativeptr);
+            std::unique_ptr<native::CqAttributesMutator> mutator(new native::CqAttributesMutator(*nativeptr));
```
pivotal-jbarrett: This looks overly redundant. You already have a raw pointer to a heap allocated native CqAttributeMutator so why copy construct another one on the heap? You also do need to stack allocate a std::unique_ptr to construct the managed unique pointer below. See examples of use in other classes.

PR #299, opened 2018-05-23
GEODE-5209: Adding GCC code coverage

mcmellawatt:
Co-authored-by: Ryan McMahon <rmcmahon@pivotal.io>
Co-authored-by: Mike Martell <mmartell@pivotal.io>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ cmake_minimum_required(VERSION 3.10)
 project(nativeclient LANGUAGES C CXX)
 
 option(USE_PCH "Use precompiled headers (PCH)." OFF)
+option(USE_CPP_COVERAGE "Enable profiling and coverage report analysis for apache-geode cpp library" OFF)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
@@ -256,7 +257,7 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
 
   # Enables multiprocess compiles
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
-  
+
   # Enables whole program optimizaion and link time code generation for release.
   set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /GL")
   set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /LTCG:incremental")
--- a/cppcache/integration-test/CMakeLists.txt
+++ b/cppcache/integration-test/CMakeLists.txt
@@ -83,8 +83,7 @@ foreach(FILE ${SOURCES})
 
   if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" AND ${USE_CPP_COVERAGE})
       set_target_properties(${TEST}
-          PROPERTIES APPEND_STRING COMPILE_FLAGS "--coverage"
-                     APPEND_STRING LINK_FLAGS "--coverage")
+          PROPERTIES APPEND_STRING LINK_FLAGS "--coverage")
   endif()
 
   target_link_libraries(${TEST}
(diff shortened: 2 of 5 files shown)


mcmellawatt:
Need to look into excluding unwanted results (boost, etc)

mcmellawatt:
Excluding unwanted coverage is done when you invoke lcov, which is not part of this PR

CMakeLists.txt
```diff
@@ -256,7 +257,7 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
 
   # Enables multiprocess compiles
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
-  
+
```
pivotal-jbarrett: Can we avoid whitespace changes in the future.

cppcache/integration-test/CMakeLists.txt
```diff
@@ -81,6 +81,12 @@ foreach(FILE ${SOURCES})
   add_executable(${TEST} ${FILE})
   set_target_properties(${TEST} PROPERTIES FOLDER test/cppTests)
 
+  if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" AND ${USE_CPP_COVERAGE})
+      set_target_properties(${TEST}
+          PROPERTIES APPEND_STRING COMPILE_FLAGS "--coverage"
+                     APPEND_STRING LINK_FLAGS "--coverage")
```
pivotal-jbarrett: Are we sure that integration tests have to be compiled with coverage? Would it be sufficient to just link?

cppcache/shared/CMakeLists.txt
```diff
@@ -47,6 +47,12 @@ set_target_properties(apache-geode PROPERTIES
   CXX_VISIBILITY_PRESET hidden
   VISIBILITY_INLINES_HIDDEN ON)
 
+if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" AND ${USE_CPP_COVERAGE})
```
pivotal-jbarrett: Rather than duplicating these settings in each target how about using an `INTERFACE` target?

In root CMakeLists.txt:
```
add_library(INTERFACE _Coverage)
if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" AND ${USE_CPP_COVERAGE})
  set_target_properties(_Coverage PROPERTIES
    COMPILE_FLAGS "--coverage"
    LINK_FLAGS "--coverage")
 endif()
```

In other CMakeLists:
```
target_link_libraries(<target>
  ...
  PRIVATE
  ...
  _Coverage)
```

The bonus is that if we add support for other compilers and linkers we only have to change the root CMakeLists.txt. The other will inherit the coverage flags from he `_Coverage` interface target.

cppcache/static/CMakeLists.txt
```diff
@@ -16,21 +16,21 @@
 cmake_minimum_required(VERSION 3.4)
 project(apache-geode-static LANGUAGES CXX)
 
-add_library(apache-geode-static STATIC ${SOURCES_ALL})
+add_library(${PROJECT_NAME} STATIC ${SOURCES_ALL})
```
pivotal-jbarrett: Don't use ${PROJECT_NAME} like this. It breaks integration with CLion and other CMake integrations. Revert these changes.

CMakeLists.txt
mcmellawatt: We'll watch out for this next time

cppcache/integration-test/CMakeLists.txt
mcmellawatt: The lcov docs indicate that you need both:
http://ltp.sourceforge.net/coverage/lcov.php

cppcache/shared/CMakeLists.txt
mcmellawatt: Was able to implement that and I like that it reduces duplicated CMake code.

cppcache/static/CMakeLists.txt
mcmellawatt: Done

PR #302, opened 2018-05-25
GEODE-5209: Adds coverage to clang compilers.

pivotal-jbarrett:


CMakeLists.txt
```diff
@@ -242,6 +242,11 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     -Wno-documentation-unknown-command
     )
 
+  if (USE_CPP_COVERAGE)
+    target_compile_options(_CppCodeCoverage INTERFACE -fprofile-arcs -ftest-coverage)
```
austxcodemonkey: Do we need to explicitly use these two flags, as opposed to just '--coverage'?  The internet seems to indicate they are synonymous....
https://llvm.org/docs/CommandGuide/llvm-cov.html 
https://clang.llvm.org/docs/SourceBasedCodeCoverage.html
https://stackoverflow.com/questions/28840261/gcov-what-is-the-difference-between-coverage-and-ftest-coverage-when-buildi

It'd be nicer if everyone could use the same flags

pivotal-jbarrett: Sure we could do that.

PR #305, opened 2018-06-05
GEODE-5289: Update ComparePdxTypes.java to use public API instead of test hooks

echobravopapa:


tests/javaobject/ComparePdxTypes.java
```diff
@@ -174,7 +174,7 @@ private void saveAllJavaPdxTypes() {
     TypeRegistry registry = cache.getPdxRegistry();
     
     //this will clear pdx registry
```
pivotal-jbarrett: Captain obvious on the comments here...

PR #311, opened 2018-07-05
GEODE-4728: api docs - fix SystemProperties docs

davebarnes97:
Simple corrections identified [paired with @mmartell] back in May that weren't included in the May check-in.


igodwin:
Changes look great!

PR #319, opened 2018-07-23
Try reverting all the way back to cmake 3.10 and see if things get an…

austxcodemonkey:
…y better.


austxcodemonkey:
Removing/closing this PR for now.  The problem we're having with build machines is now due to a Chocolatey glitch with the Microsoft toolset install, and has remedied itself.

PR #324, opened 2018-08-06
Geode 4336

austxcodemonkey:
Add new C++ and C# examples for DataSerializable, PdxSerializable, and OQL


examples/cpp/dataserializable/Order.cpp
```diff
@@ -0,0 +1,60 @@
...
+void Order::toData(DataOutput& dataOutput) const {
+  dataOutput.writeInt(order_id_);
+  dataOutput.writeString(name_);
+  dataOutput.writeInt(quantity_);
+}
+
+std::string Order::toString() const {
+  std::stringstream strm;
+
+  strm << "OrderID: " << order_id_ << " Product Name: " << name_ << " Quantity: " << quantity_;
```
pivotal-jbarrett: I’d suggest using `std::to_string` and basic string concatenation over stream api for something this simple.

examples/cpp/dataserializable/Order.cpp
```diff
@@ -0,0 +1,60 @@
...
+  auto objectSize = sizeof(Order);
+  objectSize += name_.capacity();
+  return objectSize;
+}
+
+std::shared_ptr<DataSerializable> Order::create() {
+  return std::make_shared<Order>();
+}
+
+int32_t Order::getClassId() const {
```
pivotal-jbarrett: This should not be here anymore.

examples/dotnet/DataSerializableCs/Order.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public Int32 ClassId
+        {
+            get { return 0x42; }
+        }    
+
+        public ulong ObjectSize
+        {
+            get { return 0x04; }
```
pivotal-jbarrett: Where do you get this magic number? This doesn’t help anyone with an example.

examples/dotnet/DataSerializableCs/Order.cs
```diff
@@ -0,0 +1,74 @@
...
+        public void FromData(DataInput input)
+        {
+            OrderId = input.ReadInt32();
+            Name = input.ReadUTF();
+            Quantity = input.ReadInt16();
+        }
+
+        public Int32 ClassId
+        {
+            get { return 0x42; }
```
pivotal-jbarrett: ClassId is gone.

PR #331, opened 2018-08-13
Exclude build directory etc from source distribution ...

austxcodemonkey:
generated by CPack.  Ignore .git and .idea (and .DS_Store)


austxcodemonkey:
To build a source distro, from the build directory run 'cpack --config CPackSourceConfig.cmake'

igodwin:
I know we don't want to get too crazy with the list of possible things to exclude, but I vote that the default cmake build directory (/cmake-build-*) be added. We already include .idea...

CMakeLists.txt
```diff
@@ -107,6 +107,7 @@ set(CPACK_INCLUDE_TOPLEVEL_DIRECTORY 0)
 set(CPACK_PACKAGING_INSTALL_PREFIX "/${CPACK_PACKAGE_NAME}")
 set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/dist/LICENSE")
 set(CPACK_PACKAGE_INSTALL_DIRECTORY "${CPACK_PACKAGE_NAME}")
+set(CPACK_SOURCE_IGNORE_FILES "/build/;/docker/;/packer/;/\.git/;/\.idea/;/\.DS_Store/")
```
pivotal-jbarrett: `build` is not a standard excluded directory by any means. It just happens that you may have chosen to put your build in the source directory and called it `build` but other may not. 

Can we leverage the .gitignore files?

pivotal-jbarrett: Use `list(APPEND CPACK_SOURCE_IGNORE_FILES somefile)` to append to list rather than replacing the build it values with your values.

austxcodemonkey: We _could_ leverage the .gitignore file(s), but is it worth the extra effort?  Or should we just modify this list to more-or-less have parity with the .gitignore and move on?  `build` and `build-*` are in .gitignore as well, so it seems reasonable to have them here.

austxcodemonkey: We were able to use .gitignore as a starting point, I think it worked out well.

PR #332, opened 2018-08-14
GEODE-5554: Standardize naming of Geode Native example *.md files

davebarnes97:


examples/cpp/BUILD-CPP-EXAMPLES.md
```diff
@@ -0,0 +1,28 @@
...
+
+1. Copy the `examples` directory from the native client installation folder to a folder in your workspace.
+
+1. Navigate to the directory for a specific example:
+
+    ```
+    $ cd workspace/examples/cpp/<example>
+    $ mkdir build
+    $ cd build
+    $ cmake ..
```
austxcodemonkey: This requires additional parameters for Windows, specifically the `-G"Visual Studio 14 2015 Win64" -Thost=x64` business we're all used to.

examples/cpp/BUILD-CPP-EXAMPLES.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Prerequisites
+* Install [Apache Geode](https://geode.apache.org)
+* Install [CMake](https://cmake.org/download/)
+* Build and install [Apache Geode Native](https://github.com/apache/geode-native)
+* For Windows, Visual Studio 2015
+
+## Building the Examples
+
+1. Copy the `examples` directory from the native client installation folder to a folder in your workspace.
```
austxcodemonkey: Important safety tip, for Windows users - this needs to be copied to a folder that _does not include long filenames_.     This may or may not impact the .sh scripts you use to start/stop the server for cpp examples, but it _definitely_ doesn't work with LFNs for dotnet examples with the .ps1 scripts.  Best to just always use short filenames on Windows.

PR #334, opened 2018-08-15
Geode 4978 pass hash into cmake for version

austxcodemonkey:
Generating version.h now happens in configure step, and only changes when parameters change on cmake configure command line.  This prevents partial rebuilds that used to occur when you made a non-code change that nevertheless updated the Git hash.


mmartell:
Looks good. I like being able to pass in the source release info on the cmake configure step. Nice enhancement to the build process.

metatype:
Just curious, if I have a source release of geode-native does it contain the information needed to generate the version.h file?  IOW, can I take a source release and create binaries that have the correct version and SHA info embedded?

PR #335, opened 2018-08-22
Geode-2607: improve release artifacts

austxcodemonkey:
See notes in JIRA for full details.  tl;dr version: Geode includes their Docker/Packer files, so we will too, and the one-line change to CMakeLists.txt is (probably) the last thing we need for cpack.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,7 @@ set(CPACK_PACKAGING_INSTALL_PREFIX "/${CPACK_PACKAGE_NAME}")
 set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/dist/LICENSE")
 set(CPACK_PACKAGE_INSTALL_DIRECTORY "${CPACK_PACKAGE_NAME}")
 set(CPACK_GENERATOR TGZ;ZIP)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ set(CPACK_PACKAGING_INSTALL_PREFIX "/${CPACK_PACKAGE_NAME}")
 set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/dist/LICENSE")
 set(CPACK_PACKAGE_INSTALL_DIRECTORY "${CPACK_PACKAGE_NAME}")
 set(CPACK_GENERATOR TGZ;ZIP)
+set(CPACK_PACKAGE_CHECKSUM "SHA512")
```
echobravopapa: Isn't CMake nice!

echobravopapa: The ticket only asked for a SHA-256 hash, is 512 consistent with Geode?  Let's get the ticket updated accordingly.

metatype: 512 is fine...and actually faster on most cpus :-)

austxcodemonkey: GPG signature gen shouldn't be automated, per @metatype.  Release will be signed by release manager personally during release process. 

With respect to hash, I went with 512 because that's what the Geode source release uses.

PR #336, opened 2018-08-24
GEODE-5634: Xunit modules upgraded to latest

echobravopapa:


mmartell:
Yes looks good. Helps us with debugging our xunit tests locally in Visual Studio.

clicache/hippos
```diff
@@ -0,0 +1,350 @@
+./hippos
```
pivotal-jbarrett: What is this new hippos file?

PR #338, opened 2018-08-24
GEODE-5634: Reverting the Xunit upgrade

echobravopapa:
- found unexpected some incompatibilities, need to revisit upgrade at a later time


mmartell:
Looks like the revert didn't work. When I checkout develop branch, the following files still reference the new xunit versions:

- clicache/integration-test2/packages.config

- clicache/test2/packages.config

I see the previous xunt versions on a new GEODE-5634 branch.

PR #339, opened 2018-08-27
GEODE-5259: remove class id from data serializable

mmartell:


clicache/integration-test/SerializationTestsN.cs
```diff
@@ -92,19 +92,19 @@ public void CreateRegionForOT(string locators)
...
       CacheHelper.CreateTCRegion2<object, object>(RegionNames[0], true, false,
         null, locators, false);
-      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType.CreateDeserializable);
-      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType22.CreateDeserializable);
-      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType4.CreateDeserializable);
-      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType2.CreateDeserializable);
-      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType42.CreateDeserializable);
-      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType43.CreateDeserializable);
+      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType.CreateDeserializable, 0);
+      CacheHelper.DCache.TypeRegistry.RegisterType(OtherType22.CreateDeserializable, 0x8C0);
```
pivotal-jbarrett: What is the purpose of the class id in hex?

cppcache/include/geode/DataSerializable.hpp
```diff
@@ -48,14 +51,13 @@ class APACHE_GEODE_EXPORT DataSerializable : public virtual Serializable {
...
+   * @brief Return the type name of the instance being serialized.
    * This is used by deserialization to determine what instance
    * type to create and deserialize into.
    *
-   * The classId must be unique within an application suite.
-   * Using a negative value may result in undefined behavior.
+   * The type must be unique within an application suite.
    */
-  virtual int32_t getClassId() const = 0;
+  virtual const std::string getType() const { return typeid(*this).name(); }
```
pivotal-jbarrett: Why is this virtual? Does it need to be? Do you plan on allowing derived classes to override it?

cppcache/include/geode/TypeRegistry.hpp
```diff
@@ -44,7 +44,7 @@ class APACHE_GEODE_EXPORT TypeRegistry {
    * registered, or there is an error in registering the type; check errno
    * for more information in the latter case.
    */
-  void registerType(TypeFactoryMethod creationFunction);
+  void registerType(TypeFactoryMethod creationFunction, uint32_t id);
```
pivotal-jbarrett: Why an `uint32_t`? Isn't the Java API using a signed int?

cppcache/integration-test-2/DataSerializableTest.cpp
```diff
@@ -0,0 +1,156 @@
...
+      : name_(std::move(name)), csArray_(csArray), simple_(simple) {}
+
+  ~DataSerializableObject() noexcept override = default;
+
+  using DataSerializable::fromData;
+  using DataSerializable::toData;
+
+  void fromData(DataInput& dataInput) override {
+    name_ = dataInput.readString();
+    csArray_ =
```
pivotal-jbarrett: Why the ugliness of using an array of strings?

cppcache/integration-test/testSerialization.cpp
```diff
@@ -68,8 +68,7 @@ class OtherType : public DataSerializable {
   CData m_struct;
   int32_t m_classIdToReturn;
 
-  explicit OtherType(int32_t classIdToReturn = g_classIdToReturn)
-      : m_classIdToReturn(classIdToReturn) {
+  explicit OtherType() {
```
pivotal-jbarrett: Why `explicit` here? This is now a default constructer and could never implicitly be invoked.

cppcache/integration-test/testSerialization.cpp
```diff
@@ -161,9 +149,12 @@ DUNIT_TASK(Sender, SetupAndPutInts)
     auto serializationRegistry =
         CacheRegionHelper::getCacheImpl(cacheHelper->getCache().get())
             ->getSerializationRegistry();
-    serializationRegistry->addType(OtherType::createDeserializable);
-    serializationRegistry->addType(OtherType::createDeserializable2);
-    serializationRegistry->addType(OtherType::createDeserializable4);
+    serializationRegistry->addType(OtherType::createDeserializable,
+                                   g_classIdToReturn);
+    // serializationRegistry->addType(OtherType::createDeserializable2,
```
pivotal-jbarrett: Delete commented code.

cppcache/integration-test/testSerialization.cpp
```diff
@@ -180,9 +171,10 @@ DUNIT_TASK(Sender, SendCT)
   {
     for (int32_t i = 0; i < 30; i += 3) {
       try {
-        regionPtr->put(i, OtherType::uniqueCT(i, g_classIdToReturn));
-        regionPtr->put(i + 1, OtherType::uniqueCT(i + 1, g_classIdToReturn2));
-        regionPtr->put(i + 2, OtherType::uniqueCT(i + 2, g_classIdToReturn4));
+        regionPtr->put(i, OtherType::uniqueCT(i));
```
pivotal-jbarrett: Delete

cppcache/integration-test/testSerialization.cpp
```diff
@@ -193,8 +185,8 @@ DUNIT_TASK(Sender, RValidateCT)
   {
     for (int32_t i = 0; i < 30; i += 3) {
       OtherType::validateCT(i, regionPtr->get(i));
-      OtherType::validateCT(i + 1, regionPtr->get(i + 1));
-      OtherType::validateCT(i + 2, regionPtr->get(i + 2));
+      // OtherType::validateCT(i + 1, regionPtr->get(i + 1));
```
pivotal-jbarrett: Delete

cppcache/src/SerializationRegistry.hpp
```diff
@@ -130,6 +138,11 @@ class TheTypeMap : private NonCopyable {
   inline void unbindPdxType(const std::string& objFullName);
 
   void rebindPdxType(std::string objFullName, TypeFactoryMethodPdx func);
+
+ private:
+  void bind(TypeFactoryMethod func, DSCode id) {
+    bind(func, static_cast<int32_t>(id));
```
pivotal-jbarrett: You should make sure that `DataSerializable` , `DataSerializablePrimitive` and `DataSerializableFixedId` don't use the same map or they can walk on each others IDs. Previously they did some bit shift magic but we should now have distinct maps for the different types.

cppcache/src/SerializationRegistry.hpp
```diff
@@ -320,8 +334,12 @@ class APACHE_GEODE_EXPORT SerializationRegistry {
 
   inline void serialize(const DataSerializable* obj, DataOutput& output,
                         bool isDelta) const {
-    auto id = obj->getClassId();
-    auto dsCode = getSerializableDataDsCode(id);
+    auto&& type = obj->getType();
+    auto typeIterator = theTypeMap.typeToClassId.find(type);
```
pivotal-jbarrett: Fairly certain this should be `auto&` but safest to be `auto&&`.

cppcache/src/SerializationRegistry.hpp
```diff
@@ -320,8 +334,12 @@ class APACHE_GEODE_EXPORT SerializationRegistry {
...
+    auto id = typeIterator->second;
+
+    typeid(*obj).name();
```
pivotal-jbarrett: What is the purpose of this call. Am I missing some side effect of calling this function and not keeping the results?

cppcache/src/SerializationRegistry.hpp
```diff
@@ -320,8 +334,12 @@ class APACHE_GEODE_EXPORT SerializationRegistry {
...
+    auto dsCode = getSerializableDataDsCode((int32_t)id);
```
pivotal-jbarrett: If `DataSerializable` class ID is now `int32_t` then the call to getting the DSCode to encode the ID shouldn't need any casting right?

cppcache/src/SerializationRegistry.hpp
```diff
@@ -400,7 +418,7 @@ class APACHE_GEODE_EXPORT SerializationRegistry {
 
   [[noreturn]] void deserialize(DataInput& input,
                                 std::shared_ptr<PdxSerializable> obj) const;
-};
+};  // namespace client
```
pivotal-jbarrett: How is this a namespace here?

cppcache/src/TypeRegistry.cpp
```diff
@@ -25,8 +25,9 @@
 
 TypeRegistry::TypeRegistry(CacheImpl* cache) : m_cache(cache) {}
 
-void TypeRegistry::registerType(TypeFactoryMethod creationFunction) {
-  m_cache->getSerializationRegistry()->addType(creationFunction);
+void TypeRegistry::registerType(TypeFactoryMethod creationFunction,
+                                uint32_t id) {
```
pivotal-jbarrett: As above, pretty sure this should be `int32_t` to match Java/Server API.

tests/cli/NewTestObject/DeltaEx.cs
```diff
@@ -73,13 +73,9 @@ public void FromData(Apache.Geode.Client.DataInput DataIn)
             FromDataCount++;
         }
 
-        public Int32 ClassId
+        public String Type
```
pivotal-jbarrett: Why define this on .NET? All .NET classes already have a `GetType()` method that we can use, similar to the Java side where we use `getClass()`. Remove this method from the interface.

cppcache/include/geode/DataSerializable.hpp
pivotal-jbarrett: Also, why not return the `type_info` and use that rather than strings?

```
const std::type_info& getType() const {
    return typeid(*this);
  }
```

clicache/integration-test/SerializationTestsN.cs
mmartell: No purpose.

cppcache/include/geode/DataSerializable.hpp
mmartell: Shouldn't be virtual.

Returning type_info& doesn't work for managed user classes as it resolves to the exact same value even for different custom classes.

cppcache/include/geode/TypeRegistry.hpp
mmartell: The lower level type (SerializationRegistry.addType) is indeed an int32_t. Not sure why we had a uint here. Good catch.

cppcache/integration-test-2/DataSerializableTest.cpp
mmartell: We originally thought we had a serialization problem related to CacheableStringArray, which is one of the types in our Portfolio object. The only tests using CacheableStringArray were Cq tests (which are marked as FLAKY), so we wrote this very simple test.

cppcache/include/geode/DataSerializable.hpp
pivotal-jbarrett: Managed types should not influence unmanned types. Managed should be using `Object.GetType()` during mapping and serialization.

cppcache/integration-test-2/DataSerializableTest.cpp
pivotal-jbarrett: There are a ton of serialization tests. You are telling me that none of the PDX or DataSerializable tests have a string array in them?

clicache/integration-test/SerializationTestsN.cs
echobravopapa: Legacy "id"

cppcache/include/geode/TypeRegistry.hpp
echobravopapa: that's a mulligan

cppcache/integration-test-2/DataSerializableTest.cpp
echobravopapa: you might be surprised...

clicache/integration-test/SerializationTestsN.cs
echobravopapa: i.e. that was the ClassId that the `OtherType22` class always contained, we simply moved that info along to the new api

cppcache/include/geode/DataSerializable.hpp
echobravopapa: It became necessary for this to be`virtual` due to the need to make this call in managed and unmanaged code....  both managed/un-managed types end up in the same map

`std::type_info` has a deleted which meant that if we used `Object.getClass()` we would not be able to translate into a `std::type_info` that would be needed inside `cppcache`

`clicache` has the class `ManagedCacheableKeyGeneric` which derives from `native::DataSerializable` and does some marshaling between `clicahe` and `cppcache`
 
We went to using a string return value since ultimately calls to `clicache`'s `TypeRegistry::RegisterType(...)` and `cppcache`'s `SerializationRegistry::AddType(...)` end up in  common map held in `cppcache`

clicache/integration-test/SerializationTestsN.cs
pivotal-jbarrett: Ok, hex just made it look "special".

cppcache/include/geode/DataSerializable.hpp
pivotal-jbarrett: I really don't think it should be doing that. Ugh that design was all sorts of broken and shouldn't be dictating design between the two languages. Let me percolate on this.

cppcache/integration-test-2/DataSerializableTest.cpp
pivotal-jbarrett: *facepalm*

cppcache/integration-test/testSerialization.cpp
mmartell: This is preexisting test code. Our change was to remove the classId parameter.

cppcache/integration-test/testSerialization.cpp
mmartell: Done.

cppcache/integration-test/testSerialization.cpp
echobravopapa: done

cppcache/integration-test/testSerialization.cpp
echobravopapa: done

cppcache/src/SerializationRegistry.hpp
mmartell: Done.

cppcache/src/TypeRegistry.cpp
mmartell: Done.

cppcache/integration-test/testSerialization.cpp
pivotal-jbarrett: Right, and by removing that param you made `explicit` redundant. Please remove.

tests/cli/NewTestObject/DeltaEx.cs
mmartell: See previous comments above. 
```
        public String Type
        {
            get { return this.GetType().ToString(); }
        }
```

pivotal-jbarrett:
I still have strong reservations about the how the type information is reduced to a string. Please give me a day or two to think about this.

cppcache/include/geode/DataSerializable.hpp
pivotal-jbarrett: `const std::string` as a return type does not make sense. Each call to this method will allocate a new `std::string` you are then saying it must not be changed so the caller must allocate a new `std::string` to mutate or will accidentally do so if they don't remember to add const or use `auto&&`.

```
# results in single string allocation.
auto&& type = object.getType();

# results in 2 string allocations.
auto type = object.getType();

# results in 2 allocations.
std::string type = object.getType();

std::unordered_map<std::string, int32_t> map;
# results in 2-3 allocations of std::string.
map.emplace(object.getType(), someInt);
```

pivotal-jbarrett: If you eliminate the path where native code needs to access the class information for managed types then these issues can all be resolved with the `const std::type_info&` proposed. When serializing the lookup of type to classId should be happening in the managed layer anyway, so the table for type to classId in the unmanaged code should not be invoked at all. When deserializing we could hook earlier that we are actually managed code and divert to managed code for the type lookups. We do this already for PDX types.

cppcache/src/SerializationRegistry.cpp
```diff
@@ -126,11 +126,11 @@ std::shared_ptr<Serializable> SerializationRegistry::deserialize(
     DataInput& input, int8_t typeId) const {
   bool findinternal = false;
   auto typedTypeId = static_cast<DSCode>(typeId);
-  int64_t compId = typeId;
+  int32_t classId = typeId;
```
pivotal-jbarrett: There is no need to init the classId with the DSCode

cppcache/src/SerializationRegistry.cpp
```diff
@@ -126,11 +126,11 @@ std::shared_ptr<Serializable> SerializationRegistry::deserialize(
...
 
   if (typeId == -1) {
-    compId = input.read();
-    typedTypeId = static_cast<DSCode>(compId);
+    classId = input.read();
```
pivotal-jbarrett: This is a DSCode and not classId.

cppcache/src/SerializationRegistry.cpp
```diff
@@ -151,31 +151,29 @@ std::shared_ptr<Serializable> SerializationRegistry::deserialize(
...
     }
     case DSCode::CacheableUserData4: {
-      int32_t classId = input.readInt32();
-      compId |= ((static_cast<int64_t>(classId)) << 32);
+      classId = input.readInt32();
       break;
     }
     case DSCode::FixedIDByte: {
-      compId = input.read();
+      classId = input.read();
```
pivotal-jbarrett: This is a fixedId and probably shouldn't be confused with a classId. Maybe when you extract the handler functions it will be more clear.

cppcache/src/SerializationRegistry.cpp
```diff
@@ -189,22 +187,22 @@ std::shared_ptr<Serializable> SerializationRegistry::deserialize(
   TypeFactoryMethod createType = nullptr;
 
   if (findinternal) {
-    theTypeMap.find2(compId, createType);
+    theTypeMap.findInternal(classId, createType);
```
pivotal-jbarrett: This should be `findDataSerializableFixedId`.

cppcache/src/SerializationRegistry.cpp
```diff
@@ -189,22 +187,22 @@ std::shared_ptr<Serializable> SerializationRegistry::deserialize(
...
   } else {
-    theTypeMap.find(compId, createType);
+    theTypeMap.findSerializable(classId, createType);
```
pivotal-jbarrett: `findDataSerializable`

cppcache/src/SerializationRegistry.cpp
```diff
@@ -287,25 +285,21 @@ void SerializationRegistry::addPdxType(TypeFactoryMethodPdx func) {
   theTypeMap.bindPdxType(func);
 }
 
-void SerializationRegistry::addType(int64_t compId, TypeFactoryMethod func) {
-  theTypeMap.rebind(compId, func);
+void SerializationRegistry::addType(int32_t id, TypeFactoryMethod func) {
+  theTypeMap.rebind(id, func);
```
pivotal-jbarrett: This should probably be `addDataSerializableType(...)` and `rebindDataSerializable`

cppcache/src/SerializationRegistry.cpp
```diff
@@ -287,25 +285,21 @@ void SerializationRegistry::addPdxType(TypeFactoryMethodPdx func) {
...
 }
 
-void SerializationRegistry::removeType(int64_t compId) {
-  theTypeMap.unbind(compId);
-}
+void SerializationRegistry::removeType(int32_t id) { theTypeMap.unbind(id); }
```
pivotal-jbarrett: Similar to above.

cppcache/src/SerializationRegistry.cpp
```diff
@@ -287,25 +285,21 @@ void SerializationRegistry::addPdxType(TypeFactoryMethodPdx func) {
...
 
 void SerializationRegistry::addType2(TypeFactoryMethod func) {
```
pivotal-jbarrett: Should be `addDataSerializableFixedId(...)` and `rebindDataSerializableFixedId`.

cppcache/src/SerializationRegistry.cpp
```diff
@@ -287,25 +285,21 @@ void SerializationRegistry::addPdxType(TypeFactoryMethodPdx func) {
...
   theTypeMap.bind2(func);
 }
 
-void SerializationRegistry::addType2(int64_t compId, TypeFactoryMethod func) {
-  theTypeMap.rebind2(compId, func);
+void SerializationRegistry::addType2(int32_t id, TypeFactoryMethod func) {
```
pivotal-jbarrett: See above

cppcache/src/SerializationRegistry.cpp
```diff
@@ -287,25 +285,21 @@ void SerializationRegistry::addPdxType(TypeFactoryMethodPdx func) {
...
+  theTypeMap.rebind2(id, func);
```
pivotal-jbarrett: See above

cppcache/src/SerializationRegistry.cpp
```diff
@@ -287,25 +285,21 @@ void SerializationRegistry::addPdxType(TypeFactoryMethodPdx func) {
...
 }
 
-void SerializationRegistry::removeType2(int64_t compId) {
-  theTypeMap.unbind2(compId);
-}
+void SerializationRegistry::removeType2(int32_t id) { theTypeMap.unbind2(id); }
```
pivotal-jbarrett: See above

cppcache/src/SerializationRegistry.cpp
```diff
@@ -374,118 +368,121 @@ std::shared_ptr<Serializable> SerializationRegistry::GetEnum(
...
 void TheTypeMap::clear() {
-  std::lock_guard<util::concurrent::spinlock_mutex> guard(m_mapLock);
-  m_map->unbind_all();
+  std::lock_guard<util::concurrent::spinlock_mutex> guard(
+      m_UserDataSerializablesMapLock);
+  m_UserDataSerializablesMap->unbind_all();
 
-  std::lock_guard<util::concurrent::spinlock_mutex> guard2(m_map2Lock);
-  m_map2->unbind_all();
+  std::lock_guard<util::concurrent::spinlock_mutex> guard2(m_InternalsMapLock);
```
pivotal-jbarrett: Use `m_DataSerializableFixedIdMapLock`.

cppcache/src/SerializationRegistry.cpp
```diff
@@ -374,118 +368,121 @@ std::shared_ptr<Serializable> SerializationRegistry::GetEnum(
...
+void TheTypeMap::findSerializable(int32_t id, TypeFactoryMethod& func) const {
+  std::lock_guard<util::concurrent::spinlock_mutex> guard(
+      m_UserDataSerializablesMapLock);
+  m_UserDataSerializablesMap->find(id, func);
 }
 
-void TheTypeMap::find2(int64_t id, TypeFactoryMethod& func) const {
-  std::lock_guard<util::concurrent::spinlock_mutex> guard(m_map2Lock);
-  m_map2->find(id, func);
+void TheTypeMap::findInternal(int32_t id, TypeFactoryMethod& func) const {
```
pivotal-jbarrett: Similar theme of using names that reflect `DataSerializableFixedId` and `DataSerializable`.

PR #341, opened 2018-08-28
GEODE-5626: Reject getInitialValues=true in registerAllKeys when regi…

pdxcodemonkey:
…on is not caching.

    - Refactor Cluster class so we can apply locators separately.
    - Clean up server directories at test startup.
    - Add test cases for caching proxy and proxy regions with getInitialValues set to true

Co-authored-by: Ivan Godwin <igodwin@pivotal.io>

### DIFF
--- a/cppcache/integration-test-2/RegisterKeysTest.cpp
+++ b/cppcache/integration-test-2/RegisterKeysTest.cpp
@@ -1,0 +1,199 @@
+/* * Licensed to the Apache Software Foundation (ASF) under one or more *
+ * contributor license agreements.  See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to You under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <iostream>
+
+#include <gtest/gtest.h>
+
+#include <geode/CacheFactory.hpp>
+#include <geode/Cache.hpp>
+#include <geode/PoolManager.hpp>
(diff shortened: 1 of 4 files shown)


cppcache/integration-test-2/RegisterKeysTest.cpp
```diff
@@ -0,0 +1,199 @@
...
+    region2->registerKeys(keys, false, true);
+
+    auto&& entryAfterGet = std::dynamic_pointer_cast<CacheableInt16>(
+        region2->getEntry("one")->getValue());
+    ASSERT_NE(entryAfterGet, nullptr);
+    ASSERT_EQ(entryAfterGet->value(), 1);
+  }
+}
+
+/**
```
pivotal-jbarrett: Either add documentation to the test or don't but don't add empty documentation.

cppcache/integration-test-2/framework/Cluster.h
```diff
@@ -182,6 +182,8 @@ class Cluster {
         initialServers_(initialServers.get()) {
     jmxManagerPort_ = Framework::getAvailablePort();
 
+    // Clean up from previous run, if any, to ensure a clean server
```
pivotal-jbarrett: This comment is redundant to the statement below. Code should be expressive such that it doesn't need comments. Code needing preceding comments should be refactored to be more expressive.

cppcache/integration-test-2/framework/Cluster.h
```diff
@@ -208,32 +210,38 @@ class Cluster {
 
   void stop();
 
+  void removeServerDirectory() {
+    boost::filesystem::path serverDir = boost::filesystem::current_path().concat("/" + name_);
```
pivotal-jbarrett: Why do you need to `concat` and `+` to get a sub path?

cppcache/integration-test-2/framework/Cluster.h
```diff
@@ -208,32 +210,38 @@ class Cluster {
...
+    boost::filesystem::remove_all(serverDir);
```
pivotal-jbarrett: Will this not operate relative to the current working directory?

cppcache/integration-test-2/framework/Cluster.h
pivotal-jbarrett: I would like to see an integration test written to test this change.
I would also like to see one that asserts that `GfshCommand::execute()` throws an exception when `gfsh` exits non-zero.

PR #346, opened 2018-09-08
 GEODE-5691: Fix using namespace violations

pivotal-jbarrett:
* Updates clang-tidy integration using CMake built in support.
* Updates .clang-tidy.


pivotal-jbarrett:
@echobravopapa @pdxcodemonkey @mmartell Sorry for another bit commit. Don't focus on each file but rather the theme. This commit updates our .clang-tidy/.clang-format to the latest from CLang 6. It also fixes one of the biggest style violations, `using namespace ...`. All `using namespace ...` have been replaced with explicit `using ...` declarations.

pivotal-jbarrett:
In this task I’m only concerned with removing the offending code.

PR #347, opened 2018-09-10
GEODE-5707: CommitConflictException Test case.

vaijira:
Create 2 async tasks to force a CommitConflictException message.
After the exception the following tx should work.


cppcache/integration-test-2/CommitConflictExceptionTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    cache.getCacheTransactionManager()->begin();
+    begunTx1.set_value();
+    region->put("one", "one");
+    ASSERT_EQ(std::future_status::ready,
+              putTx2.get_future().wait_for(debug_safe(minutes(1))));
+    putTx1.set_value();
+    ASSERT_EQ(std::future_status::ready,
+              committedTx2.get_future().wait_for(debug_safe(minutes(1))));
+    try {
```
pivotal-jbarrett: Would `EXPECT_THROW` not work in this case?

vaijira: The problem with EXPECT_THROW or ASSERT_THROW is that i cannot assert that is a CommitConflictException only an apache::geode::client::Exception (at least i couldn't make it work with apache::geode::client::CommitConflictException) and to be sure that is a CommitConflictException i have to examine the content.

echobravopapa:
@vaijira any news on this PR?  if not please close and re-open when its active again, Thanks!

vaijira:
I added a new test and reproduce the issue.

pdxcodemonkey:
Took a look this morning and discovered the new test was always passing here, because the test code didn't _actually_ stop the server.  You'll want to cherry-pick over commit abaf4061514399b89949d9dfd3792208aee86cc7 from https://github.com/pdxcodemonkey/geode-native.git or apply a similar change to make the test work.  Can confirm, with the change the test is 100% accurate, always fails without the TXCleaner line and always passes with it.  Nicely done!

Thanks,
Blake

PR #348, opened 2018-09-10
GEODE-5713: fix valgrind uninitialized value

pdxcodemonkey:
When the simple example code listed in GEODE-5713 was run under Valgrind on Linux, it generated a complaint about use of an uninitialized variable in a loop in the statistics code.  This _appears_ to have been a red herring, since as far as we can tell the value in question was definitely initialized, but... cleaning things up a bit without changing semantics seems to get rid of the error message.

- replaced use of member var `numOfDescps` with a local var initialized from the contained type object
- removed (no longer used) member var `numOfDescps`
- removed gratuitous use of `this` pointer


pdxcodemonkey:
Commit messages fixed

PR #351, opened 2018-09-12
GEODE-5626: Fix crash in register all keys

pdxcodemonkey:
- Refactor Cluster class so we can apply locators separately.
- Clean up server directories at test startup.
- Add test cases for caching proxy and proxy regions with getInitialValues set to true
- remove extraneous comment and empty doc comments
- throw exception if gfsh exits non-zero
- use relative path when deleting server directory

### DIFF
--- a/cppcache/integration-test-2/RegisterKeysTest.cpp
+++ b/cppcache/integration-test-2/RegisterKeysTest.cpp
@@ -55,8 +55,6 @@ std::shared_ptr<Region> setupProxyRegion(Cache& cache) {
   return region;
 }
 
-/**
- */
 TEST(RegisterKeysTest, RegisterAllWithCachingRegion) {
   Cluster cluster{LocatorCount{1}, ServerCount{1}};
   cluster.getGfsh()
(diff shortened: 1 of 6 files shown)


cppcache/integration-test-2/RegisterKeysTest.cpp
```diff
@@ -0,0 +1,191 @@
+/* * Licensed to the Apache Software Foundation (ASF) under one or more *
```
pivotal-jbarrett: Too many `*` here. This is not a documentation block, just a comment block.

cppcache/integration-test-2/RegisterKeysTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+  cluster.getGfsh()
+      .create()
+      .region()
+      .withName("region")
+      .withType("PARTITION")
+      .execute();
+
+  {
+    auto cache = createTestCache();
+    auto pf =
```
moleske: `pf` isn't a great name for such otherwise beautiful new code.

pdxcodemonkey: Renamed all instances `poolFactory`

PR #356, opened 2018-09-18
GEODE-5751: Move 2 userguide source directories up one level in the directory tree.

davebarnes97:
Move 2 userguide source directories up one level in the directory tree.

../docs/geode-native-docs becomes ../geode-native-docs
../docs/geode-native-book becomes ../geode-native-book
../docs directory remains with other content (../docs/api and ../docs/docker)


igodwin:
The restructuring looks good, though I would ask that you rename the PR to match current convention.

davebarnes97:
I agree with reviewer @mmartell - this directory change is really a workaround for an issue with bookbinder. As such, it's an imperfect and temporary solution.
I will withdraw this PR in favor of a more lasting fix to the tool itself.

PR #360, opened 2018-09-21
GEODE-5259: Fix clang (xcode 10) build breaks

pdxcodemonkey:
- Remove problematic clang-format options
- Anonymize unused parameters
- Remove unnecessary type cast

### DIFF
--- a/cppcache/src/SerializableHelper.hpp
+++ b/cppcache/src/SerializableHelper.hpp
@@ -93,7 +93,7 @@ inline bool SerializableHelper<PdxSerializable>::equalTo(
 
 template <>
 inline bool SerializableHelper<DataSerializableInternal>::metadataEqualTo(
-    const DataSerializableInternal& lhs, const DataSerializableInternal& rhs) {
+    const DataSerializableInternal&, const DataSerializableInternal&) {
   return true;
 }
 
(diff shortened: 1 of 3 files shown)


cppcache/src/SerializableHelper.hpp
```diff
@@ -93,7 +93,7 @@ inline bool SerializableHelper<PdxSerializable>::equalTo(
 
 template <>
 inline bool SerializableHelper<DataSerializableInternal>::metadataEqualTo(
-    const DataSerializableInternal& lhs, const DataSerializableInternal& rhs) {
+    const DataSerializableInternal&, const DataSerializableInternal&) {
```
echobravopapa: I think a more complete refactor for this `metadataEqualTo` is in order... look at the call site of this method and let's think about if `DataSerializables` have meta data for comparison, maybe it all just gets removed...?  and we'll want a test to drive out the change in any case as there was not legit coverage for the previous change.

pivotal-jbarrett: The other `DataSerializable`* have metadata to compare. Rather than rewriting the common bits of the comparison in each type we use this template class to do the compare. At compile time this results in a no-op for the `DataSerializableInternal`.

PR #361, opened 2018-09-21
GEODE-5750: Fix incorrect optype enum value

pdxcodemonkey:
- Also fix issue with CQ and AppDomains

### DIFF
--- a/clicache/integration-test2/CqOperationTest.cs
+++ b/clicache/integration-test2/CqOperationTest.cs
@@ -18,16 +18,60 @@
 using System;
 using System.IO;
 using Xunit;
+using System.Diagnostics;
 
 namespace Apache.Geode.Client.IntegrationTests
 {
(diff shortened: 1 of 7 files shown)


clicache/integration-test2/CqOperationTest.cs
```diff
@@ -0,0 +1,220 @@
...
+        private readonly Cache _cacheOne;
+        private readonly GeodeServer _geodeServer;
+        private static int _waitInterval = 1000;
+
+        public CqOperationTest()
+        {
+            var cacheFactory = new CacheFactory()
+                .Set("log-level", "error");
+
+            _cacheOne = cacheFactory.Create();
```
pivotal-jbarrett: Do you have more than one cache??

igodwin:
Requested change to client cache name has been made.

PR #362, opened 2018-09-22
Fix cli integration-test2 flakiness

mmartell:
Allows these tests to be run in parallel. Also reformats the source file.

### DIFF
--- a/clicache/integration-test2/GeodeServer.cs
+++ b/clicache/integration-test2/GeodeServer.cs
@@ -7,199 +7,214 @@ using Xunit;
 
 public class GeodeServer : IDisposable
 {
-    #region Properties/Fields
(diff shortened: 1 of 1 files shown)


mmartell:
It’s not just SSL. I got other tests to fail, but far less often. The Dispose was shutting down the wrong server. Connecting on the Jmx-manager port to shutdown the locator seems to fix it, but I don’t really understand why. I’m hoping you can explain.

pivotal-jbarrett:
Does it still hang even with the jmx port for shutdown too? Does the C++ use jmx port to connect for shutdown?

mmartell:
Yah looks like still not fixed. Still can get a region destroyed message from server.

PR #366, opened 2018-10-03
GEODE-5025: Uses templates to define Cacheable types

pivotal-jbarrett:
Replaces the old macros with templates to define `Cacheable` built in types. This allows you to debug into the types should there be issues.


pivotal-jbarrett:
@echobravopapa Please re-review.

PR #370, opened 2018-10-08
GEODE-3806: Removes use of Xerces.

pivotal-jbarrett:
Removes unused sources.


cppcache/integration-test/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ target_link_libraries(${TEST_UTILS_LIB}
     _WarningsAsError
   PUBLIC
     apache-geode
-    fwk
+    framework
```
echobravopapa: letters used to be so expensive

pivotal-jbarrett: Oh, no it is worse. There was a directory, deleted by this, called `fwk` that linked to this library. This library was always named `framework` but was a transitive dependency through `fwk`. The `framework` library is in a directory called, wait for it.... "fwklib".

PR #372, opened 2018-10-08
 GEODE-5770: Fixes clang-analyzer-optin.performance.Padding warning

pivotal-jbarrett:


.clang-tidy
```diff
@@ -1,6 +1,6 @@
 ---
 Checks:          '-*,clang-diagnostic-*,clang-analyzer-*,-clang-analyzer-alpha*,google-*,-google-readability-todo,-google-runtime-references,-google-default-arguments'
-WarningsAsErrors: 'google-build-using-namespace,google-readability-redundant-smartptr-get,google-explicit-constructor,google-global-names-in-headers,google-runtime-int,google-readability-casting'
+WarningsAsErrors: 'google-build-using-namespace,google-readability-redundant-smartptr-get,google-explicit-constructor,google-global-names-in-headers,google-runtime-int,google-readability-casting,google-readability-namespace-comments,clang-analyzer-optin.performance.Padding'
```
pdxcodemonkey: What exactly does this do?  It looks like you've swapped around initialization order and perhaps declaration order of a bunch of member variables in several classes.  I'm familiar with having to do this quite often to get rid of compiler warnings when using gcc, but this is Clang and referring to "performance.padding" - does this gripe if you're declaring pointers on unaligned boundaries in your class or something?  I didn't think any modern compilers did this...

pdxcodemonkey: Okay yeah Google tells me it reduces the size of your class by reordering items to avoid padding.  Seems a little obsessive tbh, but it's not a huge change to be compliant so why not....

pivotal-jbarrett: What this is doing is trying to avoid a lot of unnecessary padding that can bloat the size of you objects in memory. The order is important because certain types will have to pad out to fix the alignment types. For example, I can't have an int that starts after a single boolean without padding out at least 3 bytes.

For example:
```C++
bool a;
int b;
bool c;
```
Results in 3 bytes of padding after `a` and 3 bytes after `c` for a total of 6 bytes of padding.

Reordering to:
```C++
bool a;
bool c;
int b;
```
Results in 2 bytes of padding after `c` only.

PR #374, opened 2018-10-09
GEODE-5693: Use clang-tidy in travis build.

pivotal-jbarrett:


pivotal-jbarrett:
See output from travis build. Sadly it takes about 2h to run. I was able to get travis.ci to increase our build timeout to accommodate. Hopefully as we improve includes that build time will improve.

PR #376, opened 2018-10-10
GEODE-5768: New ContinuousQuery tests using DataSerializable objects

mmartell:

### DIFF
--- a/clicache/integration-test2/CqOperationTest.cs
+++ b/clicache/integration-test2/CqOperationTest.cs
@@ -91,12 +91,10 @@ namespace Apache.Geode.Client.IntegrationTests
 
         public virtual void OnError(CqEvent<TKey, TResult> ev)
         {
-            Debug.WriteLine("CqListener::OnError called");
         }
 
         public virtual void Close()
         {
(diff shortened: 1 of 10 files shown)


clicache/integration-test2/CqOperationTest.cs
```diff
@@ -87,8 +87,25 @@ public MyCqListener()
 
         public virtual void OnEvent(CqEvent<TKey, TResult> ev)
         {
-            Debug.WriteLine("MyCqListener::OnEvent called");
-            MyOrder val = ev.getNewValue() as MyOrder;
+        }
+
+        public virtual void OnError(CqEvent<TKey, TResult> ev)
+        {
+            Debug.WriteLine("CqListener::OnError called");
```
pivotal-jbarrett: Tests shouldn't be writing things to console. Shouldn't there be some flag this call is setting so that you can assert it was called later?

echobravopapa: We'll clean that up

echobravopapa:
@pivotal-jbarrett those debug statements are cleaned up and GEODE-5746 will take care of your other concerns.

PR #377, opened 2018-10-10
 GEODE-5847: Integrate clang-format into CMake.

pivotal-jbarrett:
Depends on #374


mmartell:
This looks awesome! I assume this corrects any IDE formatting issues so that as long as I compile before I check-in, they will be corrected.

pivotal-jbarrett:
If clang-format is found, yes.

PR #380, opened 2018-10-16
GEODE-5746: deflakify cli integration test2

pdxcodemonkey:
- Refactor integration-test-2 framework for CLI to work like the C++ framework (mostly)
- Make DataSerializable Cq test work again with new framework
- verify tests pass in parallel and clean up after themselves properly when done


cppcache/integration-test-2/framework/Cluster.cpp
```diff
@@ -70,8 +70,8 @@ void Server::start() {
       .withBindAddress(serverAddress_.address)
       .withPort(serverAddress_.port)
       .withMaxHeap("1g")
-      .withLocators(locators_.front().getAdddress().address + "[" +
```
moleske: Wait it wasn't supposed to be `getAddddress()` :-p

mmartell:
I love the new builder pattern here. I'll take it on faith that this fixes the problems seen with running these tests in parallel. Even if there are remaining issues it should be much easier to debug with this new design.

cppcache/integration-test-2/framework/Cluster.cpp
echobravopapa: right, that was for very long addresses...

PR #384, opened 2018-10-25
GEODE-5251: parameterize assembly info

pdxcodemonkey:


CMakeLists.txt
```diff
@@ -43,7 +43,8 @@ set(BUILD_BITS 64 CACHE STRING "Build for 64 (Geode default) or 32 bit.")
 
 set(PRODUCT_VENDOR "Apache" CACHE STRING "Product vendor")
 set(PRODUCT_VENDOR_NAME "The Apache Software Foundation" CACHE STRING "Product vendor full legal name")
-set(PRODUCT_NAME "Geode Native" CACHE STRING "Product name")
+set(PRODUCT_BASE_NAME "Geode" CACHE STRING "Product base name")
+set(PRODUCT_NAME "${PRODUCT_BASE_NAME} Native" CACHE STRING "Product name")
```
echobravopapa: Looks like this was done to be able to string build vendor.basename...

CMakeLists.txt
```diff
@@ -52,7 +53,7 @@ set(PRODUCT_SOURCE_REPOSITORY "<unspecified>" CACHE STRING "Product source branc
 set(PRODUCT_BITS "${BUILD_BITS}bit")
 
 set(PRODUCT_LIB_NAME "apache-geode" CACHE STRING "Binary name")
-set(PRODUCT_DLL_NAME "Apache.Geode" CACHE STRING ".Net Binary name")
+set(PRODUCT_DLL_NAME "${PRODUCT_VENDOR}.${PRODUCT_BASE_NAME}" CACHE STRING ".Net Binary name")
```
echobravopapa: Not sure we need this... given the line above was not changed..

CMakeLists.txt
mmartell: @pdxcodemonkey do you have any updates?

pdxcodemonkey: Waiting for a response to a question about Copyright string etc. then I'll update...

CMakeLists.txt
pdxcodemonkey: We don't specifically need this in CMakeLists.txt, but per the above we need PRODUCT_BASE_NAME in the assembly info, so may as well use it if we have it.

PR #385, opened 2018-10-26
GEODE-5942: prefix variables in PersistenceManager

sboorlagadda:
PAGE_SIZE causes conflicts with system library header files.
  So adding a PM prefix to all variables in PersistenceManager.

Signed-off-by: Matthew Reddington <mreddington@gmail.com>


pivotal-jbarrett:
See https://github.com/apache/geode-native/pull/387/commits/2fb3838dfc5f4249d0479586af1ed75a14606f64

pivotal-jbarrett:
Well this sucks. Even after making this change this `PAGE_SIZE` name still trips you up from time to time on MacOS. I suggest we change it but but suffixing with `_KEY` since these are used as keys in a property object.

pdxcodemonkey:
This is superseded by PR #393, closing.

PR #387, opened 2018-10-30
GEODE-5025: Fixes template visiblity when linked with executables.

jake-at-work:


cppcache/include/geode/CacheableBuiltins.hpp
```diff
@@ -42,30 +42,14 @@ using internal::CacheableArrayPrimitive;
 using internal::CacheableContainerPrimitive;
 using internal::CacheableKeyPrimitive;
```
pdxcodemonkey: Do these belong in a header file?  I believe we should lose them.  If not, at least make the usage consistent and get rid of `internal::` prior to references to the types we're using.

mreddington: If you have to include an internal namespace in a public namespace, if you have to ship internal headers along with the public API headers, then your internal is public. I recommend we move these classes into the public API space.

pivotal-jbarrett: The use of "internal" namespaces and classes in public headers is consistent with other C++ libraries, most notably Boost. Boost uses the namespace "detail" for detail specific/internal headers.

PR #388, opened 2018-10-30
GEODE-5316: Updated protocol version to Geode 1.8.0

mmartell:


echobravopapa:
@mmartell I think you can close this, should be superceded by https://github.com/apache/geode-native/pull/392

PR #389, opened 2018-10-31
GEODE-4337: example cpp function execution

pdxcodemonkey:
- Simplified example to only call one server function
- Added example.jar file containing only necessary function
- Stopped installing test javaobject.jar file


examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,113 @@
...
-        std::cout << buf;
-        ++i;
+    auto key = CacheableKey::create(keyStream.str().c_str());
+    regionPtr->put(key, value);
+  }
+}
+
+std::shared_ptr<CacheableVector> populateQueryObject() {
+  auto routingObj = CacheableVector::create();
+  for (int i = 1; i < EXAMPLE_ITEM_COUNT; i+=2) {
```
pivotal-jbarrett: Why every other value? Why not fixed values here. This stinks like the integration test. This should be a good clean room function example.

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,113 @@
...
+  }
+  return routingObj;
+}
+
+std::vector<std::string> executeFunctionOnServer(const std::shared_ptr<Region> regionPtr,
+    const std::shared_ptr<CacheableVector> queryObject) {
+  std::vector<std::string> resultList;
+
+  auto functionService = FunctionService::onServer(regionPtr->getRegionService());
+  if(auto executeFunctionResult = functionService.withArgs(queryObject).execute(getFuncIName)->getResult()) {
```
pivotal-jbarrett: What's a `queryObject`? The method is `withArgs` so why not use a variable name that matches? `arguments`?

examples/utilities/ExampleMultiGetFunction.java
```diff
@@ -0,0 +1,73 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package javaobject;
+
+
+import java.util.*;
+import java.io.*;
+import org.apache.geode.*; // for DataSerializable
```
pivotal-jbarrett: Import what you need, not *.

examples/utilities/ExampleMultiGetFunction.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.execute.FunctionAdapter;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.cache.execute.ResultSender;
+
+public class ExampleMultiGetFunction extends FunctionAdapter implements Declarable{
+
+  public void execute(FunctionContext context) {
+    ArrayList vals = new ArrayList();
+    if (context.getArguments() instanceof Boolean) {
```
pivotal-jbarrett: Why do we have different types in this argument checking? Your example only sends strings.

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,113 @@
...
+      .create();
+}
 
-const auto getFuncIName = std::string("MultiGetFunctionI");
-const auto putFuncIName = std::string("MultiPutFunctionI");
-const auto getFuncName = std::string("MultiGetFunction");
+void createPool(const Cache& cache) {
+  auto pool = cache.getPoolManager()
+      .createFactory()
+      .setSubscriptionEnabled(true)
```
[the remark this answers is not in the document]
pdxcodemonkey: Removed

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,113 @@
...
+  return regPtr0;
+}
 
-    auto routingObj = CacheableVector::create();
-    for (int i = 1; i < EXAMPLE_ITEM_COUNT; i+=2) {
-      sprintf(buf, "KEY--%d", i);
-      auto key = CacheableKey::create(buf);
-      routingObj->push_back(key);
-    }
+void populateRegion(const std::shared_ptr<Region>& regionPtr) {
```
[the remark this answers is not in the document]
pdxcodemonkey: Fixed

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,113 @@
...
+  for (int i = 0; i < EXAMPLE_ITEM_COUNT; i++) {
+    std::stringstream keyStream;
+    std::stringstream valueStream;
+    keyStream << "KEY--" << i;
```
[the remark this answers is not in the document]
pdxcodemonkey: Using fixed keys and values cleans this up nicely - thanks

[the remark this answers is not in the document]
pdxcodemonkey: Yep

examples/cpp/function-execution/main.cpp
pdxcodemonkey: Fixed

examples/utilities/ExampleMultiGetFunction.java
pdxcodemonkey: Um, because I just lifted existing Java code from javaobject.jar?.  I'll clean this up.

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,114 @@
...
+const int EXAMPLE_SERVER_PORT = 50505;
+
+std::vector<std::string> keys = {
+    "KEY--1",
+    "KEY--2",
+    "KEY--3",
+};
+
+std::vector<std::string> values = {
+    "VALUE--1",
```
pivotal-jbarrett: Why not have the key and values be more relatable than this. Like how does "KEY--1" relate to "VALUE--3". Seems so random.

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,114 @@
...
+using apache::geode::client::CacheableVector;
+using apache::geode::client::CacheFactory;
+using apache::geode::client::Exception;
+using apache::geode::client::FunctionService;
+using apache::geode::client::Region;
+using apache::geode::client::RegionShortcut;
+
+const auto getFuncIName = std::string("ExampleMultiGetFunction");
+
+const int EXAMPLE_ITEM_COUNT  = 6;
```
pivotal-jbarrett: Does this still have a place now that we have fixed set of keys and values?

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,114 @@
...
+const int EXAMPLE_SERVER_PORT = 50505;
+
+std::vector<std::string> keys = {
```
pivotal-jbarrett: `const`?

examples/cpp/function-execution/main.cpp
```diff
@@ -16,20 +16,16 @@
...
- * 4. Create Execute Objects
- * 5. Execute Functions
- * 6. Close the Cache.
- *
+ * 4. Create Execute Object
+ * 5. Execute Function
  */
 #include <iostream>
 #include <memory>
+#include <sstream>
```
pivotal-jbarrett: Still need this?

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,114 @@
...
-      for (auto &cachedString: *resultList) {
-        sprintf(
-            buf, "get result[%d]=%s\n", i,
-            std::dynamic_pointer_cast<CacheableString>(cachedString)->value().c_str());
-        std::cout << buf;
-        ++i;
+std::shared_ptr<CacheableVector> populateArguments() {
+  auto arguments = CacheableVector::create();
+  for (int i = 0; i < keys.size(); i++) {
+    arguments->push_back(CacheableKey::create(keys[i]));
```
pivotal-jbarrett: I think you can send vector of `std::string`. You should try it. If not you should open a ticket to support it.

examples/cpp/function-execution/main.cpp
```diff
@@ -38,147 +34,114 @@
...
+  }
+  return arguments;
+}
+
+std::vector<std::string> executeFunctionOnServer(const std::shared_ptr<Region> regionPtr,
```
pivotal-jbarrett: Stop the `Ptr` suffix.

examples/cpp/function-execution/main.cpp
pdxcodemonkey: Yeah, just changed values to 1,2,3 as well - makes more sense, at least they agree now.

examples/cpp/function-execution/main.cpp
pdxcodemonkey: Nope - removed.

examples/cpp/function-execution/main.cpp
pdxcodemonkey: Nope.

examples/cpp/function-execution/main.cpp
pdxcodemonkey: At first blush it doesn't appear to work.  GEODE-6004 filed for tracking.

examples/cpp/function-execution/main.cpp
pdxcodemonkey: Fixed.

PR #390, opened 2018-11-02
GEODE-5968: Fixing license issues

echobravopapa:


pivotal-jbarrett:
There is a commit with no JIRA

PR #391, opened 2018-11-03
GEODE-5968 Updated missing source headers

metatype:
Remove BOM so that rat can check for correct headers. Add cotire
reference to LICENSE since it is included in the source and under
an MIT license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -209,3 +209,29 @@ with separate copyright notices and license terms. Your use of those
 components are subject to the terms and conditions of the following
 licenses.
 
+---------------------------------------------------------------------------
+The MIT License (http://opensource.org/licenses/mit-license.html)
+---------------------------------------------------------------------------
+
+Apache Geode bundles the following files under the MIT license:
+
+  - cotire (https://github.com/sakra/cotire), Copyright (c) 2012-
+    2018 Sascha Kratky
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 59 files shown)


LICENSE
```diff
@@ -209,3 +209,29 @@ with separate copyright notices and license terms. Your use of those
 components are subject to the terms and conditions of the following
 licenses.
 
+---------------------------------------------------------------------------
+The MIT License (http://opensource.org/licenses/mit-license.html)
+---------------------------------------------------------------------------
+
+Apache Geode bundles the following files under the MIT license:
+
+  - cotire (https://github.com/sakra/cotire), Copyright (c) 2012-
```
pivotal-jbarrett: Can you explain why something like Cotire would be included in the LICENSE file? It is used only at build time. Is it because it will be shipped in source form with the source distribution?

examples/dotnet/ContinuousQueryCs/Program.cs
```diff
@@ -1,4 +1,4 @@
-﻿/*
```
pivotal-jbarrett: Were these all BOM issues?

LICENSE
metatype: Exactly.  The LICENSE file needs to declare things that are bundled in the release (the source distribution).  If we choose to provide a convenience binary, there would be a second LICENSE file that also includes any bundled binary dependencies.

examples/dotnet/ContinuousQueryCs/Program.cs
metatype: Yep, the GitHub review doesn't show the control chars.

LICENSE
pivotal-jbarrett: Would it be preferable to change the source to fetch Cotire from GitHub when PCH is enabled at configure time?

metatype: Either way is fine in this case.  If the bundled dependency were Category X (e.g. GPL) you would need take the optional download approach.

PR #395, opened 2018-11-07
GEODE-6012: Corrected package name in Dockerfile

igodwin:
- libc++api-dev (non-existent) changed to libc++abi-dev

Docker image successfully builds and geode-native is able to be built on resulting image.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,14 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu
+#FROM ubuntu
+FROM ubuntu:16.04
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -14,14 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu
+#FROM ubuntu
+FROM ubuntu:16.04
```
pivotal-jbarrett: Why lock this down to an older version of ubuntu?

igodwin: That's a mistake, leftover from testing on 16.04. I've pushed the change and payed attention, this time.

PR #400, opened 2018-11-09
GEODE-5957: Parse (previously) unknown server error messages

pdxcodemonkey:
- All are just essentially string error messages
- Provides more detail about what happened on server

### DIFF
--- a/cppcache/integration-test-2/FunctionExecutionTest.cpp
+++ b/cppcache/integration-test-2/FunctionExecutionTest.cpp
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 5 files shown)


cppcache/integration-test-2/FunctionExecutionTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+      .create()
+      .region()
+      .withName("region")
+      .withType("REPLICATE")
+      .execute();
+
+  auto cache = cluster.createCache();
+  auto region = setupRegion(cache);
+
+  try {
```
moleske: Is `EXPECT_THROW` not sufficient?  A `try` statement in a test always makes me a little jumpy

pdxcodemonkey: Fixed.

PR #404, opened 2018-11-14
GEODE-6054: Build with Visual Studio 2017 runtime.

pdxcodemonkey:
- The only build error was a warning-as-error failure due to a gtest header.  Well-known problem, fixed with a cmake variable in the gtest CMakeLists.txt

FWIW, This is fixed in a gtest version newer than 1.8, but 1.8 is what we're currently on:  https://github.com/google/googletest/issues/1111


dependencies/gtest/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "SunPro")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -m64")
 endif()
 
+if(WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -m64 /D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING")
```
pivotal-jbarrett: What is giving this warning in our sources?

pdxcodemonkey: It's in gtest: gtest-port.h

pivotal-jbarrett: This isn't how you do `/D` or `-D` flags in CMake. Also it should be defined in the gtest library so that anything compiling/linking with gtest gets this flag.

pivotal-jbarrett: See `boost/CMakeList` for
```
target_compile_definitions(${PROJECT_NAME} INTERFACE
  BOOST_ALL_NO_LIB
)
```
and add it there.
```
target_compile_definitions(${PROJECT_NAME} INTERFACE
  BOOST_ALL_NO_LIB
  _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING
)
```
Even better would be to wrap around `if (MSVC)` conditional.

pdxcodemonkey:
Gonna close this for a bit - we still have quite a bit of work to do, it seems...

PR #406, opened 2018-11-19
GEODE-4728: Removed all usage of grid client

echobravopapa:


cppcache/integration-test/testThinClientRemoteQuerySS.cpp
```diff
@@ -226,7 +226,7 @@ void compareMaps(HashMapOfCacheable &map, HashMapOfCacheable &expectedMap) {
 }
 
 void stepOne() {
-  initGridClient(true);
+  initClient(true);
```
pivotal-jbarrett: This may be a redundant test if it was initializing grid client. Is there another test that covers these same methods but didn't init grid?

cppcache/integration-test/testThinClientTransactionsWithSticky.cpp
```diff
@@ -19,9 +19,6 @@
 
 DUNIT_MAIN
   {
-    CALL_TASK(Alter_Client_Grid_Property_1);
```
pivotal-jbarrett: Does removing this create a duplicate test where the same methods are being exercised already without grid set?

cppcache/integration-test/testThinClientTransactionsWithoutSticky.cpp
```diff
@@ -19,9 +19,6 @@
 
 DUNIT_MAIN
   {
-    CALL_TASK(Alter_Client_Grid_Property_1);
```
pivotal-jbarrett: Duplicate test?

cppcache/integration-test/testThinClientTransactionsWithSticky.cpp
echobravopapa: No, just removes non-sense of toggling `grid-client`

cppcache/integration-test/testThinClientTransactionsWithoutSticky.cpp
echobravopapa: see above...

echobravopapa:
@pivotal-jbarrett please see responses

PR #407, opened 2018-11-20
GEODE-6054: build with vs 2017

mmartell:


cppcache/CMakeLists.txt
```diff
@@ -41,6 +41,14 @@ set_target_properties(version-header PROPERTIES
 	FOLDER cpp
 )
 
+if (WIN32)
+  add_compile_definitions(WIN32)
```
pivotal-jbarrett: I am 99% certain that is already defined. I know 100% the _WIN32 is.

cppcache/CMakeLists.txt
```diff
@@ -41,6 +41,14 @@ set_target_properties(version-header PROPERTIES
...
+endif()
+
+IF(MSVC)
+    SET(CMAKE_CXX_FLAGS "/EHsc")
```
pivotal-jbarrett: This should be set at the root project so all inherit now if it’s required on all. Otherwise I think there is a set target property for this. At the very least if you’re changing this variablenit should be appended to the current value and not replacing.

mmartell: Moved this to root.

mmartell:
Thanks pivotal-jbarrett for a diligent review! All changes have been implemented.

mmartell:
I think some new Warnings as Errors shows up with the new (and likely stricter compiler). I will forward to you as soon as I finish this episode of Ozark. Not sure about using a different exception model between the managed and unmanaged sides.

mmartell:
We were seeing a C4530 Warning:
https://msdn.microsoft.com/en-us/library/2axwkyt4.aspx

But I can't reproduce it. Trying a build without the /EHsc. Also, it seems
like this would be the default for unmanaged but I can't find docs on that.

mmartell:
Reverted the change to a different exception model. All issues addressed with latest commit.

echobravopapa:
@mmartell if this has stalled you might consider closing the PR until ready for review again...

dependencies/gtest/CMakeLists.txt
```diff
@@ -44,6 +44,10 @@ set( ${PROJECT_NAME}_gtest_BINARY_DIR ${BINARY_DIR}/googlemock/gtest/${_DEBUG_OR
 set( ${PROJECT_NAME}_gmock_BINARY_DIR ${BINARY_DIR}/googlemock/${_DEBUG_OR_RELEASE} )
 set( DEPENDENCIES_${PROJECT_NAME}_DIR ${${PROJECT_NAME}_BINARY_DIR} PARENT_SCOPE)
 
+if(MSVC)
+	set(CMAKE_STATIC_LIBRARY_SUFFIX "$<$<CONFIG:Debug>:d>${CMAKE_STATIC_LIBRARY_SUFFIX}")
```
pivotal-jbarrett: Format

mmartell:
This is being replaced by PR #411.

PR #408, opened 2018-11-27
Geode-5962: Fix putAll crash with null values

mmartell:
This adds an error handler to TcrMessage to catch PUT_DATA_ERROR messages from the server. Also adds a test to ensure a proper exception is bubbled up to the user.

### DIFF
--- a/cppcache/src/TcrMessage.cpp
+++ b/cppcache/src/TcrMessage.cpp
@@ -862,7 +862,7 @@ void TcrMessage::processChunk(const uint8_t* bytes, int32_t len,
       if (bytes != nullptr) {
         _GEODE_SAFE_DELETE_ARRAY(bytes);
       }
-      // nothing else to done since this will be taken care of at higher level
+
       break;
     }
     default: {
--- a/cppcache/src/TcrMessage.cpp
+++ b/cppcache/src/TcrMessage.cpp
@@ -841,7 +841,7 @@ void TcrMessage::processChunk(const uint8_t* bytes, int32_t len,
       }
       break;
     }
-    case PUT_DATA_ERROR: {
+    case TcrMessage::PUT_DATA_ERROR: {
       chunkSecurityHeader(1, bytes, len, isLastChunkAndisSecurityHeader);
       if (nullptr != bytes) {
         auto input =
--- a/cppcache/src/TcrMessage.cpp
+++ b/cppcache/src/TcrMessage.cpp
@@ -54,6 +54,8 @@ namespace {
 uint32_t g_headerLen = 17;
 }  // namespace
 
+extern void setThreadLocalExceptionMessage(const char*);
+
 // AtomicInc TcrMessage::m_transactionId = 0;
 uint8_t* TcrMessage::m_keepalive = nullptr;
 const int TcrMessage::m_flag_empty = 0x01;
@@ -836,6 +838,22 @@ void TcrMessage::processChunk(const uint8_t* bytes, int32_t len,
       if (bytes != nullptr) {
         DeleteArray<const uint8_t> delChunk(bytes);
         LOGFINEST("processChunk - got response from secondary, ignoring.");
+      }
+      break;
+    }
+    case PUT_DATA_ERROR: {
+      chunkSecurityHeader(1, bytes, len, isLastChunkAndisSecurityHeader);
+      if (nullptr != bytes) {
+        auto input =
+            m_tcdm->getConnectionManager().getCacheImpl()->createDataInput(
+                bytes, len);
+        auto errorString = readStringPart(input);
+
+        if (!errorString.empty()) {
+          setThreadLocalExceptionMessage(errorString.c_str());
+        }
+
+        _GEODE_SAFE_DELETE_ARRAY(bytes);
       }
       break;
     }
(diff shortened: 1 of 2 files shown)


cppcache/integration-test-2/RegionPutGetAllTest.cpp
```diff
@@ -198,4 +199,38 @@ TEST(RegionPutGetAllTest, variousPdxTypes) {
...
+
+  // Add a null value
+  map.emplace(CacheableInt32::create(PdxTypesHelper1::index),
+              std::shared_ptr<PdxTypesHelper1::type>());
+
+  auto keys = to_keys(map);
+
+  try {
+    region->putAll(map);
+  } catch (CacheServerException) {
```
echobravopapa: should be able to use `ASSERT_THROW`, no?

cppcache/src/TcrMessage.cpp
```diff
@@ -838,6 +840,22 @@ void TcrMessage::processChunk(const uint8_t* bytes, int32_t len,
       }
       break;
     }
+    case PUT_DATA_ERROR: {
```
echobravopapa: `TcrMessage::` ?  for consistency if nothing else

cppcache/src/TcrMessage.cpp
```diff
@@ -838,6 +840,22 @@ void TcrMessage::processChunk(const uint8_t* bytes, int32_t len,
...
+
+        if (!errorString.empty()) {
+          setThreadLocalExceptionMessage(errorString.c_str());
+        }
+
+        _GEODE_SAFE_DELETE_ARRAY(bytes);
+      }
+      break;
+    }
     case TcrMessage::GET_ALL_DATA_ERROR: {
```
echobravopapa: These errors claim to be handled at a higher level, is the handling different than `PUT_DATA_ERROR`?

cppcache/integration-test-2/RegionPutGetAllTest.cpp
mmartell: Good catch. Replaced with EXPECT_THROW.

cppcache/src/TcrMessage.cpp
mmartell: Done.

cppcache/integration-test-2/RegionPutGetAllTest.cpp
pivotal-jbarrett: You should be using Expect exception.

pivotal-jbarrett: Use `ASSERT_EXCEPTION` https://github.com/abseil/googletest/blob/master/googletest/docs/advanced.md#exception-assertions

mreddington: Done.

cppcache/src/TcrMessage.cpp
mreddington: GET_ALL_DATA_ERROR is not a catchall for data errors, it is a specific error response to a get-all data request.

echobravopapa: My comment had to do with the claim that `GET_ALL_DATA_ERROR`s are handled at a higher level (line 864) and how does this compare to the route taken for `PUT_DATA_ERROR`? i.e. which is correct as they seem to be handled very differently...

cppcache/integration-test-2/RegionPutGetAllTest.cpp
mmartell: Done.

cppcache/src/TcrMessage.cpp
mmartell: If I understand you correctly, you're saying PUT_DATA_ERROR handler should eat these exceptions just like the other handlers. However, it seems like we should bubbling these exceptions up to the user. It looks like these aren't even logged in the client log, which seems to make it even more important to pass these back to the user.

cppcache/integration-test-2/RegionPutGetAllTest.cpp
```diff
@@ -221,16 +221,9 @@ TEST(RegionPutGetAllTest, nullValue) {
 
   auto keys = to_keys(map);
 
-  try {
-    region->putAll(map);
-  } catch (CacheServerException) {
-    localDestroy(*region, keys);
-    return;
-  }
+  EXPECT_THROW(region->putAll(map), CacheServerException);
```
echobravopapa: @pivotal-jbarrett preferred `ASSERT_EXCEPTION`

cppcache/integration-test-2/RegionPutGetAllTest.cpp
```diff
@@ -198,4 +199,31 @@ TEST(RegionPutGetAllTest, variousPdxTypes) {
...
+
+  // Add a null value
+  map.emplace(CacheableInt32::create(PdxTypesHelper1::index),
+              std::shared_ptr<PdxTypesHelper1::type>());
+
+  auto keys = to_keys(map);
+
+  EXPECT_THROW(region->putAll(map), CacheServerException);
+
+  localDestroy(*region, keys);
```
pivotal-jbarrett: Does this add value to this particular test?

mreddington: This has been removed.

PR #412, opened 2018-11-29
Geode-5769: remove app domain context

mreddington:


echobravopapa:
@mmartell in https://github.com/apache/geode-native/commit/5a46b659b86ecc4890df59c1b7abe727192e5d06 the AppDomainContext was moved to a lower level

pdxcodemonkey:
@mmartell the reference to AppDomainContext in Task can't be removed.  As Ernie pointed out, it was, in fact, added there a short while ago in the PR that spawned GEODE-5769.  The changes to Task, however, do imply that you no longer need the AppDomainContext code in TcrChunkedResponse, which is what this PR is all about.

PR #419, opened 2018-12-11
GEODE-4841: Changed typedef to using statement.

mreddington:


echobravopapa:
Given this is involved in a PIMPL model association, I'm not convinced this gains much... curious to hear thoughts...

moleske:
@echobravopapa I see it more as a "C++11 all the things."  Start setting some precedence to prefer `using` over `typedef` since you can do fun things with templates and scope better with `using`

echobravopapa:
@moleske for sure, wasn't sure the value beyond that and hoped to spur some discussion...

mreddington:
The ugly of this is the typedef is defined twice, in the same namespace, across two headers, which coincidentally are both included in the same source. May the two never differ.

The right thing to do would be to put the definition in its own header and include it where needed to remove the duplication.

PR #410, opened 2018-11-27
Geode-5825: Added unit tests to increase coverage of TcrMessage

mmartell:

PR #423, opened 2018-12-17
Geode 5825: Increase unit test code coverage of TcrMessage.

mreddington:
These tests are in response to current code coverage results, systematically going down the file and covering all paths.


mmartell:
Please use the proper ticket title format: GEODE-NNNN: Pull request headline string

cppcache/integration-test-2/TcrMessageTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+TEST(TcrMessage, readExceptionPart) {
+    auto cache = CacheFactory().create();
+    TcrMessageDummy message;
+    auto input = cache.createDataInput(nullptr, 0);
+
+    EXPECT_FALSE(message.readExceptionPart(input, 0, false));
+}
+
+TEST(TcrMessage, DISABLED_writeBytesOnly) {
```
mmartell: Please add a comment about why this test is disabled. Otherwise remove it.

pivotal-jbarrett: A disabled test is a FAILED test!

cppcache/integration-test-2/CMakeLists.txt
```diff
@@ -46,7 +47,7 @@ target_include_directories(integration-test-2
 
 target_link_libraries(integration-test-2
   PUBLIC
-    apache-geode
+    apache-geode-static
```
pivotal-jbarrett: Integration tests MUST use the dynamic library.

mreddington: Then we need separate unit tests.

cppcache/integration-test-2/TcrMessageTest.cpp
mreddington: There is what appears to be incorrect behavior when writing the cacheable string to the message. At some point after the object is written, the write pointer is moved back an incorrect number of bytes. The buffer contains a 1 byte DSCode and then a 2 byte size field; the write pointer is moved back to the size field, where the first byte of this 2 byte field is subsequently read as the DSCode. It warrants further investigation.

cppcache/integration-test-2/CMakeLists.txt
pivotal-jbarrett: Integration tests should not test internal classes.

cppcache/integration-test-2/TcrMessageTest.cpp
pivotal-jbarrett: Do we not have any old integration tests that pass through this same code? See coverage reports.

mreddington: This test exists because the coverage reports indicate this code path was not exercised.

cppcache/integration-test-2/CMakeLists.txt
mreddington: I'm not testing an internal class, I'm dependent upon it. Efforts seem focused on migrating to this test suite and indications point toward abandoning the others, and there is misinformation: it was suggested to me to link to the static library, so not everyone is clear on the nature of this test suite.

We definitely need unit tests separate from integration tests. We should test the shared object as well as the static object and our internal classes. This much we agree. I'll discuss this with the team because it's clear to me not everyone is on the same page.

mmartell:
It looks like you forgot to close PR #410.

PR #430, opened 2019-01-04
GEODE-6210: Add 'transaction' example for cpp

moleske:
- demonstrates commit and rollback of transactions
 - handles exceptions

### DIFF
--- a/examples/cpp/transaction/main.cpp
+++ b/examples/cpp/transaction/main.cpp
@@ -23,91 +23,68 @@
 #include <geode/RegionFactory.hpp>
 #include <geode/RegionShortcut.hpp>
 
-using namespace apache::geode::client;
+using apache::geode::client::Cache;
+using apache::geode::client::CacheFactory;
+using apache::geode::client::CacheTransactionManager;
+using apache::geode::client::RegionShortcut;
+
+auto keys = {
+    "Key1",
+    "Key2",
+    "Key3",
+    "Key4",
+    "Key5",
+    "Key6",
+    "Key7",
+    "Key8",
+    "Key9",
+    "Key10"
+};
+
+void initExternalSystem() {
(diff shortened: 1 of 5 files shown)


pivotal-jbarrett:
Something like:
```c++
main() {
  ...
  transactionManager.begin();
  try {
    for (auto& key : batch) {
      auto value = getValueFromExternalSystem(key);
      cache.put(key, value);
    }
    transactionManager.commit();
  } catch ( ... ) {
    transactionManager.rollback();
  }
}

int32_t getValuFromExternalSystem(int32_t key) {
  if (random(10) == 0) {
    throw "failed to get from external system";
  }

  return random(key);
}
```

pdxcodemonkey:
I was getting stuck on the idea that we needed to force an exception external to the example code to have a "real" example.  Charlie has suggested we implement a retry path as well, which also seems like a good suggestion.  Thoughts?

pdxcodemonkey:
Le sigh - is there a modern C++ version that's not way more complex?  STL stuff seems like overkill

examples/cpp/transaction/main.cpp
```diff
@@ -0,0 +1,91 @@
...
+    "Key6",
+    "Key7",
+    "Key8",
+    "Key9",
+    "Key10"
+};
+
+
+int getValueFromExternalSystem() {
+  std::random_device rd{};
```
pivotal-jbarrett: This compiles? So we have non-static `rd` passed to a static `e`. At best here we allocating more `rd` on each call that we will never use.

pivotal-jbarrett:
```C++
static thread_local std::default_random_engine generator(std::random_device{}());
auto value = std::uniform_int_distribution<int32_t>{0, 9}(generator);
...
```

PR #431, opened 2019-01-07
GEODE-6249: port attributes mutator test

pdxcodemonkey:
... and rename it "ExpirationTest," since it doesn't actually use/test AttributeMutator.


pdxcodemonkey:
Yeah we'll need to check and see if we have coverage of attributes elsewhere,  cause this test definitely did/does nothing the original name would suggest.

PR #443, opened 2019-02-19
GEODE-4348: Dotnet SSL Example

mmartell:


examples/dotnet/SslPutGet/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/CMakeLists.txt.in ${CMAKE_CURRENT_BINARY_DIR}/CMakeLists.txt @ONLY)
+
+install(FILES
+  ${CMAKE_CURRENT_BINARY_DIR}/CMakeLists.txt
+  Program.cs
+  README.md
+  startserver.ps1
+  stopserver.ps1
+  DESTINATION examples/dotnet/SslPutGet)
```
echobravopapa: what? no whitespace here...

echobravopapa: you have a `.in` file, should this file be in repo? as well

mreddington: The ssl example has additional requirements the standard example template doesn't cover, so it has its own. That's why the root cmake file uses add_directory instead of add_example, the cmake file in the ssl example directory populates the template in the example install directory. It has to be done this way because there are variables available when installing examples that are not available when building examples.

igodwin:
Changes required from recently merged PR demands rework. Closing until completed.

PR #449, opened 2019-02-27
GEODE-4339: cpp ssl example

mmartell:


cppcache/src/TcrConnection.cpp
```diff
@@ -568,9 +568,9 @@ Connector* TcrConnection::createConnection(
                                .getSystemProperties();
   if (systemProperties.sslEnabled()) {
     socket = new TcpSslConn(endpoint, connectTimeout, maxBuffSizePool,
-                            systemProperties.sslKeystorePassword().c_str(),
                             systemProperties.sslTrustStore().c_str(),
-                            systemProperties.sslKeyStore().c_str());
+                            systemProperties.sslKeyStore().c_str(),
+                            systemProperties.sslKeystorePassword().c_str());
```
pdxcodemonkey: Whitespace issue here?  This diff looks weird...

examples/cpp/sslputget/main.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+using apache::geode::client::CacheableString;
+using apache::geode::client::CacheFactory;
+using apache::geode::client::RegionShortcut;
+
+std::string myGetcwd() {
+  char buf[PATH_MAX];
+  std::string cwd;
+
+#ifdef _MSC_VER
```
pdxcodemonkey: If we changed this to something like:
```
#ifdef _MSC_VER
  char buf[MAX_PATH];
  if (_getcwd(buf, MAX_PATH)) {
    cwd = buf;
  }
#else
  char buf[PATH_MAX];
  if (getcwd(buf, PATH_MAX)) {
    cwd = buf;
  }
#endif
```
could we get rid of all the awful _MSC_VER stuff above?

cppcache/src/TcrConnection.cpp
mmartell: Hmm. Not seeing the issue here.

examples/cpp/sslputget/main.cpp
mmartell: Agreed we can make this a little cleaner. (Ultimately we'll go to std::filesystem)

PR #453, opened 2019-03-01
GEODE-4339: Parameterize target name in copy step for C++ SSL example

igodwin:


pdxcodemonkey:
Closed, because fix in this branch has been subsumed by larger fix in PR #456.

PR #463, opened 2019-03-19
GEODE-6538: Bump native client user guide version to 1.9.

davebarnes97:
Also: Update for compatibility with bookbindery 10.1.15.


docs/geode-native-book/Gemfile.lock
```diff
@@ -101,7 +101,7 @@ GEM
     i18n (0.7.0)
     ipaddress (0.8.3)
     kramdown (1.17.0)
-    libv8 (3.16.14.7)
+    libv8 (3.16.14.19-x86_64-darwin-18)
```
igodwin: Should this really include the reference to Darwin? It would make sense if to act as reference to the OS that was used to build the docs, but I am lacking context.

davebarnes97: @igodwin  If a user downloads the release and follows directions to generate the user guide, the Gemfile.lock file will be regenerated from the Gemfile (which specifies libv8 with no particular version) for the user's environment. It's possible that it doesn't need to be included here at all - I'm just following precedent, not sure what constitutes 'best practice'. 
Would you recommend that I (a) delete the line, (b) delete the file, or (c) just leave it as is? I'll take your advice.

igodwin: If the .lock file is generated during the process, and there is no need for one to already be in place, I would delete it.

igodwin:
Please update docs/docker/Dockerfile and replace the obsolete arguments, `-no-ri` and `-no-rdoc`, on line 28, with `-no-document`.

PR #464, opened 2019-03-24
GEODE-5708: fix memory corruption.

vaijira:
Use shared pointer directly instead of create a shared pointer from a raw pointer.

### DIFF
--- a/cppcache/src/VersionedCacheableObjectPartList.hpp
+++ b/cppcache/src/VersionedCacheableObjectPartList.hpp
@@ -117,7 +117,7 @@ class VersionedCacheableObjectPartList : public CacheableObjectPartList {
   }
 
   VersionedCacheableObjectPartList(
-      std::vector<std::shared_ptr<CacheableKey>>* keys,
+      std::shared_ptr<std::vector<std::shared_ptr<CacheableKey>>> keys,
       std::recursive_mutex& responseLock)
       : m_tempKeys(keys), m_responseLock(responseLock) {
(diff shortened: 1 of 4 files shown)


cppcache/src/VersionedCacheableObjectPartList.hpp
```diff
@@ -117,7 +117,7 @@ class VersionedCacheableObjectPartList : public CacheableObjectPartList {
   }
 
   VersionedCacheableObjectPartList(
-      std::vector<std::shared_ptr<CacheableKey>>* keys,
+      std::shared_ptr<std::vector<std::shared_ptr<CacheableKey>>> keys,
```
pivotal-jbarrett: If the vector of keys is completely owned by the VersionedCacheableObjectPartList it would be preferable to just make the `std::vector` and member value or at most a `std::unique_ptr`.

vaijira: The member keys are a shared_ptr because sometimes is needed this way (like in ThinkClientRegion call), other times not (like in PutAllPartialResult call). How do you propose that the keys member are sometimes unique and other times shared? upgrading a unique_ptr to shared when necessary?

pivotal-jbarrett: Oh, I see what's going on now. Yes, you're are correct, a `shared_ptr` is warranted here. I didn't see that the ownership of the vector was external to the `VersionedCacheableObjectPartList` initially. Thanks!

PR #465, opened 2019-03-27
GEODE-6568: Use OpenSSL from local install

pivotal-jbarrett:
Support OpenSSL 1.0.1, 1.0.2, 1.1.0, and 1.1.1.


dependencies/CMakeLists.txt
```diff
@@ -19,7 +19,6 @@ project( dependencies LANGUAGES NONE )
 find_package(Patch REQUIRED)
 
 add_subdirectory(	libxml2 )
```
pdxcodemonkey: So we're no longer building OpenSSL in the tree, even if it's not installed on the system?

pdxcodemonkey: We'll need to update BUILDING.md to account for installation of OpenSSL, too...

pivotal-jbarrett: No, mixing and matching is really hard. Some things in CMake are easy to do in either the configure or build stages but not in both. Yes, need to update the BUILDING.md, good catch!

pdxcodemonkey:
This is filed in Jira as GEODE-6568

PR #466, opened 2019-03-27
GEODE-6567: Fix GCC 8 warnings.

vaijira:
Fix GCC 8 warnings.

### DIFF
--- a/cppcache/integration-test/ThinClientPutAll.hpp
+++ b/cppcache/integration-test/ThinClientPutAll.hpp
@@ -401,7 +401,7 @@ DUNIT_TASK_DEFINITION(CLIENT1, StepThree)
           PdxTests::Address::createDeserializable);
       serializationRegistry->addPdxSerializableType(
           PdxTests::PdxTypes3::createDeserializable);
-    } catch (Exception &e) {
+    } catch (Exception&) {
       LOG("Got expected Exception for Serialization, already registered");
     }
     // the client1 will register k0 and k1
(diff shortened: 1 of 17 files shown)


cppcache/integration-test/ThinClientTransactions.hpp
```diff
@@ -632,7 +632,7 @@ DUNIT_TASK_DEFINITION(CLIENT1, SuspendResumeCommit)
     bool resumeExc = false;
     try {
       txManager->resume(m_suspendedTransaction);
-    } catch (const IllegalStateException) {
+    } catch (const IllegalStateException &e) {
```
pdxcodemonkey: Haven't verified yet, but I'm fairly certain naming the variable here will generate a new warning on Clang.  I believe adding the '&' but leaving it anonymous works.

cppcache/integration-test/ThinClientPutAll.hpp
```diff
@@ -401,7 +401,7 @@ DUNIT_TASK_DEFINITION(CLIENT1, StepThree)
           PdxTests::Address::createDeserializable);
       serializationRegistry->addPdxSerializableType(
           PdxTests::PdxTypes3::createDeserializable);
-    } catch (Exception e) {
+    } catch (Exception &e) {
```
pivotal-jbarrett: Style guide says justify the `&` consistent with the rest of the file. In this case it should be next to the type and not the variable.

cppcache/integration-test/ThinClientTransactions.hpp
pdxcodemonkey: Can confirm, this breaks our clang build with an error "unused exception parameter 'e'".  Please leave these anonymous wherever they're not used.

PR #472, opened 2019-04-09
GEODE-6624: Fix nested exception in geode-native causing SIGABRT

pdxcodemonkey:

We were hitting a nested exception, which caused a SIGABRT in the client.  Now we catch the second exception, and clear the first out of the "pending" location, so it doesn't get thrown from within a catch block and blow us up.


pdxcodemonkey:
Actually, gonna close this for a while until I can add a repro test case.

PR #473, opened 2019-04-11
Add an instruction to pass OPENSSL_ROOT_DIR

sboorlagadda:


BUILDING.md
```diff
@@ -22,7 +22,9 @@ Building requires access to an installation of Geode. There are two ways to achi
 * Pass in `GEODE_ROOT` during the CMake configuration step.
   * e.g.  add `-DGEODE_ROOT=/path/to/geode` to the _initial_ `cmake` execution command.
 
-
+## Setting Path to OpenSSL
```
pivotal-jbarrett: This should only be necessary if not using a an OpenSSL installed in a well known location or provided by the operating system.

sboorlagadda: Thanks. Updated accordingly.

BUILDING.md
```diff
@@ -37,6 +33,7 @@ $ cmake .. <platform-specific generator parameters (see below)>
 # build step
 $ cmake --build . -- <platform-specific parallelism parameters (see below)>
 ```
+If OpenSSL is installed in a custom location, then one can pass `OPENSSL_ROOT_DIR` during the CMake configuration step. e.g.  add `-DOPENSSL_ROOT_DIR=/path/to/openssl`.
```
pdxcodemonkey: I think we need sort of a combination of these two comments.  This one doesn't mention that it's _required_, and 'can pass' should probably be replaced by 'must pass', because the build will fail without it.

sboorlagadda: I haven't mentioned this as required at this step because it is one of the dependency that one has to install and is part of prerequisites section and the reason for `can pass` is because it is only required to pass if you have not installed in the standard location.

pdxcodemonkey: Well exactly - the sentence begins 'If OpenSSL is installed in a custom location', which means it's required.

sboorlagadda: Agreed. Updated to `must pass`.

PR #474, opened 2019-04-11
GEODE-6635: Native Client API docs: Use a 3-part version header

davebarnes97:
Sets the header on API doc pages to the 3-part version number, based on the PRODUCT_VERSION specified at build time. If PRODUCT_VERSION has less than 3 components, adds "0" elements to fill out the string.
Old behavior was a 2-part version number, or an empty string if only one component was specified.
Updated BUILDING.md to clarify the option of defining PRODUCT_VERSION on the cmake config command line, and to fix a typo.


igodwin:
Verified building on supported platforms, using CI.

PR #477, opened 2019-04-15
GEODE-6576: Improvements in cleanStaleConnections method

mivanac:
Improvements in cleanStaleConnections method:
- introduce phasing
- return connections to pool as soon as possible

### DIFF
--- a/cppcache/src/ThinClientPoolDM.cpp
+++ b/cppcache/src/ThinClientPoolDM.cpp
@@ -18,6 +18,7 @@
 #include "ThinClientPoolDM.hpp"
 
 #include <algorithm>
+#include <thread>
 
 #include <ace/INET_Addr.h>
 
@@ -430,6 +431,14 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
     size_t replaceCount =
         m_attrs->getMinConnections() - static_cast<int>(savelist.size());
 
+    LOGDEBUG("Preserving %d connections", savelist.size());
+
+    for (auto savedconn : savelist) {
+      put(savedconn, false);
+    }
+    savelist.clear();
+    int count = 0;
+
     for (std::vector<TcrConnection*>::const_iterator iter = replacelist.begin();
          iter != replacelist.end(); ++iter) {
       TcrConnection* conn = *iter;
@@ -451,7 +460,7 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
           if (nextIdle > std::chrono::seconds::zero() && nextIdle < _nextIdle) {
             _nextIdle = nextIdle;
           }
-          savelist.push_back(newConn);
+          put(newConn, false);
           if (newConn != conn) {
             GF_SAFE_DELETE_CON(conn);
             removeEPConnections(1, false);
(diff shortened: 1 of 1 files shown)


cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -430,6 +431,15 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
     size_t replaceCount =
         m_attrs->getMinConnections() - static_cast<int>(savelist.size());
 
+    LOGDEBUG("Preserving %d connections", savelist.size());
+
+    for (std::vector<TcrConnection*>::const_iterator iter = savelist.begin();
```
pdxcodemonkey: Please use a range-based for loop here

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -474,20 +484,19 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
...
             }
-            savelist.push_back(conn);
+            put(conn, false);
           }
         }
       }
       replaceCount--;
+      count++;
+      if (count % 40 == 0) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
pdxcodemonkey: This looks like debug code(?!), please remove

mivanac: This is  part of solution for this issue. Main problem in this issue is that this thread takes over processor for long period (if we have many connections), and due to that we have latency in request processing. So idea is every 40 cycles to give chance to other threads(this is what we tested). Maybe we can try with sleep 1ms every 10 cycles, or use of yield.

echobravopapa: a couple things:  has a test been created to drive out this issue so it can be seen to be fixed; second, it would be nice to find an alternative to sleeping the thread

mivanac:
As I described in ticket, this issue is reported with real environment, and with simulation we did not get realistic data. So this is why I think, that it is difficult to write TC for this issue. Steps to reproduce on real environment are:
-increase number of min and max connections to 400
-run get operation, and meassure time response
At LoadConditioningInterval you will see increase in time reponse.

With yield we did not get good results (same as without any solution), but with sleep of 1ms every 10 cycles, we got even better.

mivanac:
This fault can be reproduced with geode-native-examples - putgetremove. For reproduction test:
- set min and max connections to 800.
- set LoadConditioningInterval to 60 secs
- run continues get operation, and measure duration of get operation.
  Every 60 seconds you will see peaks with time duration over 100ms.
And this is the problem.

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -474,20 +483,19 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
                 nextIdle < _nextIdle) {
               _nextIdle = nextIdle;
             }
-            savelist.push_back(conn);
+            put(conn, false);
           }
         }
       }
       replaceCount--;
+      count++;
```
pivotal-jbarrett: I'd be more comfortable with `std::this_thread::yield` every time through this loop. Let the scheduler decide if and when there are free cycles to run this task hot or not. Remove the counter and mod condition.

mivanac: Testing of solution alternatives is ongoing. Ticket will be updated with test results.

mivanac: After testing of yield solution, it is observed that delays are same as without solution, which is not acceptable. Delays are in range of 1 - 2 seconds per get at some peeks.
When tested solution with "sleep", max delays are cca. 150 ms.

pivotal-jbarrett: Odd... Ok.

PR #479, opened 2019-04-29
GEODE-6718: Refactor TcrConnection::sendRequestForChunkedResponse, derived class, and remove pass-through calls in TrcConnection and derived

pdxcodemonkey:
The method in TcrConnection greatly needed clarification/simplification, and the corresponding methods in Tcr*Endpoint literally do nothing.  This is a nice small thing that can be done to make things more readable and maintainable.

### DIFF
--- a/cppcache/src/TcrConnection.cpp
+++ b/cppcache/src/TcrConnection.cpp
@@ -714,12 +714,12 @@ void TcrConnection::sendRequestForChunkedResponse(
 bool TcrConnection::replyHasValidTimeout(const TcrMessage& request) const {
   auto messageType = request.getMessageType();
   return ((messageType == TcrMessage::QUERY) ||
-      (messageType == TcrMessage::QUERY_WITH_PARAMETERS) ||
-      (messageType == TcrMessage::EXECUTECQ_WITH_IR_MSG_TYPE) ||
-      (messageType == TcrMessage::GETDURABLECQS_MSG_TYPE) ||
-      (messageType == TcrMessage::EXECUTE_FUNCTION) ||
-      (messageType == TcrMessage::EXECUTE_REGION_FUNCTION) ||
-      (messageType == TcrMessage::EXECUTE_REGION_FUNCTION_SINGLE_HOP));
+          (messageType == TcrMessage::QUERY_WITH_PARAMETERS) ||
+          (messageType == TcrMessage::EXECUTECQ_WITH_IR_MSG_TYPE) ||
+          (messageType == TcrMessage::GETDURABLECQS_MSG_TYPE) ||
+          (messageType == TcrMessage::EXECUTE_FUNCTION) ||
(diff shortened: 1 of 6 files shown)


cppcache/src/TcrConnection.cpp
```diff
@@ -691,51 +691,70 @@ char* TcrConnection::sendRequest(const char* buffer, size_t len,
...
-      break;
-    }
-    default:
-      break;
-  }
-
+std::chrono::microseconds TcrConnection::sendWithTimeouts(
+    const char* data, size_t len, std::chrono::microseconds sendTimeout,
+    std::chrono::microseconds receiveTimeout) {
   std::chrono::microseconds timeSpent{0};
```
echobravopapa: the diff is a bit dodgy to look at... the new method `sendWithTimeouts` is an extraction of the next several lines, no?

cppcache/src/TcrConnection.cpp
```diff
@@ -691,51 +691,70 @@ char* TcrConnection::sendRequest(const char* buffer, size_t len,
...
-      return;
+      hasResult = false;
     }
   }
-  readMessageChunked(reply, receiveTimeoutSec, true);
+
+  return hasResult;
+}
+
+void TcrConnection::sendRequestForChunkedResponse(
```
echobravopapa: complete and total nitpick... maybe next time put the all the extracted methods below the parent method, just to end with a more digestible Diff... i.e. I think the changes would at least difference from block of code to new method call etc... e.g. the `switch(msgType)` block would diff to the call of `replayHasValidTimeout()`

pdxcodemonkey: ima try moving that and see how it affects the diff.  Thanks for this feedback, it's always a tough call to know exactly where to put new methods.

cppcache/src/TcrConnection.cpp
pdxcodemonkey: Yeah, all the new methods are just extractions of blocks of code

cppcache/src/TcrConnection.cpp
```diff
@@ -695,47 +695,66 @@ void TcrConnection::sendRequestForChunkedResponse(
...
-    case TcrMessage::EXECUTE_FUNCTION:
-    case TcrMessage::EXECUTE_REGION_FUNCTION:
-    case TcrMessage::EXECUTE_REGION_FUNCTION_SINGLE_HOP: {
-      receiveTimeoutSec = reply.getTimeout();
-      sendTimeoutSec = reply.getTimeout();
-      break;
-    }
-    default:
-      break;
+  if (replyHasValidTimeout(request)) {
```
pivotal-jbarrett: This method name confuses me? Is there an invalid timeout otherwise? Consider `useReplyTimeout`?

pdxcodemonkey: Done.  Good suggestion, thanks!

PR #482, opened 2019-05-01
Remove TestConfig.cpp since it is empty

moleske:
Rat is failing on this file and there's a `TestConfig.cpp.in` so seems this can just go away.


pdxcodemonkey:
Thanks - no idea how this snuck into our fix for GEODE-6624, but there it is.  TestConfig.cpp.in is exactly the correct file, and it produces TestConfig.cpp in the build directory where it's used.

PR #483, opened 2019-05-03
GEODE-3415: Add apache-geode dependency to integration-test2

mmartell:
With the new SslException class, cryptoimpl now depends on apache-geode.dll. This commit adds the copy to the post build step of cli's integration-test2.

### DIFF
--- a/clicache/integration-test2/CMakeLists.txt
+++ b/clicache/integration-test2/CMakeLists.txt
@@ -104,4 +104,8 @@ add_custom_command(TARGET ${PROJECT_NAME} POST_BUILD
     $<SHELL_PATH:$<TARGET_FILE:cryptoImpl>>
     $<$<CONFIG:Debug>:$<SHELL_PATH:$<TARGET_PDB_FILE:cryptoImpl>>>
     $<SHELL_PATH:$<TARGET_FILE_DIR:${PROJECT_NAME}>>
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    $<SHELL_PATH:$<TARGET_FILE:apache-geode>>
+    $<$<CONFIG:Debug>:$<SHELL_PATH:$<TARGET_PDB_FILE:apache-geode>>>
+    $<SHELL_PATH:$<TARGET_FILE_DIR:${PROJECT_NAME}>>
 )


pdxcodemonkey:
Wait, isn't this a circular dependency?  I thought when @pivotal-jbarrett asked for a more specific exception class, we'd be adding code for one to cryptoimpl, not importing one from the native client.  This seems bad to me at first blush, but I'll wait for others to comment....

pdxcodemonkey:
@echobravopapa This is the first use a an NC exception in *cryptoimpl*, which makes cryptoimpl dependent on apache-geode.dll.  Which we have concluded is probably bad.

mmartell:
Agreed. I was discussing in our slack channel.

I think we should just rollback to throw std::exception as in the initial
commit. Alternatively, we could still use an SslException class that
inherits std::exception instead of apache.geode.excpetion.

pivotal-jbarrett:
The circular dependency has always been there but it looks the the VC++ linker noticed that no symbols were imported and left the dll out of the implicit load set. While I prefer to solve the circular linker problem for this plugin so it can use meaningful exceptions, I am fine with just throwing a `std::exception` like `std::invalid_argument`. This exception should be caught inside the library and translated into a library `Exception`, probably `SslException` still makes sense in this case.

It also turns out that the pre-existing circular dependency is artificial. The real dependency was removed when we replaced `SharedPtr` with `std::shared_ptr`. Remove the unused header include in `DHImpl.cpp`, `#include <geode/internal/geode_globals.hpp>`. Then you can remove the dependency from the CMakeLists.txt, `apache-geode`.

pdxcodemonkey:
Okay I’ll ask the vicious question here:  why do we have any exception classes implemented in a .cpp file at all?  They should be trivial classes, the implementation should all be in a header, and using them shouldn’t create a library dependency at all.  Is there a story there?

pivotal-jbarrett:
Because we had this silly idea that exceptions should capture a stack trace. Booo!

The stack trace code needs to be in the cpp to hide the fact it uses boost from the public api. It used to use all sorts of nasties in the past. If we got rid of stack tracing then yes it could be header only.

PR #484, opened 2019-05-06
Moved header implemenations into the source file where they belong.

mreddington:


cppcache/src/TcrMessage.cpp
```diff
@@ -388,17 +707,6 @@ inline void TcrMessage::readKeyPart(DataInput& input) {
   }
 }
 
-inline void TcrMessage::writeInt(uint8_t* buffer, uint16_t value) {
```
pivotal-jbarrett: Now the compiler can’t inline these frequently executed methods.

mreddington: Yes it can. An optimizing compiler is free to inline this function given the entire definition is visible and constrained to this source file.

pivotal-jbarrett: Ah, didn't see that it was yanked from private.

cppcache/src/TcrMessage.cpp
```diff
@@ -225,7 +544,7 @@ void TcrMessage::readCqsPart(DataInput& input) {
   // LOGDEBUG("mapsize = %d", m_cqs.size());
 }
 
-inline void TcrMessage::readCallbackObjectPart(DataInput& input,
```
pivotal-jbarrett: Why remove this inlining here?

cppcache/src/TcrMessage.cpp
```diff
@@ -2960,6 +3254,236 @@ void TcrMessage::readHashSetForGCVersions(
   }
 }
 
+
+TcrMessageDestroyRegion::~TcrMessageDestroyRegion() {}
```
pivotal-jbarrett: Any not `~X() noexcept = default`?
Why not in the header?

cppcache/src/TcrMessage.hpp
```diff
@@ -675,9 +493,7 @@ class TcrMessageClearRegion : public TcrMessage {
                         std::chrono::milliseconds messageResponsetimeout,
                         ThinClientBaseDM* connectionDM);
 
-  virtual ~TcrMessageClearRegion() {}
-
- private:
+  ~TcrMessageClearRegion() override;
```
pivotal-jbarrett: Why not `~TcrMessageClearRegion() noexcept override = default;`?

At least do `noexcept` on these cleanups.

cppcache/src/TcrMessage.hpp
```diff
@@ -1178,194 +910,73 @@ class TcrMessageExecuteFunction : public TcrMessage {
...
-            methodName);
-        throw MessageException(exMsg);
-      }
-    } else if (partType == DSCode::NullObj) {
-      // special null object is case for scalar query result
-      return ChunkObjectType::NULL_OBJECT;
-    }
-    return ChunkObjectType::OBJECT;
-  }
+  TcrMessageHelper() = delete;
```
pivotal-jbarrett: Deleted constructors and other special functions should always be `public`.

cppcache/src/TcrMessage.cpp
mreddington: As per the spec:

"function specifiers, only allowed in function declarations"

This is not a function declaration.

cppcache/src/TcrMessage.cpp
mreddington: We are moving all implementation out of the headers.

cppcache/src/TcrMessage.hpp
mreddington: I was trying not to get too fancy about changing implementation details - the idea was really just to move stuff without getting carried away. We shouldn't have to do ```noexcept```, as that's implicit since C++11, which is our target spec. I do know default doesn't always generate the right thing, I need to read up on the subject before I commit to changing the code accordingly. Technically I screwed up with override, which I should remove.

cppcache/src/TcrMessage.hpp
mreddington: Sure. Is there a reason for that?

dgkimura: Assumption is you'll generally get better error messages, specifically in this case GCC version < 8.

cppcache/src/TcrMessage.hpp
mreddington: C++11 noexcept is implicitly true. I defaulted the derived destructors.

PR #485, opened 2019-05-07
Geode-3415: Use SslException with fixed circular dependency

mmartell:


mmartell:
Just noticed I left some test code in.

cryptoimpl/SSLImpl.cpp
```diff
@@ -21,9 +21,9 @@
 
 #include <ace/Guard_T.h>
 
-#include <geode/ExceptionTypes.hpp>
+//#include <geode/ExceptionTypes.hpp>
```
pivotal-jbarrett: Delete not comment.

mmartell: Good catch. Done.

mmartell:
For context, refer to PR #483: https://github.com/apache/geode-native/pull/483

pivotal-jbarrett:
Never! No! 

`std::exception` isn’t meant to be thrown directly. It lacks a constructors for `what`.

Throw `std::invalid_argument`.

mmartell:
std::invalid_argument seems totally appropriate for passing an invalid password. What about failure to read a certificate? Would std::runtime_error be more appropriate? It allows sets the what member to the string.

echobravopapa:
Is it failure to "read" the certificate; i.e. there was no certificate?  or this is an invalid certificate, which aligns to an invalid_argument...

pivotal-jbarrett:
All failures are a result of the arguments passed in. So `std::invalid_argument` is close enough.

PR #486, opened 2019-05-15
GEODE-6641: Code impacts for support for two phase commit transactions

mivanac:
Coding impacts, to add support of 2 phase commit to public interface.


mivanac:
So, should I add some tests to verify there is no problems, or return old interface, and make it public?

pivotal-jbarrett:
Yes you should make sure there are tests for this. I don’t think there is much in the way of tests around either of these two transaction types.

cppcache/integration-test/ThinClientTransactions.hpp
```diff
@@ -1002,6 +1005,98 @@ DUNIT_TASK_DEFINITION(CLIENT2, UpdateClient2Entries)
     LOG("StepSix complete.");
   }
 END_TASK_DEFINITION
+DUNIT_TASK_DEFINITION(CLIENT1, CreateClient1additionalEntries)
```
pivotal-jbarrett: This test should be written in the new test framework.

cppcache/integration-test/ThinClientTransactionsNewXA.hpp
```diff
@@ -0,0 +1,659 @@
+/*
```
pivotal-jbarrett: Maybe you misunderstood my previous ask. This integration test should not be written in the old integration test framework. Please put in the new framework. See cppcache/integration/test for examples.

mivanac: Updated

PR #487, opened 2019-05-23
GEODE-3287: remove Assert.hpp

pdxcodemonkey:
Also remove Assert.cpp, replace any instance of Assert that was actually used with assert() from <cassert> in std.


cppcache/src/TcrHADistributionManager.cpp
```diff
@@ -36,7 +38,7 @@ TcrHADistributionManager::TcrHADistributionManager(
     ThinClientRegion* theRegion, TcrConnectionManager& connManager)
     : ThinClientDistributionManager(connManager, theRegion),
       m_theTcrConnManager(connManager) {
-  GF_R_ASSERT(theRegion != nullptr);
+  assert(theRegion != nullptr);
```
pivotal-jbarrett: Why delete everywhere else but here? There were a few others that you converted to `assert`, why?

pdxcodemonkey: GF_D_ASSERT, as defined, was always *off* by default, so I removed it.  GF_R_ASSERT was always *on*, so I replaced those with asserts to minimize impact on actual behavior of the code

pivotal-jbarrett: Make sense. As much as I don't like `asserts` in productions code, that is outside the scope of this issue. Good call!

pdxcodemonkey: I kinda expect we can safely get rid of the 3 instances of it altogether, but yeah we should have a quick chat about that later

PR #488, opened 2019-05-23
GEODE-6800: CacheableFileName Linker Error

igodwin:
Having the ctors inline is throwing some versions of gcc and clang through a loop. Not having them in the vtable is tolerated by some, but causing linker errors on others.


igodwin:
This does force a recompile.

PR #489, opened 2019-05-29
GEODE-6623: Native Client user guide - document the client-cache schema

davebarnes97:


docs/geode-native-book/redirects.rb
```diff
@@ -15,6 +15,8 @@
 
 # Links to API Documentation #
 r301 %r{/releases/latest/javadoc/(.*)}, 'https://geode.apache.org/releases/latest/javadoc/$1'
+r302 %r{/cppdocs/(.*)}, 'https://gemfire-nc-100-cppdocs.docs.pivotal.io/$1'
```
moleske: Is this the right url?  Was expecting something on `geode.org`

davebarnes97: Good catch. Thanks!

PR #494, opened 2019-06-06
GEODE-6844: Clear headers of method bodies

mreddington:
* No longer including LRUList.cpp in other source files
* All function and method bodies are moved into source files
* All function and method declarations have had "inline" removed
* All derived destructors are declared "override"
* Many methods declared "override" have had their redundant "virtual" declarations removed
* All destructors have had "noexcept" removed (explicit "noexcept" is redundant)
* All redundant "final" declarations are removed if the derived class is final
* No constructor is declared "explicit" (it serves to prevent universal initialization since C++11)
* Source files added as necessary
* Innumerable linker and compiler errors fixed (it's a wonder how some of this ever compiled)
* Removed some redundant use of "this->"
* Several classes aren't derived, so their virtuality has been removed

Compilation times are down by >25%.


.clang-tidy
```diff
@@ -1,5 +1,5 @@
 ---
-Checks:          '-*,clang-diagnostic-*,clang-analyzer-*,-clang-analyzer-alpha*,google-*,-google-readability-todo,-google-runtime-references,-google-default-arguments,-clang-analyzer-core.uninitialized.UndefReturn,-clang-analyzer-core.UndefinedBinaryOperatorResult,-clang-analyzer-optin.cplusplus.VirtualCall'
+Checks:          '-*,clang-diagnostic-*,clang-analyzer-*,-clang-analyzer-alpha*,google-*,-google-readability-todo,-google-runtime-references,-google-default-arguments,-clang-analyzer-core.uninitialized.UndefReturn,-clang-analyzer-core.UndefinedBinaryOperatorResult,-clang-analyzer-optin.cplusplus.VirtualCall,-google-explicit-constructor'
```
pivotal-jbarrett: This tests you have excluded should be passing, why exclude it?

cppcache/CMakeLists.txt
```diff
@@ -24,6 +24,28 @@ if (CMAKE_USE_PTHREADS_INIT)
   check_function_exists("pthread_setname_np" HAVE_pthread_setname_np)
 endif()
 
+# Add clang-tidy if available
+if(CMAKE_VERSION VERSION_GREATER 3.6)
```
pivotal-jbarrett: So we used to have something like this in CMake before. It was removed because clang-tidy doesn't handle parallel execution well. It will "fix" the same header included via two different source files in parallel and produce garbage output.

cppcache/src/CacheImpl.hpp
```diff
@@ -84,12 +83,13 @@ class TcrConnectionManager;
...
  */
 
-class APACHE_GEODE_EXPORT CacheImpl : private NonCopyable,
-                                      private NonAssignable {
+class APACHE_GEODE_EXPORT CacheImpl {
   /**
    * @brief public methods
    */
  public:
+  CacheImpl(const CacheImpl&) = delete;
```
pivotal-jbarrett: Yay!

cppcache/src/CacheImpl.hpp
```diff
@@ -191,7 +191,7 @@ class APACHE_GEODE_EXPORT CacheImpl : private NonCopyable,
    */
   std::vector<std::shared_ptr<Region>> rootRegions();
 
-  virtual RegionFactory createRegionFactory(RegionShortcut preDefinedRegion);
+  RegionFactory createRegionFactory(RegionShortcut preDefinedRegion);
```
pivotal-jbarrett: Good find!

cppcache/src/ClientProxyMembershipID.cpp
```diff
@@ -51,7 +51,7 @@ ClientProxyMembershipID::ClientProxyMembershipID()
       m_hostAddrLocalMem(false),
       m_vmViewId(0) {}
 
-ClientProxyMembershipID::~ClientProxyMembershipID() noexcept {
+ClientProxyMembershipID::~ClientProxyMembershipID() {
```
pivotal-jbarrett: Why remove the `noexcept`?

cppcache/src/TcrChunkedContext.hpp
```diff
@@ -106,11 +106,11 @@ class TcrChunkedContext {
                     TcrChunkedResult* result, uint8_t isLastChunkWithSecurity,
                     const CacheImpl* cacheImpl);
 
-  ~TcrChunkedContext() = default;
+  ~TcrChunkedContext();
 
   const uint8_t* getBytes() const;
 
-  size_t getLen() const;
+  int32_t getLen() const;
```
pivotal-jbarrett: Why the type change? The implementation is returning the size of a `std::vector` which is `size_t`.

.clang-tidy
mreddington: This is intentional. We want to use universal initialization, using the "explicit" keyword disables that for those constructors.

cppcache/CMakeLists.txt
mreddington: I'll remove the "fix" option but leave generating the report in place. It still fails the build if the check doesn't pass.

cppcache/src/ClientProxyMembershipID.cpp
mreddington: Per C++11 spec, all destructors and only destructors are implicitly noexcept(true). that makes this redundant. There are edge cases where you can willfully break things but I'm not going to quote every nuance of the spec.

cppcache/src/TcrChunkedContext.hpp
mreddington: This was rebase damage. I'm still combing.

.clang-tidy
pivotal-jbarrett: I am not familiar with that being the case. Can you point me to some documentation? `explicit` should just be keeping decay-able parameters from implicitly matching on the constructor.

cppcache/src/ClientProxyMembershipID.cpp
pivotal-jbarrett: https://en.cppreference.com/w/cpp/language/noexcept_spec
> default constructors, copy constructors, move constructors that are implicitly-declared or defaulte on their first declaration unless
The keyword being "default". In this case these destructors are not default. Can you point me at the spec that says non-default are also implicitly `noexcept`?

pivotal-jbarrett:
Please reduce the scope and ongoing changes to this PR. It’s really hard to kept coming back to PRs for continuous change and review.

PR #497, opened 2019-06-21
GEODE-6844: Header cleanup

mreddington:
Method bodies in header files were moved into source files under the cppcache/src/ directory.

Some classes had their visibility set to public because their unit tests are in the integration test suite. We'll have to sort that later.


pivotal-jbarrett:
Please provide a JIRA that describes the purpose for this change. Describe in your PR/commit how these changes achieve the goals of the JIRA.

cppcache/integration-test/testLRUList.cpp
```diff
@@ -29,7 +29,7 @@ END_TEST(NotOnWindows)
 
 //#define BUILD_CPPCACHE
 
-#include <LRUList.cpp>
+#include <LRUList.hpp>
```
pivotal-jbarrett: Haha.

mreddington: GEODE-6844

This is a much simpler PR than the last attempt, I hope you'll agree.

pivotal-jbarrett:
Can you please update the JIRA with the rational for making these changes? I assume it's and attempt to speed up build times on internal source files. If that is the case can you provide insight in the PR as to how much this change set improves time?

cppcache/src/Log.cpp
```diff
@@ -804,6 +804,217 @@ void LogVarargs::finest(const char* fmt, ...) {
...
+#endif
+  return (((s_doingDebug && level == LogLevel::Debug) ||
+           GEODE_HIGHEST_LOG_LEVEL >= level) &&
+          s_logLevel >= level);
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
+}
+
+void Log::log(LogLevel level, const char* msg) {
```
pivotal-jbarrett: You need to make darn sure that these are going to get inlined now that they are not in the header or we might start having a real performance issue.

cppcache/src/util/string.hpp
```diff
@@ -38,133 +38,19 @@ constexpr std::codecvt_mode codecvt_mode_native_endian =
...
-      std::wstring_convert<std::codecvt_utf8_utf16<int16_t>, int16_t>{}
-          .from_bytes(utf8);
-  return std::u16string(reinterpret_cast<const char16_t*>(int16String.data()),
-                        int16String.size());
-#else
-  return std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>{}
-      .from_bytes(utf8);
-#endif
-}
+std::u16string to_utf16(const std::string& utf8);
```
pivotal-jbarrett: Again you have removed inline optimizations for something that is heavily used outside of this compilation unit. Without whole program optimization support in the linker we may start seeing a loss here.

cppcache/src/util/concurrent/spinlock_mutex.hpp
```diff
@@ -22,21 +22,21 @@
...
 
+#include "apache-geode_export.h"
+
 namespace apache {
 namespace geode {
 namespace util {
 namespace concurrent {
 
-class spinlock_mutex final {
+class APACHE_GEODE_EXPORT spinlock_mutex final {
```
pivotal-jbarrett: Why is this being exported?

cppcache/src/util/string.hpp
mreddington: Our intent is to run a profile pass on the native client and selectively optimize, with documentation.

cppcache/src/util/concurrent/spinlock_mutex.hpp
mreddington: ::sigh:: This is one of my gripes - if you look, you'll find some +30 classes being exported from the cppcache/src directory. Why are any of them exported? It turns out, all our integration tests rely heavily on the private implementation of the native client, either indirectly or directly. That means a fair number of those tests are actually unit tests, and the rest are innocent victims. They were exported because someone wrote a test in the integration-test framework and didn't understand why it wouldn't compile, so they just exported it and the problem magically went away, not understanding why it is this way.

We're going to make a pass and fix all the tests, a combination of reorganizing and breaking those dependencies. For now, this will have to do, and we wouldn't have known this problem existed with this class until I did this.

pivotal-jbarrett: I understand why all the previous classes had to be exported. The old integration tests are a mess. This class however only has a unit test and does not need to be exported.

cppcache/src/util/string.hpp
pivotal-jbarrett: Not keen on the concept of throwing out the baby with the bathwater only to later come back and later selectively test. Without testing these changes you have no idea what the impact of these changes have on the profile of the library.

cppcache/src/Log.cpp
mreddington: We're replacing the logger with a logger library, so this point is moot.

pdxcodemonkey: Wasn't under the impression this would require the heavier-weight process with a proposal etc.  We're not planning to modify the API or the output of the logger, only replace the code in the Log class, which is _one file_.  If you disagree, let me know and I'll write something up and bring it to the dev list.

pivotal-jbarrett: Introduction of new dependencies should have a discussion. That discussion can be a simple as a PR. In this case I’m asking to about a potential negative impact caused by this change and being told it’s being addressed in a wholesale replacement of logging. How as a community member am I supposed to be aware of this intent. In the mean time of this PR is accepted we don’t know the impact this intermediate change is going to have. No effort has been expended to quantify the cost of this change.

mreddington:
Performance benchmarks can be found here: https://docs.google.com/spreadsheets/d/1N50kDTOZsUWDTP2c5ejt9niGS-1al5Yox8O21FC5U-s/

In conclusion, the difference in runtime performance is insignificant and build times improved (which was the desired outcome).

cppcache/src/util/string.hpp
mreddington: Performance pass is done. See the comment below.

cppcache/src/Log.cpp
mreddington: Benchmarks are available, inlining of the logger had no runtime performance impact.

PR #499, opened 2019-06-27
GEODE-6795: Update and correct Native Client XSD

davebarnes97:
Work in progress - took this as far as I was able, soliciting input.

### DIFF
--- a/xsds/cpp-cache-1.10-DRAFT.xsd
+++ b/xsds/cpp-cache-1.10-DRAFT.xsd
@@ -1,0 +1,236 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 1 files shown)


xsds/cpp-cache-1.10-DRAFT.xsd
```diff
@@ -0,0 +1,235 @@
...
+                  </xsd:element>
+                </xsd:sequence>
+              </xsd:extension>
+            </xsd:complexContent>
+          </xsd:complexType>
+        </xsd:element>
+      </xsd:choice>
+    </xsd:sequence>
+    <xsd:attribute name="caching-enabled" type="xsd:boolean" />
+    <xsd:attribute name="cloning-enabled" type="xsd:boolean" />
```
pdxcodemonkey: Still a duplicate attribute here, apparently...

PR #500, opened 2019-07-03
GEODE 6835 - fix auth retry logic

pdxcodemonkey:
This turned out to be caused by (as far as we can tell) a _completely different_ issue than retry logic for authentication.  The actual issue is/was that the native client was sending an UPDATE_CLIENT_NOTIFICATION request to the server, and this request hasn't been supported on the server for a very long time, like several years.  When the server gets this request in a scenario without a security manager, it is simply ignored, so the behavior was never discovered.  With a security manager, however, the server first attempts to retrieve the unique ID part of the request from the end, which should be there for an authenticated connection, and when it doesn't find the secured ID it throws a ClientAuthenticationRequired exception and closes the connection.  Fix is to stop sending this request altogether, since at the very best it was accomplishing nothing.  This PR also includes a number of simple refactors (variable renaming, etc) found in the course of debugging, as well as a new test case for this scenario and required enhancements to the integration test framework. 

@mreddington @steve-sienk

### DIFF
--- a/cppcache/integration/framework/Cluster.cpp
+++ b/cppcache/integration/framework/Cluster.cpp
@@ -33,7 +33,7 @@ void Locator::start() {
     cluster_.getGfsh().stop().locator().withDir(name_).execute();
   }
 
-  auto locator = cluster_.getGfsh()
+  cluster_.getGfsh()
       .start()
       .locator()
       .withDir(name_)
@@ -42,23 +42,10 @@ void Locator::start() {
       .withPort(locatorAddress_.port)
       .withMaxHeap("256m")
       .withJmxManagerPort(jmxManagerPort_)
-      .withHttpServicePort(0);
-
-  if (!cluster_.getClasspath().empty()) {
-    locator.withClasspath(cluster_.getClasspath());
-  }
-
-  if (!cluster_.getSecurityManager().empty()) {
-    locator.withSecurityManager(cluster_.getSecurityManager());
-  }
-
-  if ((!cluster_.getUser().empty()) && (!cluster_.getPassword().empty())) {
-    locator.execute(cluster_.getUser(), cluster_.getPassword());
-  } else {
-    locator.execute();
-  }
-
-//  std::cout << "locator: " << locatorAddress_.port << ": started" << std::endl << std::flush;
+      .withHttpServicePort(0)
+      .withClasspath(cluster_.getClasspath())
+      .withSecurityManager(cluster_.getSecurityManager())
+      .execute(cluster_.getUser(), cluster_.getPassword());
 
   started_ = true;
 }
@@ -74,7 +61,7 @@ void Server::start() {
   auto safeName = name_;
   std::replace(safeName.begin(), safeName.end(), '/', '_');
 
-  auto server = cluster_.getGfsh()
+  cluster_.getGfsh()
       .start()
       .server()
       .withDir(name_)
@@ -83,25 +70,12 @@ void Server::start() {
       .withPort(serverAddress_.port)
       .withMaxHeap("1g")
       .withLocators(locators_.front().getAddress().address + "[" +
-                    std::to_string(locators_.front().getAddress().port) + "]");
-
-  if (!cluster_.getClasspath().empty()) {
-    server.withClasspath(cluster_.getClasspath());
-  }
-
-  if (!cluster_.getSecurityManager().empty()) {
-    server.withSecurityManager(cluster_.getSecurityManager());
-  }
-
-  if (!cluster_.getUser().empty()) {
-    server.withUser(cluster_.getUser());
-  }
-
-  if (!cluster_.getPassword().empty()) {
-    server.withPassword(cluster_.getPassword());
-  }
-
-  server.execute();
+                    std::to_string(locators_.front().getAddress().port) + "]")
+      .withClasspath(cluster_.getClasspath())
+      .withSecurityManager(cluster_.getSecurityManager())
+      .withUser(cluster_.getUser())
+      .withPassword(cluster_.getPassword())
+      .execute();
 
 //  std::cout << "server: " << serverAddress_.port << ": started" << std::endl << std::flush;
 
--- a/cppcache/integration/framework/Cluster.cpp
+++ b/cppcache/integration/framework/Cluster.cpp
@@ -33,7 +33,7 @@ void Locator::start() {
     cluster_.getGfsh().stop().locator().withDir(name_).execute();
   }
 
-  cluster_.getGfsh()
+  auto locator = cluster_.getGfsh()
       .start()
       .locator()
       .withDir(name_)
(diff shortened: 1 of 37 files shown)


cppcache/integration/framework/Cluster.cpp
```diff
@@ -42,27 +42,39 @@ void Locator::start() {
       .withPort(locatorAddress_.port)
       .withMaxHeap("256m")
       .withJmxManagerPort(jmxManagerPort_)
-      .withHttpServicePort(0)
-      .execute();
+      .withHttpServicePort(0);
+
+  if (!cluster_.getClasspath().empty()) {
```
pivotal-jbarrett: Why check this, just set the empty string on the locator class. All the lines above don't care about empty strings.

cppcache/integration/framework/Cluster.cpp
```diff
@@ -71,19 +83,35 @@ void Server::start() {
       .withPort(serverAddress_.port)
       .withMaxHeap("1g")
       .withLocators(locators_.front().getAddress().address + "[" +
-                    std::to_string(locators_.front().getAddress().port) + "]")
-      .execute();
+                    std::to_string(locators_.front().getAddress().port) + "]");
+
+  if (!cluster_.getClasspath().empty()) {
```
pivotal-jbarrett: Same, why worry about empty strings.

cppcache/integration/test/AuthInitializeTest.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <geode/RegionShortcut.hpp>
+
+#include "CacheRegionHelper.hpp"
+#include "SimpleAuthInitialize.hpp"
+#include "SimpleCqListener.hpp"
+#include "framework/Cluster.h"
+#include "framework/Framework.h"
+#include "framework/Gfsh.h"
+
+namespace {
```
pivotal-jbarrett: Turns out this doesn't really do anything in cpp file since the compilation units won't leak namespaces anyway.

cppcache/integration/test/AuthInitializeTest.cpp
```diff
@@ -0,0 +1,216 @@
...
+      .execute();
+  auto authInitialize = std::make_shared<SimpleAuthInitialize>(
+      "unauthorized-user", "root-password");
+  auto cache = createCache(authInitialize);
+  auto pool = createPool(cluster, cache, false);
+  auto region = setupRegion(cache, pool);
+
+  try {
+    region->put("foo", "bar");
+  } catch (const apache::geode::client::NotConnectedException&) {
```
pivotal-jbarrett: Why not include these in the `using` block so we can use the short name?

cppcache/integration/test/CMakeLists.txt
```diff
@@ -13,9 +13,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-file(GLOB_RECURSE SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
-
-add_executable(cpp-integration-test ${SOURCES})
+add_executable(cpp-integration-test
```
pivotal-jbarrett: yay, no globs!

cppcache/src/AdminRegion.cpp
```diff
@@ -70,7 +70,7 @@ TcrConnectionManager* AdminRegion::getConnectionManager() {
 void AdminRegion::put(const std::shared_ptr<CacheableKey>& keyPtr,
                       const std::shared_ptr<Cacheable>& valuePtr) {
   GfErrType err = putNoThrow(keyPtr, valuePtr);
-  GfErrTypeToException("AdminRegion::put", err);
+  throwExceptionIfError("AdminRegion::put", err);
```
pivotal-jbarrett: Good rename!

cppcache/src/CqQueryImpl.cpp
```diff
@@ -273,7 +273,7 @@ GfErrType CqQueryImpl::execute(TcrEndpoint* endpoint) {
   err = m_tccdm->sendRequestToEP(request, reply, endpoint);
 
   if (err != GF_NOERR) {
-    // GfErrTypeToException("CqQuery::execute(endpoint)", err);
+    // throwExceptionIfError("CqQuery::execute(endpoint)", err);
```
pivotal-jbarrett: Delete commented code.

cppcache/src/ErrType.hpp
```diff
@@ -40,7 +40,7 @@ typedef enum {
   GF_NOTSUP = 12,         /**< operation not supported          */
   GF_SCPGBL = 13,         /**< attempt to exit global scope     */
   GF_SCPEXC = 14,         /**< maximum scopes exceeded          */
-  GF_TIMOUT = 15,         /**< operation timed out              */
+  GF_TIMEOUT = 15,        /**< operation timed out              */
```
pivotal-jbarrett: Hahah

cppcache/src/TcrConnection.cpp
```diff
@@ -943,9 +943,14 @@ void TcrConnection::readMessageChunked(TcrMessageReply& reply,
       header = readChunkHeader(headerTimeout);
     }
   } catch (const Exception&) {
-    auto ex = reply.getChunkedResultHandler()->getException();
-    LOGDEBUG("Found existing exception ", ex->what());
-    reply.getChunkedResultHandler()->clearException();
+    auto handler = reply.getChunkedResultHandler();
```
pivotal-jbarrett: I prefer the shorthand for scoping reasons.
```c++
if (auto handler = reply.getChunkedResultHandler()) {
  if (auto ex = handler->getException()) {
    LOGDEBUG("Found existing exception ", ex->what());
    handler->clearException();
  }
}
```

cppcache/src/TcrEndpoint.cpp
```diff
@@ -828,7 +812,7 @@ GfErrType TcrEndpoint::sendRequestConn(const TcrMessage& request,
                                   reply.getTimeout(), request.getMessageType());
     reply.setMessageTypeRequest(type);
     reply.setData(
-        data, static_cast<int32_t>(dataLen), this->getDistributedMemberID(),
+        data, static_cast<int32_t>(dataLen), getDistributedMemberID(),
```
pivotal-jbarrett: 👍

cppcache/src/TcrMessage.cpp
```diff
@@ -2835,13 +2825,20 @@ void TcrMessage::addSecurityPart(int64_t connectionId, int64_t unique_id,
...
+               .c_str());
+
   writeObjectPart(encryptBytes);
   writeMessageLength();
   m_securityHeaderLength = 4 + 1 + encryptBytes->length();
-  LOGDEBUG("TcrMessage addsp = %s ",
-           Utils::convertBytesToString(m_request->getBuffer(),
-                                       m_request->getBufferLength())
-               .c_str());
+  //  LOGDEBUG("TcrMessage addsp = %s ",
```
pivotal-jbarrett: Delete

PR #504, opened 2019-07-30
GEODE-7019: Fix closing of idle connections in C++ Native client

albertogpz:


cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -429,7 +429,7 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
       }
     }
 
-    size_t replaceCount =
+    int replaceCount =
```
pdxcodemonkey: As long as you're changing the declaration of replaceCount, please make it 'auto' rather than an explicit type.

cppcache/src/PoolFactory.cpp
```diff
@@ -74,7 +74,7 @@ PoolFactory::PoolFactory(const Cache& cache)
 PoolFactory& PoolFactory::setFreeConnectionTimeout(
     std::chrono::milliseconds connectionTimeout) {
   if (connectionTimeout <= std::chrono::milliseconds::zero()) {
-    throw IllegalArgumentException("connectionTimeout must greater than 0.");
+    throw IllegalArgumentException("connectionTimeout must be greater than 0.");
```
pdxcodemonkey: These strings are repetitive, how about we use a couple of named constants?  e.g.:

```
const std::string greaterThanZeroWarning = " must be greater than 0.";
...

throw IllegalArgumentException("connectionTimeout" + greaterThanZeroWarning);
```
Thanks for fixing the typos and grammar!

.lgtm.yml
```diff
@@ -4,9 +4,9 @@ extraction:
       command:
         - mkdir _lgtm_build_dir
         - cd _lgtm_build_dir
-        - wget -O apache-geode.zip http://mirror.transip.net/apache/geode/1.7.0/apache-geode-1.7.0.zip
-        - unzip apache-geode.zip
-        - cmake -DGEODE_ROOT="`pwd`/apache-geode-1.7.0" ..
+        - wget -O apache-geode.tgz http://mirror.transip.net/apache/geode/1.9.0/apache-geode-1.9.0.tgz
+        - tar xzf apache-geode.tgz
+        - cmake -DCMAKE_INSTALL_PREFIX="`pwd`/apache-geode-native-install" -DGEODE_ROOT="`pwd`/apache-geode-1.9.0" ..
```
pivotal-jbarrett: We don't need to set the install directory, especially if we aren't installing it.

CONTRIBUTING.md
```diff
@@ -41,6 +41,16 @@ $ ctest -R <test_name> -C <Debug|Release>
 ```
 .NET integration tests can be executed similarly from `build/clicache/integration-test`.
 
+### Running new integration tests
```
pivotal-jbarrett: 👍 thanks for adding this section.

cppcache/integration/test/CleanIdleConnections.cpp
```diff
@@ -0,0 +1,136 @@
...
+      .execute();
+
+  auto cache = createCache();
+  auto minConns = 1;
+  auto maxConns = -1;
+  char poolName[] = "default";
+  auto pool = createPool(cluster, cache, minConns, maxConns, poolName);
+  auto region = setupRegion(cache, pool);
+
+  int poolSize = getCacheImpl(&cache)->getPoolSize(poolName);
```
pivotal-jbarrett: If you go back into this PR for other things it would be preferable for you to update these internal APIs to take a `const std::string &`. All the external APIs have been updated but very few of the internals. In the future if you hit any `char*` c-style strings please make them go away.

cppcache/src/ThinClientPoolDM.cpp
albertogpz: Sure

cppcache/integration/test/CleanIdleConnections.cpp
albertogpz: ok

.lgtm.yml
albertogpz: I agree. It was something I added to try fix the problem with the linker with LGTM but obviously did not help.

cppcache/src/PoolFactory.cpp
albertogpz: I'd rather not use a constant and leave this as it it.
I think the code is more clear with the text, even if it is repetitive. Also if I use the constant, I would have to leave out the final stop in the sentence because at least one log add something after "greater than zero".

albertogpz:
@pivotal-jbarrett Would you please review the latest changes and if you agree merge the PR?

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -408,93 +408,90 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
...
-      if (replaceCount <= 0) {
-        GF_SAFE_DELETE_CON(conn);
-        removeEPConnections(1, false);
-        getStats().incLoadCondDisconnects();
-        LOGDEBUG("Removed a connection");
+  for (std::vector<TcrConnection*>::const_iterator iter = removelist.begin();
+       iter != removelist.end(); ++iter) {
+    TcrConnection* conn = *iter;
+    if (replaceCount <= 0) {
+      GF_SAFE_DELETE_CON(conn);
```
pivotal-jbarrett: For another PR, sure would be nice to replace all these raw pointers with `std::shared_ptr` or `std::unique_ptr`. Look at all these places we delete, and potentially leak, connections.

PR #507, opened 2019-08-01
Geode-7041: Add FunctionService::OnServers Test in New Framework

mmartell:
In addition to adding the test to the new framework, this PR includes updates to the framework itself to support server side configuration via XML.


cppcache/integration/test/FunctionExecutionTest.cpp
```diff
@@ -145,20 +148,98 @@ TEST(FunctionExecutionTest,
       .jar(getFrameworkString(FrameworkVariable::JavaObjectJarPath))
       .execute();
 
-  auto cache = CacheFactory().set("log-level", "none").create();
-  auto pool = cache.getPoolManager()
-                  .createFactory()
-                  .addLocator("localhost", 10334)
-                  .create("pool");
+  auto cache = cluster.createCache();
+  //  CacheFactory().set("log-level", "none").create();
```
moleske: just a heads up for this also 🙂

cppcache/integration/framework/Cluster.h
```diff
@@ -257,32 +297,29 @@ class Cluster {
...
+  std::string &getSecurityManager() { return securityManager_; }
 
-  std::string& getPassword() {
-    return password_;
-  }
+  std::string &getUser() { return user_; }
+
+  std::string &getPassword() { return password_; }
+
+  //std::string &getCacheXMLFile() { return cacheXMLFiles_[0]; }
```
moleske: Looks like some missed commented out code

cppcache/integration/framework/Cluster.cpp
```diff
@@ -89,7 +90,7 @@ void Server::stop() {
   started_ = false;
 }
 
-void Cluster::start() {
+void Cluster::start(std::function<void()> fn) {
```
moleske: I would feel a little bit better if we could name `fn` to represent what kind of function we intend to pass in here (but not feeling too strongly about).

cppcache/integration/test/FunctionExecutionTest.cpp
mmartell: Actually, this test (FunctionReturnsObjectWhichCantBeDeserializedOnServer) is disabled because it had a few problems. We created another ticket for it so I just disabled it. But doesn't hurt to delete the commented out piece.

cppcache/integration/framework/Cluster.cpp
mmartell: I agree. fn is never a good name for a function.  Renaming to extraGfshCommands().

cppcache/integration/framework/Cluster.h
mmartell: Good catch.

cppcache/integration/test/FunctionExecutionTest.cpp
pdxcodemonkey: Yeah, fixing this function is PR #506 :)

moleske: ah ok, glad this is just a case of me not keeping up on the other PRs

PR #509, opened 2019-08-07
GEODE-7056: Update Geode version in CI

alb3rtobr:


alb3rtobr:
Someone with permissions should generate the image and upload it to dockerhub.

PR #510, opened 2019-08-14
GEODE-7086: coding impacts for support of IPv6

mivanac:
These are modifications in geode-native driver to run it on IPv6 environment.

### DIFF
--- a/cppcache/src/ClientProxyMembershipID.cpp
+++ b/cppcache/src/ClientProxyMembershipID.cpp
@@ -57,11 +57,11 @@ ClientProxyMembershipID::~ClientProxyMembershipID() noexcept {
 
 ClientProxyMembershipID::ClientProxyMembershipID(
     std::string dsName, std::string randString, const char* hostname,
-    uint32_t hostAddr, uint32_t hostPort, const char* durableClientId,
-    const std::chrono::seconds durableClntTimeOut)
-    : m_hostAddrAsUInt32(hostAddr) {
+	const char* hostAddr, int len, uint32_t hostPort, const char* durableClientId,
+    const std::chrono::seconds durableClntTimeOut) {
   auto vmPID = boost::this_process::get_id();
-  initObjectVars(hostname, reinterpret_cast<uint8_t*>(&m_hostAddrAsUInt32), 4,
+
+  initObjectVars(hostname, (uint8_t*) hostAddr, len,
                  false, hostPort, durableClientId, durableClntTimeOut, DCPORT,
                  vmPID, VMKIND, 0, dsName.c_str(), randString.c_str(), 0);
 }
@@ -101,12 +101,9 @@ void ClientProxyMembershipID::initObjectVars(
   m_vmViewId = vmViewId;
   m_memID.write(static_cast<int8_t>(DSCode::FixedIDByte));
   m_memID.write(static_cast<int8_t>(DSCode::InternalDistributedMember));
-  m_memID.writeArrayLen(ADDRSIZE);
-  // writing first 4 bytes of the address. This will be same until
-  // IPV6 support is added in the client
-  uint32_t temp;
-  memcpy(&temp, hostAddr, 4);
-  m_memID.writeInt(static_cast<int32_t>(temp));
+  uint8_t * temp = new uint8_t [hostAddrLen];
+  memcpy(temp, hostAddr, hostAddrLen);
+  m_memID.writeBytes(temp, hostAddrLen);
   // m_memID.writeInt((int32_t)hostPort);
   m_memID.writeInt(static_cast<int32_t>(synch_counter));
   m_memID.writeString(hostname);
--- a/cppcache/src/ClientProxyMembershipID.cpp
+++ b/cppcache/src/ClientProxyMembershipID.cpp
@@ -50,29 +50,56 @@ ClientProxyMembershipID::~ClientProxyMembershipID() noexcept {}
 
 ClientProxyMembershipID::ClientProxyMembershipID(
     std::string dsName, std::string randString, const char* hostname,
-    const uint8_t* hostAddr, uint32_t hostAddrLen, uint32_t hostPort,
+    const ACE_INET_Addr& address, uint32_t hostPort,
     const char* durableClientId,
     const std::chrono::seconds durableClntTimeOut) {
   auto vmPID = boost::this_process::get_id();
 
-  initObjectVars(hostname, hostAddr, hostAddrLen, false, hostPort,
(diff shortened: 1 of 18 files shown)


pdxcodemonkey:
@mivanac Just accidentally pushed a change to your fork, so sorry!  Reverted the change right away and pushed it to my fork where it belongs.

cppcache/src/ClientProxyMembershipID.cpp
```diff
@@ -101,12 +100,9 @@ void ClientProxyMembershipID::initObjectVars(
   m_vmViewId = vmViewId;
   m_memID.write(static_cast<int8_t>(DSCode::FixedIDByte));
   m_memID.write(static_cast<int8_t>(DSCode::InternalDistributedMember));
-  m_memID.writeArrayLen(ADDRSIZE);
-  // writing first 4 bytes of the address. This will be same until
-  // IPV6 support is added in the client
-  uint32_t temp;
-  memcpy(&temp, hostAddr, 4);
-  m_memID.writeInt(static_cast<int32_t>(temp));
+//  uint8_t * temp = new uint8_t [hostAddrLen];
```
pdxcodemonkey: This pattern of allocating memory and passing it into another class to take ownership of is still found in numerous places in geode-native.  SOP is to remove it when we find it, usually by making the buffer in question a vector of the type we want, in this case uint8_t.  I took the liberty at https://github.com/pdxcodemonkey/geode-native/commit/3baf8c190430c8d25c27086223d0aced616a3aa5, feel free to cherry-pick if you like.

mivanac: Thanks

cppcache/src/ClientProxyMembershipID.cpp
```diff
@@ -57,11 +57,11 @@ ClientProxyMembershipID::~ClientProxyMembershipID() noexcept {
 
 ClientProxyMembershipID::ClientProxyMembershipID(
     std::string dsName, std::string randString, const char* hostname,
-    uint32_t hostAddr, uint32_t hostPort, const char* durableClientId,
-    const std::chrono::seconds durableClntTimeOut)
-    : m_hostAddrAsUInt32(hostAddr) {
+	const char* hostAddr, int len, uint32_t hostPort, const char* durableClientId,
```
mhansonp: Why are we bothering to pass around anything, but a string and a port? The IP Address should only matter at the lowest levels and in a certain sense not at all. We don't care that "142.156.37.2" or "www.whatever.com". I am not sure I understand this code, but it seems pretty clear we don't want to continue dealing with IPs.

pdxcodemonkey: AFAIK, ClientProxyMembershipID is used *only* for serializing into a handshake message to initiate a server connection.  The IP address(es) are baked into the protocol here, so I don't think you can get rid of the logic.

pivotal-jbarrett: Yup, it's needed by the protocol / message here unfortunately.

cppcache/src/TcrConnection.cpp
```diff
@@ -173,7 +173,22 @@ bool TcrConnection::initTcrConnection(
...
     ACE_INET_Addr driver(hostName);
-    uint32_t hostAddr = driver.get_ip_address();
+    uint8_t hostaddr[16];
+    int len;
+    if (driver.get_type() == AF_INET6) {
+      const struct sockaddr_in6* sa6 =
+          static_cast<const struct sockaddr_in6*>(driver.get_addr());
+      auto saddr = reinterpret_cast<const uint8_t*>(&sa6->sin6_addr);
+      len = sizeof(sa6->sin6_addr);
+      memcpy(hostaddr, saddr, len);
```
pivotal-jbarrett: This is where I would just put the host adder into the vector so there is no question of ownership later.

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -178,7 +178,23 @@ ThinClientPoolDM::ThinClientPoolDM(const char* name,
   ACE_TCHAR hostName[256];
   ACE_OS::hostname(hostName, sizeof(hostName) - 1);
   ACE_INET_Addr driver(hostName);
-  uint32_t hostAddr = driver.get_ip_address();
+
+  uint8_t hostaddr[16];
```
pivotal-jbarrett: Can we create a common method so we don't have duplicate IPv4 and IPv6 code in multiple places?

cppcache/src/PoolFactory.cpp
```diff
@@ -308,9 +308,15 @@ PoolFactory& PoolFactory::addCheck(const std::string& host, int port) {
   }
 
   ACE_INET_Addr addr(port, host.c_str());
-  if (!(addr.get_ip_address())) {
+
```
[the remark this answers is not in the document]
mivanac: This is modified, because for IPV6, where ever method get_ip_address() is called, error is reported, and get_addr() must be used. I added this check, to verify is hostname resolved.

cppcache/src/ThinClientPoolDM.cpp
pdxcodemonkey: I've been wondering about this all morning.  Given that there are multiple instances of this code, and they all just generate this address buffer, pass it into the ClientProxyMembershipID ctor and then forget about it, will we lose any generality if we just remove the hostaddr parameter from that ctor and do this work inside ClientProxyMembershipID?

pivotal-jbarrett: It's an option to try. Both are better than duplicated logic.

cppcache/src/PoolFactory.cpp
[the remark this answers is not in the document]
pivotal-jbarrett: I think just checking that `get_addr()` is not `nullptr` should suffice.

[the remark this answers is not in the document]
mivanac: Tried this proposal, and in test, in case hostname is not resolved, get_addr() returns pointer (it is not nullptr).

pdxcodemonkey:
Quick update-

We fixed a very minor build issue on Windows, and are now working on
getting test boxes routing IPv6 locally.  As soon as that's complete, Ivan
or I will merge the PR.

Thanks,

Blake

mivanac:
Hi,
is there any update. Could we merge this PR?

BR

pdxcodemonkey:
I'll check with Ivan when he gets in.  I'd rather be safe than sorry with respect to our test machines, but if it holds up the works too much longer I'll go ahead and merge, no worries.

PR #517, opened 2019-09-11
GEODE-6007: Fix LGTM issues

pdxcodemonkey:
This doesn't catch _all_ of the issues, because the LGTM web UI makes it quite difficult when there are a ton of issues in the report like this, but it gets _almost_ all the way there.  Plan is to get this one merged, then look at the clean report from LGTM and fix the remaining < 10 issues.

@mreddington @steve-seink @vfordpivotal

### DIFF
--- a/clicache/integration-test/PutGetTestsN.cs
+++ b/clicache/integration-test/PutGetTestsN.cs
@@ -240,8 +240,8 @@ namespace Apache.Geode.Client.UnitTests
         catch (Exception ex)
         {
           Util.Log("Exception: Put caused networkhop ");
-          Util.Log("Got Exception (0} {1} {2} ", ex.Message, ex.StackTrace, ex.Source);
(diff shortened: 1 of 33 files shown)


pivotal-jbarrett:
Are there any additional clang-tidy or compiler warning rules we can enable now that these are addressed?

pdxcodemonkey:
Probably?  I'll look into it, but I don't think it's required for this PR.  The LGTM checks in PRs will keep us from introducing new issues.

FYI, at the time I originally put this together I did look into using the LGTM IDE plugin(s) a little bit, but they don't _appear_ to be FOSS.  What I found was references to "asking your LGTM administrator" for the download, so it kinda looks like they're expecting you to have a paid license of some kind to use it.  This is why it's hard to clean up the last few of these - no IDE plugin available, and the web UI for LGTM doesn't show you a view with _remaining_ issues, only the whole original report and a count of the ones you fixed.

PR #518, opened 2019-09-12
GEODE-7180: Disable IPv6 support by default

pdxcodemonkey:
This gets tests passing again until we can sort out the issues with the fix for GEODE-7086.


cppcache/src/TcpConn.cpp
```diff
@@ -94,7 +94,11 @@ void TcpConn::createSocket(ACE_HANDLE sock) {
 }
 
 void TcpConn::init() {
+#ifdef SUPPORT_IPV6
```
pivotal-jbarrett: Where is this macro defined?

PR #521, opened 2019-09-15
GEODE-7209: Modify check if hostname is resolved

mivanac:


cppcache/src/PoolFactory.cpp
```diff
@@ -310,16 +310,7 @@ PoolFactory& PoolFactory::addCheck(const std::string& host, int port) {
...
-  const int maxhostlength = 256;
-  char char_array[maxlength];
-  char char_localhost[maxhostlength];
-
-  addr.get_host_name(char_array, maxlength);
-  gethostname(char_localhost, maxhostlength);  // get local hostname
-
-  if ((strcmp(char_localhost, host.c_str()) != 0) &&
-      (strcmp(char_localhost, char_array) == 0)) {
+  if (port != addr.get_port_number()) {
```
pivotal-jbarrett: How does this assert that the hostname is known?

mivanac: According to ACE_INET_Addr implementation we can not get reply if hostname is resolved (just error printout):

// Creates a ACE_INET_Addr from a PORT_NUMBER and the remote
// HOST_NAME.

ACE_INET_Addr::ACE_INET_Addr (u_short port_number,
                              const char host_name[],
                              int address_family)
  : ACE_Addr (determine_type (), sizeof (inet_addr_))
{
  ACE_TRACE ("ACE_INET_Addr::ACE_INET_Addr");
  ACE_OS::memset (&this->inet_addr_, 0, sizeof (this->inet_addr_));

  if (this->set (port_number,
                 host_name,
                 1,
                 address_family) == -1)
    ACELIB_ERROR ((LM_ERROR,
                ACE_TEXT ("ACE_INET_Addr::ACE_INET_Addr: %p\n"),
                ACE_TEXT_CHAR_TO_TCHAR ((host_name == 0) ?
                                        "<unknown>" : host_name)));

}

mivanac: So with deeper investigation I learned, that in  function
```
// Initializes a ACE_INET_Addr from a PORT_NUMBER and the remote
// HOST_NAME.

int
ACE_INET_Addr::set (u_short port_number,
                    const char host_name[],
                    int encode,
                    int address_family)
```
port will be set only in case hostname is resolved (there is no error).
So this is the reason why port is checked.

pivotal-jbarrett: Ok, cool. Maybe put a comment in the source since unless you look at the ACE source this would not be intuitive. 
`// ACE will not initialize port if hostname is not resolved.`

mivanac: Thanks for comment.

PR #523, opened 2019-09-20
GEODE-7225: Change download URL from 'c' to 'd'

pdxcodemonkey:
- reflects update on OpenSSL site


pdxcodemonkey:
@mreddington @steve-sienk @vford-pivotal

PR #524, opened 2019-09-20
GEODE-7203: Fix build break in examples

pdxcodemonkey:
- std::endl doesn't work in string concatenation expression

@mreddington @steve-sienk @vfordpivotal @charliemblack @dihardman


pdxcodemonkey:
@moleske yes, `\n` will work everywhere

PR #525, opened 2019-09-20
GEODE-7203: Update transaction example README files

davebarnes97:
Fine-tuning the explanation of example output. Applying parallel edits to both C++ and .NET versions.


dihardman:
Looks good, thanks!

PR #526, opened 2019-09-20
GEODE-7207: Fix remaining LGTM "Error" level alerts

pdxcodemonkey:


clicache/integration-test/ThinClientPdxTests1.cs
```diff
@@ -1872,7 +1872,7 @@ public MyClass(int key)
     {
       Key = key.ToString();
       SecKey = key;
-      ShareQuantity = (double)((long)key * 9278L);
+      ShareQuantity = key * 9278L;
```
steve-sienk: LGTM: This didn't actually get fixed, ShareQuantity is a double.

clicache/integration-test/UnitTestsN.cs
```diff
@@ -174,7 +174,7 @@ public void LogTaskTiming(ClientBase client, string taskName, int numOps)
       TimeSpan elapsed = m_endTime - m_startTime;
       Util.Log("{0}Time taken for task [{1}]: {2}ms {3}ops/sec{4}",
         Util.MarkerString, taskName, elapsed.TotalMilliseconds,
-        ((long)numOps * 1000) / elapsed.TotalMilliseconds, Util.MarkerString);
+        (numOps * 1000) / elapsed.TotalMilliseconds, Util.MarkerString);
```
steve-sienk: LGTM: This didn't actually get fixed, numOps is an int (not the expected long).

clicache/integration-test2/Cluster.cs
```diff
@@ -32,7 +32,6 @@ public class Cluster : IDisposable
         private int serverCount_;
         private bool started_;
         private List<Locator> locators_;
```
pivotal-jbarrett: Why remove the servers set? They are going to be needed to stop individual servers.

cppcache/src/ClientProxyMembershipID.hpp
```diff
@@ -95,14 +96,12 @@ class ClientProxyMembershipID : public DSMemberForVersionStamp {
...
+    std::stringstream hostAddressString;
+    hostAddressString << std::hex;
     for (uint32_t i = 0; i < getHostAddrLen(); i++) {
-      offset +=
-          std::snprintf(hostInfo + offset, 255 - offset, ":%x", m_hostAddr[i]);
+      hostAddressString << ":" << static_cast<int>(m_hostAddr[i]);
     }
-    result +=
-        internal::geode_hash<std::string>{}(std::string(hostInfo, offset));
+    result += internal::geode_hash<std::string>{}(hostAddressString.str());
```
pivotal-jbarrett: Nice!!

clicache/integration-test2/Cluster.cs
moleske: Was totally just fixing complaints, didn't actually look at the substance 😞 

Apparently we are never actually using the list the to stop individual servers, that should probably be the real fix instead of just deleting the list.

For record keeping, LGTM was pointing out that we added to the list of servers, but never actually retrieved any information from that list later.

PR #527, opened 2019-09-24
GEODE-7235: Disable a couple of new Clang warnings temporarily

pdxcodemonkey:
- Got a new version of the compiler from Apple, this is breaking Mac builds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,6 +190,10 @@ if (NOT HAVE_STDC_FORMAT_MACROS)
   target_compile_definitions(c++11 INTERFACE __STDC_FORMAT_MACROS)
 endif()
 
+if ("Darwin" STREQUAL ${CMAKE_SYSTEM_NAME} )
+  set(DISABLED_MAC_WARNINGS  -Wno-defaulted-function-deleted -Wno-c++2a-compat)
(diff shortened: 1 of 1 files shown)


pdxcodemonkey:
@mreddington @steve-sienk @vfordpivotal @charliemblack @dihardman

pdxcodemonkey:
Filed GEODE-7326 to track the actual fix.

pivotal-jbarrett:
@pdxcodemonkey 
I wish I would have been able to jump on this sooner. Matching based on the OS doesn't seem appropriate for something that is compiler specific. You should be matching compiler. Better yet, you should probably be using `check_cxx_compiler_flag(<flag> <var>)` to test that the version of the compiler in question even has the flags you are going to be passing to it. Since this is a temporary fix I am not going to get caught up on it. Please make sure the fix JIRA references these commits so that they can be backed out later.

pdxcodemonkey:
That's interesting - think I may have seen this feature in cmake and forgotten about it.  Hard to tell, given the ridiculous # of features available in cmake, but anyway this is clearly the right way to go.  I guess it's good in the sense that this will give me a strong incentive to get the code fix in place and back this out ASAP.

PR #529, opened 2019-09-29
GEODE-7222: refactor the build

mmartell:
The current apache-geode build system builds both a dynamic and a static library. This PR adds support for building only the static library and its dependencies. It provides for more efficient building of clients of apache-geode that wish to link it in statically.


mmartell:
Left in some bogus commits.

PR #530, opened 2019-09-29
GEODE-7222: refactor build

mmartell:
The current apache-geode build system builds both a dynamic and a static library. This PR adds support for building only the static library and its dependencies. It provides for more efficient building of clients of apache-geode that wish to link it in statically.


CMakeLists.txt
```diff
@@ -191,7 +192,7 @@ if (NOT HAVE_STDC_FORMAT_MACROS)
 endif()
 
 if ("Darwin" STREQUAL ${CMAKE_SYSTEM_NAME} )
-  set(DISABLED_MAC_WARNINGS  -Wno-defaulted-function-deleted -Wno-c++2a-compat)
+ set(DISABLED_MAC_WARNINGS  -Wno-defaulted-function-deleted -Wno-c++2a-compat)
```
pivotal-jbarrett: Spacing

cppcache/static/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ project(apache-geode-static LANGUAGES CXX)
 
 add_library(apache-geode-static STATIC ${SOURCES_ALL})
 
+generateExportFile( "apache-geode-static" )
```
pivotal-jbarrett: Technically this doesn’t need quoting. Also it would be nice to follow the naming convention of CMake.

CMakeLists.txt
mmartell: Done

cppcache/static/CMakeLists.txt
mmartell: Fixed to follow the cmake function naming convention. I realize cmake doesn't required the quotes in this case, but I prefer them.

PR #531, opened 2019-10-01
GEODE-7262: Split NC User Guide into language-specific editions

davebarnes97:


davebarnes97:
Picked up an extraneous commit - will close and reopen

PR #533, opened 2019-10-02
GEODE-7235: Add backwards compatibility for Clang

igodwin:
-  One of the new flags added for Clang 11 support, -Wno-defaulted-function-deleted, was not available to earlier versions of Clang

Co-authored-by: Michael Oleske <moleske@pivotal.io>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,8 +191,15 @@ if (NOT HAVE_STDC_FORMAT_MACROS)
   target_compile_definitions(c++11 INTERFACE __STDC_FORMAT_MACROS)
 endif()
 
-if ("Darwin" STREQUAL ${CMAKE_SYSTEM_NAME} )
-  set(DISABLED_MAC_WARNINGS  -Wno-defaulted-function-deleted -Wno-c++2a-compat)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -191,8 +191,12 @@ if (NOT HAVE_STDC_FORMAT_MACROS)
   target_compile_definitions(c++11 INTERFACE __STDC_FORMAT_MACROS)
 endif()
 
-if ("Darwin" STREQUAL ${CMAKE_SYSTEM_NAME} )
-  set(DISABLED_MAC_WARNINGS  -Wno-defaulted-function-deleted -Wno-c++2a-compat)
+if(${CMAKE_CXX_COMPILER_ID} STREQUAL "AppleClang")
```
pdxcodemonkey: Below, we set ignored compiler warnings based on the following test:
```
if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
```
if that check works, we should be able to use the same one here, and change the variable name to something like CONDITIONALLY_DISABLED_CLANG_WARNINGS, a lot more generic than "AppleClang" and MAC.

pdxcodemonkey: Come to think of it, we could/should check and disable these warnings separately, for correctness' sake.

PR #537, opened 2019-10-15
GEODE-7299: Fix PdxInstanceImpl memory leak

pdxcodemonkey:
updatePdxStream was stepping on the m_buffer value with a newly-allocated copy, thus leaking the existing m_buffer value.  Made m_buffer a std::shared_ptr, and let the runtime handle the memory management properly for us.

@mreddington @steve-sienk @vfordpivotal @dihardman @charliemblack

### DIFF
--- a/cppcache/test/PdxInstanceImplTest.cpp
+++ b/cppcache/test/PdxInstanceImplTest.cpp
@@ -55,13 +55,14 @@ TEST(PdxInstanceImplTest, updatePdxStream) {
   auto cache = cacheFactory.create();
   CacheImpl cacheImpl(&cache, properties, true, false, nullptr);
   auto buffer = std::vector<uint8_t>(__1M__, 0xcc);
+  auto len = static_cast<int32_t>(buffer.size());
   PdxInstanceImpl pdxInstanceImpl(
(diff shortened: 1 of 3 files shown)


moleske (approved):
Looks good!  Had to read the [story](https://issues.apache.org/jira/browse/GEODE-7299) to get the gist of the issue.  Appreciate the new test!

cppcache/test/PdxInstanceImplTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+      *(cacheImpl.getPdxTypeRegistry()), cacheImpl, false);
+
+  for (auto i = 0; i < __100K__; i++) {
+    try {
+      pdxInstanceImpl.updatePdxStream(buffer.data(), len);
+    } catch (const std::exception&) {
+      GTEST_FAIL();
+    }
+  }
+}
```
mmartell: I like this being a unit test!

mmartell (approved):
Nice fix to a serious memory leak. Thanks for switching to a standard container instead of using pointers.

moleske (approved):
still good! now with less variables to manage!

PR #538, opened 2019-10-17
GEODE-7303: Simplify native client UG builds

alb3rtobr:
- Now each user guide has its separated book directory, removing the current need of patching two files depending on the user guide being generated.
- The docker-based solution of GEODE-7272 has been adapted to each user guide, so they can be built just running a script.


davebarnes97:
+1 for the separate book directories
+1 for the docker files and the build/view scripts
+1 for the manual-build directory - those of us who are accustomed to a more interactive process will be glad to have this.

Nice work!

PR #542, opened 2019-10-23
GEODE-7347: Guarding CacheImpl during Cache move to avoid race condition

mreddington:


moleske:
@steve-sienk @vfordpivotal

cppcache/include/geode/Cache.hpp
```diff
@@ -254,7 +254,7 @@ class APACHE_GEODE_EXPORT Cache : public GeodeCache {
         bool readPdxSerialized,
         const std::shared_ptr<AuthInitialize>& authInitialize);
 
-  std::shared_ptr<CacheImpl> m_cacheImpl;
+  std::unique_ptr<CacheImpl> m_cacheImpl;
```
mmartell: This feels much better, avoiding any ambiguity about who's responsible for a given cache. Why did it take us so long.

PR #543, opened 2019-10-26
Some cleanup in ThinClientPoolDM

moleske:
LGTM pointed out a few things, tried to fix them.  Also cleaned up some thing CLion pointed out to me.  Hopefully this doesn't add new things to LGTM

Will make a jira later if this turns out alright, this was mostly stuff I could easily pick up/put down while waiting for some other builds

### DIFF
--- a/cppcache/src/ThinClientPoolDM.cpp
+++ b/cppcache/src/ThinClientPoolDM.cpp
@@ -387,9 +387,7 @@ void ThinClientPoolDM::manageConnections(std::atomic<bool>& isRunning) {
     m_connSema.acquire();
     if (isRunning) {
       manageConnectionsInternal(isRunning);
-      while (m_connSema.tryacquire() != -1) {
-        // ignored
-      }
+      m_connSema.acquire();
     }
   }
   LOGFINE("ThinClientPoolDM: ending manageConnections thread");
@@ -2059,9 +2057,7 @@ void ThinClientPoolDM::pingServer(std::atomic<bool>& isRunning) {
     m_pingSema.acquire();
     if (isRunning && !m_connManager.isNetDown()) {
       pingServerLocal();
-      while (m_pingSema.tryacquire() != -1) {
-        // ignored
-      }
+      m_pingSema.acquire();
     }
   }
   LOGFINE("Ending ping thread for pool %s", m_poolName.c_str());
@@ -2078,9 +2074,7 @@ void ThinClientPoolDM::cliCallback(std::atomic<bool>& isRunning) {
           *(m_connManager.getCacheImpl()->getCache()));
       // this call for cpp client
       m_connManager.getCacheImpl()->getPdxTypeRegistry()->clear();
-      while (m_cliCallbackSema.tryacquire() != -1) {
-        // ignored
-      }
+      m_cliCallbackSema.acquire();
     }
   }
   LOGFINE("Ending cliCallback thread for pool %s", m_poolName.c_str());
--- a/cppcache/src/ThinClientPoolDM.cpp
+++ b/cppcache/src/ThinClientPoolDM.cpp
@@ -220,7 +220,6 @@ ThinClientPoolDM::ThinClientPoolDM(const char* name,
 void ThinClientPoolDM::init() {
   LOGDEBUG("ThinClientPoolDM::init: Starting pool initialization");
   auto cacheImpl = m_connManager.getCacheImpl();
-  auto& sysProp = cacheImpl->getDistributedSystem().getSystemProperties();
   m_isMultiUserMode = getMultiuserAuthentication();
 
   if (m_isMultiUserMode) {
@@ -241,9 +240,9 @@ void ThinClientPoolDM::init() {
 
 ThinClientPoolDM::~ThinClientPoolDM() {
   destroy();
-  _GEODE_SAFE_DELETE(m_locHelper);
-  _GEODE_SAFE_DELETE(m_stats);
-  _GEODE_SAFE_DELETE(m_manager);
+  _GEODE_SAFE_DELETE(m_locHelper)
+  _GEODE_SAFE_DELETE(m_stats)
+  _GEODE_SAFE_DELETE(m_manager)
 }
 
 std::shared_ptr<Properties> ThinClientPoolDM::getCredentials(TcrEndpoint* ep) {
@@ -392,7 +391,7 @@ void ThinClientPoolDM::manageConnections(std::atomic<bool>& isRunning) {
     if (isRunning) {
       manageConnectionsInternal(isRunning);
       while (m_connSema.tryacquire() != -1) {
-        ;
+        // ignored
       }
     }
   }
@@ -409,8 +408,6 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
   auto _idle = getIdleTimeout();
   auto _nextIdle = _idle;
 
-  TcrConnection* conn = nullptr;
-
   std::vector<TcrConnection*> removelist;
   std::set<ServerLocation> excludeServers;
 
@@ -418,6 +415,7 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
   auto savedConns = 0;
 
   for (unsigned int i = 0; (i < availableConns) && isRunning; i++) {
+    TcrConnection* conn = nullptr;
     conn = getNoWait();
     if (conn == nullptr) {
       break;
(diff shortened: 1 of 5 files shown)


cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -241,9 +237,9 @@ void ThinClientPoolDM::init() {
 
 ThinClientPoolDM::~ThinClientPoolDM() {
   destroy();
-  _GEODE_SAFE_DELETE(m_locHelper);
-  _GEODE_SAFE_DELETE(m_stats);
-  _GEODE_SAFE_DELETE(m_manager);
+  _GEODE_SAFE_DELETE(m_locHelper)
+  _GEODE_SAFE_DELETE(m_stats)
+  _GEODE_SAFE_DELETE(m_manager)
```
dgkimura: Admittedly, I don't know the current best-practice in C++, but I think this looks a little meh and leads me think the macro is wrong.  If the spurious semicolon bothers you, then I would change the macro to...
```
#define _GEODE_SAFE_DELETE(x) \
 do  {                        \
    delete x;                 \
    x = nullptr;              \
  } while (0)
```

moleske: this was default settings in CLion, and me hitting `option+enter` as fast I could, so didn't look that closely at the change.  I can change it back pretty easily (and don't have strong feelings about it).  Strangely enough CLion considered it error with the note `Empty expression statement has no effect; remove unncessary ';' to silence this warning` 🤷‍♀️

dgkimura: I agree that an empty semicolon is silly.  Again though, as someone who has been heavily tainted by C, I'd suggest altering the macro to be do/while(0) rather than deleting the semicolon at all the call sites.  Besides the aesthetic appeal of seeing statements terminated by semicolon, it would reduce the diff if a future change moved the logic into a function by the same name.

moleske: sound reasonable to me!

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -392,7 +388,7 @@ void ThinClientPoolDM::manageConnections(std::atomic<bool>& isRunning) {
     if (isRunning) {
       manageConnectionsInternal(isRunning);
       while (m_connSema.tryacquire() != -1) {
-        ;
+        // ignored
```
pivotal-jbarrett: Since this just creates a hot loop, why not just call blocking version of `acquire`?

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -409,16 +405,14 @@ void ThinClientPoolDM::cleanStaleConnections(std::atomic<bool>& isRunning) {
...
-
   std::vector<TcrConnection*> removelist;
   std::set<ServerLocation> excludeServers;
 
   auto availableConns = size();
   auto savedConns = 0;
 
   for (unsigned int i = 0; (i < availableConns) && isRunning; i++) {
-    conn = getNoWait();
+    TcrConnection* conn = getNoWait();
```
pivotal-jbarrett: `auto`

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -593,8 +587,8 @@ std::string ThinClientPoolDM::selectEndpoint(
                   outEndpoint.getPort());
     LOGFINE("ThinClientPoolDM: Locator returned endpoint [%s]", epNameStr);
     return epNameStr;
-  } else if (m_attrs->m_initServList
-                 .size()) {  // use specified server endpoints
+  } else if (!m_attrs->m_initServList
+                  .empty()) {  // use specified server endpoints
```
pivotal-jbarrett: Move the comment so that the `.empty()` isn't line split.

cppcache/src/ThinClientPoolDM.cpp
pivotal-jbarrett: Same issue found in a few other files.

moleske: Some day I'll learn to read my changes properly instead of blindly fixing warnings!

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -2488,9 +2423,8 @@ GfErrType FunctionExecution::execute(void) {
           m_poolDM->getConnectionManager().getCacheImpl()->createDataOutput()),
       funcName, m_args, m_getResult, m_poolDM, m_timeout);
   TcrMessageReply reply(true, m_poolDM);
-  ChunkedFunctionExecutionResponse* resultProcessor(
-      new ChunkedFunctionExecutionResponse(reply, (m_getResult & 2) == 2, *m_rc,
-                                           m_resultCollectorLock));
+  auto* resultProcessor(new ChunkedFunctionExecutionResponse(
+      reply, (m_getResult & 2) == 2, *m_rc, m_resultCollectorLock));
```
dgkimura: Admittedly this doesn't have much to do with your change, but any idea why resultProcessor is allocated on the heap?  Also, feel free to ignore as it doesn't pertain to your commit.

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -115,7 +113,7 @@ class GetAllWork : public PooledWork<GfErrType>,
   TcrMessage* getReply() { return m_reply; }
 
   void init() {}
-  GfErrType execute(void) {
+  GfErrType execute() {
```
dgkimura: Curious what was the warning/error message here?  At least in C, it is legal to declare a function like so

```c
void f()
{
}
```

And then call it like so...
```c
f(1, 2, 3);
```

Explicitly calling out void in the function arguments prevents that.  Perhaps C++ takes away the foot gun..

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -1077,18 +1039,16 @@ void ThinClientPoolDM::AddEnum(std::shared_ptr<Serializable> enumInfo,
...
       enumInfo, this, enumVal);
 
   TcrMessageReply reply(true, this);
 
-  err = sendSyncRequest(request, reply);
+  auto err = sendSyncRequest(request, reply);
 
   if (err != GF_NOERR) {
-    throwExceptionIfError("Operation Failed", err);
+    throwExceptionIfError("Operation Failed", err)
```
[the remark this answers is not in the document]
moleske: lol, I caught this one right before you reviewed, just not fast enough to push

cppcache/src/ThinClientPoolDM.cpp
moleske: Clang-tidy in CLion said it was `Redundant void argument list in function definition`.  My understanding was C++ didn't allow the C example you shared.  I didn't clean up the header file, it still has `(void)` with the `Redundant void argument` message

dgkimura: For a second I was confused thinking this _is_ the function definition and declaration, as it's where the class is defined.  Now I see that it inherits from another class and understand why void is redundant here.  I think `override` would help make this more obvious and I may have realized.

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -989,15 +957,13 @@ void ThinClientPoolDM::AddPdxType(std::shared_ptr<Serializable> pdxType,
...
-  GfErrType err = GF_NOERR;
-
   TcrMessageGetPdxTypeById request(
       new DataOutput(m_connManager.getCacheImpl()->createDataOutput()), typeId,
       this);
 
   TcrMessageReply reply(true, this);
 
-  err = sendSyncRequest(request, reply);
+  auto err = sendSyncRequest(request, reply);
```
pdxcodemonkey: How about we get rid of `err` here altogether?  Checking it against GF_NOERR then doing `throwExceptionIfError` when we already know it's an error seems redundant...

```
throwExceptionIfError("Operation Failed", sendSyncRequest(request, reply));
```

cppcache/src/ThinClientPoolDM.cpp
pivotal-jbarrett: There are several things wrong with this line.
1) `auto*` really isn't a thing. 
2) a `new` doesn't need to be "copied" into a pointer.
3) Since the `TcrChunkedResult` doesn't really survive outside the life of this method or the reply object it should probably be a `std::shared_ptr` owned by both this method and the reply object or maybe even a `std::unique_ptr` owned by either the method or the reply object. Either way it removes the explicit deletes in multiple places in this method and nothing can leak.

Simple fix:
```C++
  auto resultProcessor = std::unique_ptr<ChunkedFunctionExecutionResponse>(
      new ChunkedFunctionExecutionResponse(reply, (m_getResult & 2) == 2, *m_rc,
                                           m_resultCollectorLock));
  reply.setChunkedResultHandler(resultProcessor.get());
```
and delete all:
```C++
  delete resultProcessor;
  resultProcessor = nullptr;
```

cppcache/src/ThinClientPoolDM.cpp
moleske: For closure, there was a warning in the editor for missing override that I didn't notice while trying to resolve the ton of other messages.  Proves that a 4th pass at the warnings would've helped

PR #544, opened 2019-11-07
GEODE-7419: Complete info about geode-native integration tests

alb3rtobr:


CONTRIBUTING.md
```diff
@@ -6,12 +6,13 @@ see [BUILDING.md](BUILDING.md)
...
-* Test your feature branch changes
-* Check your formatting
-* Submit a pull request
+* Test your feature branch changes.
+* Check your formatting.
+* Submit a pull request.
 
 ## Testing
-Before submitting a pull request the unit and integration tests must all pass. We are using CTest, (Please see [the CTest documentation](https://cmake.org/Wiki/CMake/Testing_With_CTest) for further information.)
+Before submitting a pull request the unit and integration tests must all pass. We are using CTest (please see [the CTest documentation](https://cmake.org/Wiki/CMake/Testing_With_CTest) for further information) for unit tests and old integration tests, and [Google Test](https://github.com/google/googletest) for new integration tests.
```
pivotal-jbarrett: We actually only use CTest to run tests. CTest itself is not a testing framework but a runtime management thing. All tests, including new and old use CTest for running. You can however run the unit and new integration tests, which are written in gtest, directly, like from your IDE. Running the old unit tests, which are written in a home grown test framework similar to the main java project's DUnit, is a lot more complicated from the IDE.

CONTRIBUTING.md
```diff
@@ -20,14 +21,19 @@ $ cd cppcache/test/<Debug|Release|if needed>
 $ ./apache-geode_unittests
 ```
 
-### Running old integration tests
+### Running integration tests
+There are two sets of integration tests based on different testing frameworks: CTest and [Google Test](https://github.com/google/googletest).
+
+CTest-based integration tests are deprecated. If your changes include the implementation of new integration test/s to be verified, they should be written using Google Test. If your change implies a significant change in one or more CTest-based test cases, you should create the equivalent test case/s using Google Test to substitute the old one/s instead of adapting the old ones.
```
pivotal-jbarrett: Again the deprecated old integration tests are written in a home grown framework. CTest is just a mechanism for managing and running tests in CMake. The new integration tests framework is home grown extension on top fo gtest. The unit tests are straight up gtest.

CONTRIBUTING.md
```diff
@@ -20,14 +21,19 @@ $ cd cppcache/test/<Debug|Release|if needed>
...
+
+#### Running CTest integration tests (deprecated)
 ```bash
 $ cd <clone>
 $ cd build
 $ cmake --build . --target run-integration-tests
```
pivotal-jbarrett: This should actually be running both the old and the new integration tests. If not it's a bug. ;)

pivotal-jbarrett:
@alb3rtobr You really don't need to worry about JIRAs to modify the md files in the future too.

pivotal-jbarrett:
@alb3rtobr here is an example of running the new integration tests using CTest to manage the execution. CTest allows us to parallelize and re-run failed tests and integrates into the CMake ecosystem.

```
✘-INT ~/Develop/geode-native/build/debug/cppcache/integration/test [develop|⚑ 5] 
11:25 $ ctest -R . -j 4
Test project /Users/jbarrett/Develop/geode-native/build/debug/cppcache/integration/test
      Start  1: AuthInitializeTest.putGetWithBasicAuth
      Start  2: AuthInitializeTest.putWithBadUsername
      Start  3: AuthInitializeTest.putWithBadPassword
      Start  4: AuthInitializeTest.badCredentialsWithSubscriptionEnabled
...
```

alb3rtobr:
@pivotal-jbarrett thanks for the example, it was really useful for me. I think with my new commit the current situation of the tests is correct now.

CONTRIBUTING.md
alb3rtobr: I have been testing this command. I changed the code to run just one of the 110 old test cases. I was expecting that after that test case, the new ones were executed, but it did not happen:

```
$ cmake --build . --target run-integration-tests
Test project /home/alb3rtobr/git/Nordix/geode-native/build/cppcache/integration-test
    Start 1: testCacheless
1/1 Test #1: testCacheless ....................   Passed   69.02 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
STABLE    =  69.02 sec*proc (1 test)

Total Test time (real) =  69.03 sec
[  0%] Built target run-stable-cppcache-integration-tests
[  0%] Built target version-header
[  2%] Built target ACE-extern
[  6%] Built target boost-extern
[ 10%] Built target xerces-c-extern
[ 80%] Built target apache-geode
[ 84%] Built target framework
[ 92%] Built target testobject
[ 92%] Built target test-cppcache-utils
[ 92%] Built target cryptoImpl
[ 94%] Built target DHImpl
[ 98%] Built target sqlite-extern
[ 98%] Built target SqLiteImpl
[ 98%] Built target securityImpl
[100%] Built target testCacheless
Scanning dependencies of target integration-tests
[100%] Built target integration-tests
Scanning dependencies of target run-integration-tests
[100%] Built target run-integration-tests

```

alb3rtobr:
Taking into account that the `cmake` command in the instructions is not valid for running both integration test suites, I think its more consistent to include only `ctest` command to execute both test suites. I have included also examples of running standalone test cases.

pivotal-jbarrett:
> Taking into account that the `cmake` command in the instructions is not valid for running both integration test suites, I think its more consistent to include only `ctest` command to execute both test suites. I have included also examples of running standalone test cases.

I agree as well. I was never really a fan of those targets anyway. I don't even think the CI uses those targets. My vote would be to purge those targets from the CMake and document the `ctest` command line, minimally, and link to CTest documentation for more advanced features.

You may want to mention that the new unit and integration tests based on gtest should be automatically picked up by any IDE that supports gtest, like CLion, and can be executed and debugged directly in the IDE.

CONTRIBUTING.md
```diff
@@ -39,16 +39,24 @@ This will take ~ 2 hours, YMMV... you can up the jobs to 4 and run in parallel,
...
 
-### Running new integration tests
+#### Running Google Test integration test suite
 ```bash
 $ cd <clone>
-$ cd build
-$ cd cppcache/integration/test
-$ ./cpp-integration-test [<options>]
+$ cd build/cppcache/integration/test
+$ ctest -R . -j1
```
pivotal-jbarrett: Depending on your host, you should be safe to run a few in parallel. Also `-R .` is redundant to no `-R` argument at all. I am not sure where the `-R .` stated sneaking into this document.

PR #546, opened 2019-11-08
GEODE-7418: Fixes issue deserializing unregistered PDX types.

pivotal-jbarrett:


cppcache/integration/test/PdxSerializerTest.cpp
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  {
+    auto cache = cluster.createCache();
+    auto region = setupRegion(cache);
+    cache.getTypeRegistry().registerPdxSerializer(
+        std::make_shared<TestPdxSerializer>());
+
+    assertNonPdxType(nonPdxType, region->get("2"));
+  }
```
moleske: This test seems duplicate to me.  Is there something I'm missing, like the need to test with something that was already `put` in once before?

pivotal-jbarrett: Its doing two things. In the first block, with client A, its creating, inserting and getting the object. It is getting the object to assert that the value is properly serialized prior to the real test. The second block, client B tries to get the object without knowing anything about the type and asserts it is successful.

If there is a different structure you would like for this test please go ahead. In creating this framework I never found a way that I was completely satisfied to clearly identify two or more tracks of work on independent caches. The old framework separated them by macros and processes and it was super complicated to debug and read. I want to avoid that with this new framework.

PR #548, opened 2019-11-13
GEODE-7447: Fix logging format spec <--> variable size mismatches

pdxcodemonkey:
- std::chrono 'Rep' type changes size based on platform
- when size mismatched to format spec in log call, bad things happen


pdxcodemonkey:
@charliemblack @mreddington @dihardman @vford-pivotal

pdxcodemonkey:
This has issues, closing PR for now while I diagnose

PR #552, opened 2019-11-18
GEODE-7447: Fix test breakage on Windows

pdxcodemonkey:
- std::chrono for things like system time on Windows return a
non-standard ratio corresponding to 'ticks' (100ns units)
- This unit was blowing up our string conversion code, which defaulted
to nullptr for string value of unrecognized units

Fix is threefold, for redundancy:
- Normalize the units in the place where the tests were choking on it, since the base value was in seconds to begin with
- Don't return nullptr in the default case, to avoid the crash in other places we may have missed
- Add "ticks" units for Windows, which in obscure cases may help us distinguish between this specific problem and a units problem we may hit on another platform

@mreddington @dihardman @davebarnes97 @vfordpivotal

### DIFF
--- a/cppcache/include/geode/internal/chrono/duration.hpp
+++ b/cppcache/include/geode/internal/chrono/duration.hpp
@@ -57,12 +57,6 @@ template <>
 struct _suffix<std::micro> {
   static constexpr char const* value = "us";
 };
-#ifdef WIN32
-template <>
-struct _suffix<std::ratio<10000000>> {
-  static constexpr char const* value = "ticks";
-};
-#endif
 template <>
 struct _suffix<std::nano> {
   static constexpr char const* value = "ns";
--- a/cppcache/include/geode/internal/chrono/duration.hpp
+++ b/cppcache/include/geode/internal/chrono/duration.hpp
@@ -35,7 +35,7 @@ namespace duration {
 
 template <class Period>
 struct _suffix {
-  static constexpr char const* value = nullptr;
+  static constexpr char const* value = "<<unknown units>>";
 };
 template <>
 struct _suffix<std::ratio<3600>> {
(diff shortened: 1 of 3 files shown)


cppcache/include/geode/internal/chrono/duration.hpp
```diff
@@ -57,6 +57,12 @@ template <>
 struct _suffix<std::micro> {
   static constexpr char const* value = "us";
 };
+#ifdef WIN32
```
pivotal-jbarrett: No! Let's please leave ticks out and stick with SI units. Please for the love all all things holy!

cppcache/include/geode/internal/chrono/duration.hpp
```diff
@@ -35,7 +35,7 @@ namespace duration {
 
 template <class Period>
 struct _suffix {
-  static constexpr char const* value = nullptr;
+  static constexpr char const* value = "<<unknown units>>";
```
moleske: little weird that we can't know the units, but assuming this is the best we can do for now

pdxcodemonkey: This is the default, anything that comes in using a ratio value we understand gets a "real" suffix.  The point of the change is to stop using nullptr, because "bad" things happen :)

cppcache/src/EntryExpiryHandler.cpp
```diff
@@ -56,25 +56,19 @@ int EntryExpiryHandler::handle_timeout(const ACE_Time_Value& current_time,
     }
 
     auto elapsed = curr_time - lastTimeForExp;
-    LOGDEBUG(
-        "Entered entry expiry task handler for key [%s] of region [%s]: "
-        "%s,%s,%s,%s",
-        Utils::nullSafeToString(key).c_str(),
-        m_regionPtr->getFullPath().c_str(),
-        to_string(curr_time.time_since_epoch()).c_str(),
```
pivotal-jbarrett: Removing the time components from this log message really makes the log message useless. Either fix the issue of printing the time on Windows or delete the log message entirely.

pdxcodemonkey: Okay, will remove.  FWIW, it is/was working fine on Windows, the issue was that after I applied that fix it started breaking a test on Linux, and I couldn't figure out why in a reasonable amount of time.

PR #556, opened 2019-11-28
GEODE-7509: Fix memory leaks in C++ client

albertogpz:
This commit solves some more memory leaks
and uninitialized memory accesses found
in the C++ client when running the integration
tests that were not solved by GEODE-7476

### DIFF
--- a/cppcache/src/PdxInstanceImpl.cpp
+++ b/cppcache/src/PdxInstanceImpl.cpp
@@ -1286,7 +1286,7 @@ std::shared_ptr<CacheableStringArray> PdxInstanceImpl::getFieldNames() {
   auto pt = getPdxType();
   std::vector<std::shared_ptr<PdxFieldType>>* vectorOfFieldTypes =
       pt->getPdxFieldTypes();
-  size_t size = vectorOfFieldTypes->size();
+  auto size = vectorOfFieldTypes->size();
   if (size == 0) {
     return nullptr;
   }
--- a/cppcache/src/PdxInstanceImpl.cpp
+++ b/cppcache/src/PdxInstanceImpl.cpp
@@ -1286,21 +1286,17 @@ std::shared_ptr<CacheableStringArray> PdxInstanceImpl::getFieldNames() {
   auto pt = getPdxType();
   std::vector<std::shared_ptr<PdxFieldType>>* vectorOfFieldTypes =
       pt->getPdxFieldTypes();
-  int size = static_cast<int>(vectorOfFieldTypes->size());
-  std::shared_ptr<CacheableString>* ptrArr = nullptr;
-  if (size > 0) {
-    ptrArr = new std::shared_ptr<CacheableString>[size];
-    for (int i = 0; i < size; i++) {
-      ptrArr[i] =
-          CacheableString::create((vectorOfFieldTypes->at(i))->getFieldName());
-    }
+  size_t size = vectorOfFieldTypes->size();
+  if (size == 0) {
+    return nullptr;
   }
-
-  if (size > 0) {
-    return CacheableStringArray::create(
-        std::vector<std::shared_ptr<CacheableString>>(ptrArr, ptrArr + size));
+  std::vector<std::shared_ptr<CacheableString>> tmpFieldNames;
+  tmpFieldNames.reserve(size);
+  for (auto&& fieldType : *vectorOfFieldTypes) {
+    tmpFieldNames.emplace_back(
+        CacheableString::create(fieldType->getFieldName()));
   }
-  return nullptr;
+  return CacheableStringArray::create(std::move(tmpFieldNames));
 }
 
 PdxFieldTypes PdxInstanceImpl::getFieldType(
@@ -1920,17 +1916,13 @@ void PdxInstanceImpl::setField(const std::string& fieldName, std::string* value,
                                 " or type of field not matched " +
                                 (pft != nullptr ? pft->toString() : ""));
   }
-  std::shared_ptr<CacheableString>* ptrArr = nullptr;
   if (length > 0) {
-    ptrArr = new std::shared_ptr<CacheableString>[length];
+    std::vector<std::shared_ptr<CacheableString>> tmpValues;
+    tmpValues.reserve(length);
     for (int32_t i = 0; i < length; ++i) {
-      ptrArr[i] = CacheableString::create(value[i]);
+      tmpValues.emplace_back(CacheableString::create(value[i]));
     }
-  }
-  if (length > 0) {
-    auto cacheableObject = CacheableStringArray::create(
-        std::vector<std::shared_ptr<CacheableString>>(ptrArr, ptrArr + length));
-    m_updatedFields[fieldName] = cacheableObject;
+    m_updatedFields[fieldName] = CacheableStringArray::create(tmpValues);
   }
 }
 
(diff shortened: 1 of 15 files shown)


cppcache/src/PdxInstanceImpl.cpp
```diff
@@ -1286,21 +1286,17 @@ std::shared_ptr<CacheableStringArray> PdxInstanceImpl::getFieldNames() {
...
       pt->getPdxFieldTypes();
-  int size = static_cast<int>(vectorOfFieldTypes->size());
-  std::shared_ptr<CacheableString>* ptrArr = nullptr;
-  if (size > 0) {
-    ptrArr = new std::shared_ptr<CacheableString>[size];
-    for (int i = 0; i < size; i++) {
-      ptrArr[i] =
-          CacheableString::create((vectorOfFieldTypes->at(i))->getFieldName());
-    }
+  size_t size = vectorOfFieldTypes->size();
```
pivotal-jbarrett: `auto` please

cppcache/src/TcrMessage.cpp
```diff
@@ -479,16 +479,15 @@ void TcrMessage::readLongPart(DataInput& input, uint64_t* intValue) {
 }
 
 const std::string TcrMessage::readStringPart(DataInput& input) {
-  char* stringBuffer;
   int32_t stringLength = input.readInt32();
-  stringBuffer = new char[stringLength + 1];
-  stringBuffer[stringLength] = '\0';
+  auto stringBuffer = std::unique_ptr<char[]>(new char[stringLength + 1]);
```
pivotal-jbarrett: Could we use `std::vector`?

tests/cpp/fwklib/UDPIpc.hpp
```diff
@@ -160,7 +160,11 @@ class UDPMessageQueues : public SharedTaskObject {
   std::string m_label;
 
  public:
-  explicit UDPMessageQueues(std::string label) : m_label(label) {}
+  explicit UDPMessageQueues(std::string label) : m_label(label) {
+    m_cntInbound = 0;
```
pivotal-jbarrett: Can we be consistent and either use the initializer style of constructor body but not both.

cppcache/src/PdxInstanceImpl.cpp
albertogpz: Right

cppcache/src/TcrMessage.cpp
albertogpz: Interesting...

tests/cpp/fwklib/UDPIpc.hpp
albertogpz: Sure, sorry.

cppcache/src/TcrMessage.cpp
```diff
@@ -480,14 +480,14 @@ void TcrMessage::readLongPart(DataInput& input, uint64_t* intValue) {
...
   int32_t stringLength = input.readInt32();
-  auto stringBuffer = std::unique_ptr<char[]>(new char[stringLength + 1]);
-  stringBuffer.get()[stringLength] = '\0';
+  std::vector<char> stringBuffer(stringLength + 1);
+  stringBuffer[stringLength] = '\0';
   if (input.read()) {
     throw Exception("String is not an object");
   }
-  input.readBytesOnly(reinterpret_cast<int8_t*>(stringBuffer.get()),
+  input.readBytesOnly(reinterpret_cast<int8_t*>(stringBuffer.data()),
```
pivotal-jbarrett: This isn't your problem but maybe you want to tackle it.
I dug a little deeper in here. It felt wrong that we were just reading the raw bytes into the string. As it turns out my gut was correct. The server either sends an ASCII only string (by setting of global flag that defaults false) or Java Modified UTF-8 (see `Message.addStringPart`). Since Java Modified UTF-8 isn't a standard we have special code on the C++ client to transcode it to UTF-8. This block of code should really be invoking a call to read Java Modified UTF-8. It would be great if we could just invoke it right on `DataInput` but the method that can read JMUTF-8 is private and expects to read the length too but when encode in `Part` the length is in the part header not the body. I think the solution is to just read the JMUTF directly using `JavaModifiedUtf8::decode`, which gives you the `std::u16string` (easy transcode), then transcode that to `std::string` in UTF8 using `to_utf8` in util/string.hpp.

pivotal-jbarrett: In case it isn't clear, reading the JMUTF-8 will safely take care of the ASCII cases since ASCII is a 1:1 encoding into JMUTF-8 and UTF-8.

albertogpz: I tried to fix the bug following your directions.
I did not write a test because I did not fully understand the problem so I'd appreciate some help to write a proper test.

albertogpz:
I forgot to ask something that did not really feel right on one of the fixes.
I noticed I had to add a `delete this` sentence in the `handle_close()` method of the different ExpiryHandler classes in order to remove the leaking of the memory of the handlers in spite the fact there was a comment stating that the handler would be deleted now in `GF_Timer_Heap_ImmediateReset_T`. It seems that the handler was not deleted by that class in all cases.
Apart from that, the `TombstoneExpiryHandler.handle_close()` method did not require such delete sentence because these handlers were deleted in `TombstoneList::cleanUp()`.
If you have a suggestion to make this code clear I would be willing to improve it.

pivotal-jbarrett:
> If you have a suggestion to make this code more clear I would be willing to improve it.

The overall plan is to remove ACE library so these handlers will go away with that. I would just put in the fix as is for now.

cppcache/src/TcrMessage.cpp
pivotal-jbarrett: I doubt there is an existing test that covers this. Given that we are talking to the server here we would need an integration test. I think if you create a test that creates a cq with non-ASCII characters and you get some update events for the CQ, then you should pass through this code. Maybe you can modify an existing CQ test and port it to the new integration framework.

albertogpz: There are CQ tests both in the old and new frameworks that when executed, pass through the updated code.
I modified both tests to use CQs with names that contained non-ASCII characters and ran them with and without the changes in readStringPart but saw no difference in the results.
It looks as if the client behaves in the same way with the changes.
Am I missing something?

pivotal-jbarrett: Doh! My bad. just being non-ASCII isn’t enough. The fix is correct and the old code would have worked too For most values. The server is sending JMUTF-8 which looks a lot like UTF-8 except around NUL and characters outside the basic plane. The name would have to have a NUL (u0x00) character Or very high Unicode character in it before the old code would have produced invalid UTF-8. I’ll poke at one of the tests to see if I can make it fail and get back to you.

pivotal-jbarrett: Ok, I can reproduce the failure with string literal `u8"SimpleCQ\xF0\x90\x90\x80"` as the CQ name. This UTF-8 sequence produces the U+10400 character "𐐀", which has two UTF-16 code points and results in 6 bytes of JMUTF-8 rather then 4 bytes of UTF-8.

I also found a race condition in the new CqTest but I will write that up separately.

cppcache/src/TcrMessage.cpp
```diff
@@ -480,15 +480,14 @@ void TcrMessage::readLongPart(DataInput& input, uint64_t* intValue) {
 
 const std::string TcrMessage::readStringPart(DataInput& input) {
   int32_t stringLength = input.readInt32();
```
pivotal-jbarrett: Also, you could `auto` this for completeness.

cppcache/src/TcrMessage.cpp
albertogpz: Great! I was not able to reproduce it with a chinese character.

PR #558, opened 2019-12-02
GEODE-7520: Add convenient target to clangformat everything

pdxcodemonkey:
- Will save time on developer builds
- Stops breaking Windows builds when Visual Studio clang-format crashes

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -151,6 +151,23 @@ To override `clang-tidy` options:
 ```console
 $ cmake … -DCMAKE_CXX_CLANG_TIDY=clang-tidy;<options> …
 ```
+#### Clang-format
+Individual targets in the build tree have their own dependency of the form `<<targetName>>_clangformat`, which uses the `clang-format` executable, wherever it is found, to format and modified files according to the rules specfied in the .clang-format file.  This is helpful when submitting changes to geode-native, because an improperly formatted file will fail Travis-CI and have to be fixed prior to merging any pull request.  If clang-format is not installed on your system, clangformat targets will not be added to your project files, and geode-native should build normally.  Under some circumstances, however, it may become necessary to disable `clang-format` on a system where it _is_ installed.
+
+To disable `clang-format` in the build:
+
+```
+$ cmake … -DClangFormat_EXECUTABLE='' …
+```
+
+On the other hand, it may also be desirable to run clang-format on the entire source tree.  This is also easily done via the `all-clangformat` _in a build with clang-format enabled_.  If clang-format has been disabled in the cmake configuration step, as above, the `all-clangformat` target will not exist, and the cmake configuration step will have to be re-run with clang-format enabled.
+
+To run clang-format on the entire source tree:
+
+```
+$ cmake --build . --target all-clangformat
+```
+
 
 ## Installing
 By default a system-specific location is used by CMake as the destination of the `install` target, e.g., `/usr/local` on UNIX system. To explicitly specify the location in which the Native Client will be installed, add `-DCMAKE_INSTALL_PREFIX=/path/to/installation/destination` to the _initial_ `cmake` execution command.
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -152,7 +152,7 @@ To override `clang-tidy` options:
 $ cmake … -DCMAKE_CXX_CLANG_TIDY=clang-tidy;<options> …
 ```
 #### Clang-format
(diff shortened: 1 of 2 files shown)


pdxcodemonkey:
I respectfully disagree:
* Just as a ballpark estimate, developers on the team spend several minutes of build time per day running clang-format on code.  This doesn't strictly need to be done until it's time to push.
* In point of fact, formatting is *not* enforced in the build, or at least not completely.  _Running_ the formatter is enforced, but committing the changes is not, and can't be without the use of a Git hook or the moral equivalent.  As a result, PRs fail in Travis CI due to formatting with some regularity.  This change doesn't materially change the situation w.r.t. enforcement.
* Formatting is not alone in this situation, either.  We don't enforce RAT checks or LGTM prior to submitting a PR, but errors will fail Travis and block merging.
* Having a single target to format the entire codebase is much simpler than adding a format dependency to arbitrarily many targets in the tree.  Running clang-format on everything in a CI job currently involves a grotesque single-line sed script piped through xargs, running `cmake --build . --target clangformat-all` is a huge improvement on this.
* Just philosophically, having an explicit target so that you can turn on or off is preferable to the implicit criteria "you don't get any clang-format targets if you don't have the clang-format executable on your system".  With the current setup, if someone new to the project doesn't have clang-format on their system, how are they to know that's a problem?  OTOH, if they attempt to build the clang-format target and get a "target not found" error, it's more obvious what's up.

I'd be open to a solution that allowed this target or targets to be on by default if they can a) be explicitly turned off via cmake variable and b) run for the entire code base without doing backflips on the command line.  Let me know what you think.

pivotal-jbarrett:
> * Just as a ballpark estimate, developers on the team spend several minutes of build time per day running clang-format on code.  This doesn't strictly need to be done until it's time to push.

Can you share you measurement scheme. When I measure it at worst its seconds.
No changes:
```
$ time make apache-geode-static-clangformat -j12
[100%] Built target apache-geode-static-clangformat

real    0m0.246s
user    0m0.174s
sys     0m0.054s
```
All changes:
```
$ time make apache-geode-static-clangformat -j12
[  0%] Clang-Format /Users/jbarrett/Develop/geode-native/cppcache/src/AuthenticatedView.cpp
...
[ 96%] Clang-Format /Users/jbarrett/Develop/geode-native/cppcache/include/geode/util/LogLevel.hpp
[100%] Built target apache-geode-static-clangformat

real    0m3.980s
user    0m22.907s
sys     0m12.501s
```

pivotal-jbarrett:
> * Formatting is not alone in this situation, either.  We don't enforce RAT checks or LGTM prior to submitting a PR, but errors will fail Travis and block merging.

RAT lacks incremental support. Since new files are rarely added we decided that running rat by default had little to no value. You can turn it on locally if you want.
`$ cmake ... -DUSE_RAT=on`

LGTM and other static analysis have extremely high costs in time so we chose to delay them until the PR process, though you can also run them locally if you want.
`$ cmake ... -DCMAKE_CXX_CLANG_TIDY=clang-tidy`

pivotal-jbarrett:
> I'd be open to a solution that allowed this target or targets to be on by default if they can a) be explicitly turned off via cmake variable and b) run for the entire code base without doing backflips on the command line. Let me know what you think.

`$ cmake ... -DClangFormat_EXECUTABLE=''`

pivotal-jbarrett:
> * Having a single target to format the entire codebase is much simpler than adding a format dependency to arbitrarily many targets in the tree.  Running clang-format on everything in a CI job currently involves a grotesque single-line sed script piped through xargs, running `cmake --build . --target clangformat-all` is a huge improvement on this.

As you have done its easy enough to add that target. Following the current convention I think that would be `all-clangformat`.

pdxcodemonkey:
`-DClangFormat_EXECUTABLE=''` served my purpose of allowing me to turn it off completely for a particular build, so I've added clangformat back to all the original targets.  I also kept the `all` clangformat target, and changed the name from `clangformat-all` to `all-clangformat`.  This should be good to go now.

BUILDING.md
```diff
@@ -151,6 +151,23 @@ To override `clang-tidy` options:
 ```console
 $ cmake … -DCMAKE_CXX_CLANG_TIDY=clang-tidy;<options> …
 ```
+#### Clang-format
+Individual targets in the build tree have their own dependency of the form `<<targetName>>_clangformat`, which uses the `clang-format` executable, wherever it is found, to format and modified files according to the rules specfied in the .clang-format file.  This is helpful when submitting changes to geode-native, because an improperly formatted file will fail Travis-CI and have to be fixed prior to merging any pull request.  If clang-format is not installed on your system, clangformat targets will not be added to your project files, and geode-native should build normally.  Under some circumstances, however, it may become necessary to disable `clang-format` on a system where it _is_ installed.
```
pivotal-jbarrett: `<<targetName>>-clangformat`

PR #559, opened 2019-12-03
GEODE-7528: Fixes race condition in tests.

pivotal-jbarrett:
* Replace counters with latch.


pivotal-jbarrett:
> Changes look good. Interesting to me that there would be multiple threads notifying a CQ. I should've thought that there would be only one. Was this a design decision in NC, or did things just work out that way?

The issue was the last operation in the loop was a destroy but the check was for creates to be complete, which was two loops earlier. Checking for destroys to be complete would have been safe, which is what the auth version seems to be doing. I like erring on the side of completeness and consistency so I implemented both with latches. Maybe overkill...

PR #560, opened 2019-12-04
GEODE-7549: Fix LGTM - Remove empty conditional

moleske:
-apply some clang tidy messages

### DIFF
--- a/cppcache/src/ClientProxyMembershipID.cpp
+++ b/cppcache/src/ClientProxyMembershipID.cpp
@@ -234,20 +234,17 @@ void ClientProxyMembershipID::fromData(DataInput& input) {
 }
 
 Serializable* ClientProxyMembershipID::readEssentialData(DataInput& input) {
-  uint8_t* hostAddress;
-  int32_t length, hostPort, vmViewId = 0;
-  std::shared_ptr<CacheableString> hostname, dsName, uniqueTag, vmViewIdstr;
-
-  length = input.readArrayLength();
-  hostAddress = new uint8_t[length];
+  auto length = input.readArrayLength();
+  auto hostAddress = new uint8_t[length];
   input.readBytesOnly(hostAddress, length);
-  hostPort = input.readInt32();
+  auto hostPort = input.readInt32();
 
   // read and ignore flag
   input.read();
 
   const auto vmKind = input.read();
-
+  int32_t vmViewId = 0;
+  std::shared_ptr<CacheableString> uniqueTag, vmViewIdstr;
   if (vmKind == ClientProxyMembershipID::LONER_DM_TYPE) {
     uniqueTag = std::dynamic_pointer_cast<CacheableString>(input.readObject());
   } else {
@@ -256,7 +253,7 @@ Serializable* ClientProxyMembershipID::readEssentialData(DataInput& input) {
     vmViewId = std::stoi(vmViewIdstr->value());
   }
 
-  dsName = std::dynamic_pointer_cast<CacheableString>(input.readObject());
+  auto dsName = std::dynamic_pointer_cast<CacheableString>(input.readObject());
 
   initHostAddressVector(hostAddress, length);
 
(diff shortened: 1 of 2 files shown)


cppcache/src/ClientProxyMembershipID.cpp
```diff
@@ -121,10 +122,9 @@ void ClientProxyMembershipID::initObjectVars(
   m_memID.write(static_cast<int8_t>(DSCode::InternalDistributedMember));
   m_memID.writeBytes(m_hostAddr.data(),
                      static_cast<int32_t>(m_hostAddr.size()));
-  // m_memID.writeInt((int32_t)hostPort);
   m_memID.writeInt(static_cast<int32_t>(synch_counter));
   m_memID.writeString(hostname);
-  m_memID.write(splitBrainFlag);  // splitbrain flags
+  m_memID.write(splitBrainFlag);
```
pivotal-jbarrett: hahah

cppcache/src/ClientProxyMembershipID.cpp
```diff
@@ -201,65 +193,60 @@ void ClientProxyMembershipID::toData(DataOutput&) const {
...
   }
-  delete[] hostAddr;
+  delete[] hostAddress;
   readAdditionalData(input);
 }
 
 Serializable* ClientProxyMembershipID::readEssentialData(DataInput& input) {
-  uint8_t* hostAddr;
-  int32_t len, hostPort, vmViewId = 0;
+  uint8_t* hostAddress;
```
pivotal-jbarrett: Can you move these closer to initialization and use `auto`?

cppcache/src/ClientProxyMembershipID.cpp
```diff
@@ -201,65 +193,60 @@ void ClientProxyMembershipID::toData(DataOutput&) const {
...
-  len = input.readArrayLength();  // inetaddress len
-  /* adongre - Coverity II
-   * CID 29183: Out-of-bounds access (OVERRUN_DYNAMIC)
-   */
-  hostAddr = new uint8_t[len];
-  input.readBytesOnly(hostAddr, len);  // inetaddress
-  hostPort = input.readInt32();        // port
-  // TODO: RVV get the host name from
+  length = input.readArrayLength();
+  hostAddress = new uint8_t[length];
```
pivotal-jbarrett: Can we use a `std::vector` to avoid the `delete` and potential leak?

cppcache/src/ClientProxyMembershipID.cpp
```diff
@@ -306,8 +293,7 @@ int16_t ClientProxyMembershipID::compareTo(
     return 0;
   }
 
-  const ClientProxyMembershipID& otherMember =
-      static_cast<const ClientProxyMembershipID&>(other);
+  const auto& otherMember = dynamic_cast<const ClientProxyMembershipID&>(other);
   uint32_t myPort = getHostPort();
```
pivotal-jbarrett: `auto`

cppcache/src/ClientProxyMembershipID.cpp
moleske: I got a little stuck.  I was thinking `hostdress = std::vector<uint8_t>(input.read());` but that's not quite it.  I think I'm confused because I'm not sure how to get data from `input` into a new vector.

pivotal-jbarrett: This works:
```c++
  auto length = input.readArrayLength();
  auto hostAddress = std::vector<uint8_t>(length);
  input.readBytesOnly(hostAddress.data(), hostAddress.size());
```

or, use the method that reads the array length and bytes in a single call.
```c++
  auto hostAddress = input.readByteArray();
```
Though, this is a `std::vector<int8_t>` (signed byte like Java) now so you will need a `static_cast<>` later for the call to `initHostAddressVector`.

But you will notice that `initHostAddressVector` just copies it to the `m_hostAddr` which is a `std::vector<uint8_t>`, so maybe there is some more optimizations to be made... Rabbit hole now...

moleske: for some reason `input.readBytesOnly(hostAddress.data(), hostAddress.size());` never dawned on me as an option 😞 
The latter one I started, but saw the same issue you mentioned

PR #564, opened 2019-12-19
GEODE-7601: Memory leaks in C++ native client found when running old integration-tests

albertogpz:
Several memory leaks and other issues in memory
management have been fixed in the native client.
The issues were found when running the old integration
tests under valgrind.

Many issues in the testing framework have also been fixed
to make it easier to identify and fix the problems
in the native client.

### DIFF
--- a/cppcache/include/geode/DataOutput.hpp
+++ b/cppcache/include/geode/DataOutput.hpp
@@ -497,8 +497,12 @@ class APACHE_GEODE_EXPORT DataOutput {
   static void acquireLock();
   static void releaseLock();
 
+  struct FreeDeleter {
+    void operator()(uint8_t* p) { free(p); }
+  };
+
   // memory m_buffer to encode to.
-  std::unique_ptr<uint8_t[]> m_bytes;
+  std::unique_ptr<uint8_t, FreeDeleter> m_bytes;
   // cursor.
   uint8_t* m_buf;
   // size of m_bytes.
--- a/cppcache/integration-test/ThinClientFailover3.hpp
+++ b/cppcache/integration-test/ThinClientFailover3.hpp
@@ -38,7 +38,7 @@ using apache::geode::client::CacheableString;
 using apache::geode::client::CacheHelper;
 
 bool isLocalServer = false;
-const char* endPoints = CacheHelper::getTcrEndpoints(isLocalServer, 3);
+const std::string endPoints = CacheHelper::getTcrEndpoints(isLocalServer, 3);
 CacheHelper* cacheHelper = nullptr;
 
 #define CLIENT1 s1p1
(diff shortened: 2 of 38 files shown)


cppcache/include/geode/DataOutput.hpp
```diff
@@ -497,8 +497,12 @@ class APACHE_GEODE_EXPORT DataOutput {
   static void acquireLock();
   static void releaseLock();
 
+  struct FreeDeleter {
+    void operator()(uint8_t* p) { free(p); }
+  };
+
   // memory m_buffer to encode to.
-  std::unique_ptr<uint8_t[]> m_bytes;
+  std::unique_ptr<uint8_t, FreeDeleter> m_bytes;
```
pivotal-jbarrett: What would it look like to convert this to a `std::vector`? Then we get all this for free.

cppcache/integration-test/ThinClientFailover3.hpp
```diff
@@ -242,21 +242,22 @@ void doNetsearch(const char* name, const char* key, const char* value) {
...
 }
 
-std::vector<char*> storeEndPoints(const char* points) {
-  std::vector<char*> endpointNames;
-  if (points != nullptr) {
-    char* ep = strdup(points);
+std::vector<std::string> storeEndPoints(const std::string points) {
+  std::vector<std::string> endpointNames;
+  if (!points.empty()) {
+    char* ep = strdup(points.c_str());
```
pivotal-jbarrett: Maybe we can keep this all as `std::string`?

cppcache/integration-test/fw_dunit.cpp
```diff
@@ -753,13 +756,22 @@ class TestSlave {
           DUNIT->setSlaveState(m_sId, SLAVE_STATE_TASK_ACTIVE);
           try {
             task->doTask();
+            if (task->m_isHeapAllocated) {
```
pivotal-jbarrett: Could we just make it a `std::shared_ptr` and let it do its magic?

cppcache/integration-test/testThinClientPartitionResolver.cpp
```diff
@@ -105,8 +105,8 @@ DUNIT_TASK_DEFINITION(CLIENT1, CreatePoolAndRegions)
     initClient(true);
 
     char endpoints[1024] = {0};
-    sprintf(endpoints, "%s,%s,%s", endpointNames.at(0), endpointNames.at(1),
-            endpointNames.at(2));
+    sprintf(endpoints, "%s,%s,%s", endpointNames.at(0).c_str(),
```
pivotal-jbarrett: My not just use `std::string` concatenation?

cppcache/include/geode/DataOutput.hpp
albertogpz: Not sure if this is possible. When the DataOutput constructor is invoked the m_bytes variable is assigned a static buffer which seems to be reused later when other DataOutput instances are created.

cppcache/integration-test/fw_dunit.cpp
albertogpz: Not really. The framework uses sometimes creates instances in the heap and sometimes in the stack.

cppcache/integration-test/testThinClientPartitionResolver.cpp
albertogpz: I finally removed this code. It was useless.

cppcache/integration-test/ThinClientFailover3.hpp
albertogpz: Sure

PR #566, opened 2019-12-24
GEODE-7612: Technical debt 01 - Fix logging

mreddington:
This code was supposed to disable debug level logging for non-debug builds. Apparently it never worked. I fixed it when I moved it to the source file, but it broke the tests. That means release could log debug. That means it's out in the wild. Instead of arbitrarily disabling a log level and possibly hampering a customer, we decided to disable this compile-time feature altogether.

### DIFF
--- a/cppcache/src/Log.cpp
+++ b/cppcache/src/Log.cpp
@@ -726,13 +726,7 @@ void Log::exitFn(LogLevel level, const char* functionName) {
 }
 
 bool Log::enabled(LogLevel level) {
-#ifdef DEBUG
-  return (((level == LogLevel::Debug) || GEODE_HIGHEST_LOG_LEVEL >= level) &&
-          s_logLevel >= level);
-#else
-  return (((level != LogLevel::Debug) && GEODE_HIGHEST_LOG_LEVEL >= level) &&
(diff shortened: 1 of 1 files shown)


pdxcodemonkey:
If we "fix the test," we're going to be changing the behavior of production code, which *could be* something a customer is depending on somewhere.  We're gonna err on the side of caution here.  The logging code is going to be replaced at our earliest opportunity anyway, best to just break it once and be done.

cppcache/src/Log.cpp
```diff
@@ -726,13 +726,7 @@ void Log::exitFn(LogLevel level, const char* functionName) {
 }
 
 bool Log::enabled(LogLevel level) {
-#ifdef DEBUG
```
pivotal-jbarrett: On closer inspection of the original code you will notice that nothing is defining `DEBUG` in the preprocessor so `s_doingDebug` was always false. This latest change introduces a bug to the logic overrode the results of `GEODE_HIGHEST_LOG_LEVEL >= level`, which appears to be redundant to the `DEBUG` macro. If you rip all that out then you end up with what's in this PR.

PR #567, opened 2019-12-27
GEODE-7625: Remove broken DH.

pivotal-jbarrett:
Replace with error message when DH is configured.


cppcache/src/SystemProperties.cpp
```diff
@@ -268,7 +268,8 @@ void SystemProperties::processProperty(const std::string& property,
     m_securityPropertiesPtr->insert(property, value);
 
     if (property == SecurityClientDhAlgo) {
-      m_securityClientDhAlgo = value;
+      throw IllegalArgumentException(
```
echobravopapa: Boom there it isn't.

cppcache/src/TcrConnection.cpp
```diff
@@ -47,10 +46,10 @@ const int8_t LAST_CHUNK_MASK = 0x1;
 const int64_t INITIAL_CONNECTION_ID = 26739;
 
 #define throwException(ex)                            \
-  {                                                   \
+  do {                                                \
```
echobravopapa: haven't looked at in c++ in a while (no pun intended) - what is the rationale for the do/while here?

pivotal-jbarrett: Fixes the multi line macro issue with semicolons. It gets compiled out.

PR #572, opened 2020-01-13
GEODE-7694: fix pdx type lookup

pdxcodemonkey:
The operator< method of PdxType in the native client simply compares the class names, which is incorrect and leads to the fact that, in NC, you cannot currently have two separate PDX types with the same class name. The comparison should favor type ID first, then if both type IDs are 0 (true for local regions), fall back to class name. This will still fail for local regions in the scenario where class names are the same, but it's much, much better than we have now.

### DIFF
--- a/cppcache/integration/test/CMakeLists.txt
+++ b/cppcache/integration/test/CMakeLists.txt
@@ -25,7 +25,7 @@ add_executable(cpp-integration-test
   ExpirationTest.cpp
   FunctionExecutionTest.cpp
   PdxInstanceTest.cpp
-  PdxJsonHandlerTest.cpp
+  PdxJsonTypeTest.cpp
   PdxSerializerTest.cpp
   RegionGetAllTest.cpp
   RegionPutAllTest.cpp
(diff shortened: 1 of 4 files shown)


cppcache/integration/test/PdxJsonHandlerTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <geode/PoolManager.hpp>
+#include <geode/RegionFactory.hpp>
+#include <geode/RegionShortcut.hpp>
+#include <geode/TypeRegistry.hpp>
+
+#include "CacheRegionHelper.hpp"
+#include "LocalRegion.hpp"
+#include "NestedPdxObject.hpp"
+#include "PdxType.hpp"
+namespace {
```
pivotal-jbarrett: Some white space here would be nice.

cppcache/integration/test/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ add_executable(cpp-integration-test
   ExpirationTest.cpp
   FunctionExecutionTest.cpp
   PdxInstanceTest.cpp
+  PdxJsonHandlerTest.cpp
```
pivotal-jbarrett: Where is the term "handler" coming from in this test name? The thing you are testing is PDX encode JSON via  the `PdxInstance` and `PdxType` class. You aren't registering a new `PdxTypeHandler` so the use of handler term is confusing me.

pdxcodemonkey: Yep, started investigating type handler on this branch, then fixed this bug and never changed the terminology.

PR #573, opened 2020-01-13
GEODE-7697: OpenSSL 1.1.0 minimum required, removed openssl-compat

mreddington:


mreddington:
I've updated the minimum version to 1.1.1.

pivotal-jbarrett:
Despite that fact that people shouldn't be using OpenSSL 1.1.0 or less, is there a compelling reason to remove support for it?

mreddington:
Yes, the OpenSSL documentation is explicit that it is not thread safe, and we see a host of runtime bugs because of it. Forcing a minimum version fixes them and reduces the amount of code we have to maintain for backwards compatibility that never actually worked.

pivotal-jbarrett:
> Yes, the OpenSSL documentation is explicit that it is not thread safe, and we see a host of runtime bugs because of it. Forcing a minimum version fixes them and reduces the amount of code we have to maintain for backwards compatibility that never actually worked.

For clarity, what is not thread safe? The whole of OpenSSL prior to 1.1.1?

pdxcodemonkey:
We've merged the SSL code into the client proper, so this is no longer relevant.

PR #576, opened 2020-02-14
GEODE-7872: Revert globbing of source.

pivotal-jbarrett:


pivotal-jbarrett:
> Can you pls file a JIRA ticket to check this in against? 

Typically build script changes haven't had separate JIRAs as it doesn't effect the code. I sure can create one if you really really want one though.

pdxcodemonkey:
Yes, please, let's go ahead.  I'm right now going through the process of tallying change info for a point release, and having like 10% or so of all the PRs merged without a corresponding ticket makes for weird bookkeeping.

PR #581, opened 2020-03-10
GEODE-7868: Support for Unsigned Types in .NET (and Guid)

mmartell:
The ReflectionBasedAutoSerializer for .NET was causing applications to crash when they used unsupported types. For example, doing a put of an object with a USHORT field causes a StackOverflow exception. With this fix, unsupported types will throw an IllegalStateException. 

Also, this fix helped us to understand how GUIDs were only partially working. Although they were being serialized correctly as objects, they were throwing IllegalStateException because they do not have a default constructor.


pdxcodemonkey:
What are the implications of this fix for compatibility with prior versions?  For instance, you mention GUIDs were only partially working - were they working enough to be useful?  And do they now no longer work at all?

clicache/src/ReflectionBasedAutoSerializer.cpp
```diff
@@ -536,20 +536,14 @@ namespace Apache
...
             return FieldType::ARRAY_OF_BYTE_ARRAYS;
           }
-          /*else if(type->Equals(Internal::DotNetTypes::ObjectArrayType))
-          {
-            //Giving more preference to arraylist instead of Object[] in java side
-            //return this->WriteObjectArray(fieldName, safe_cast<System::Collections::Generic::List<Object^>^>(fieldValue));
-            return FieldType::OBJECT_ARRAY;
-          }*/
-          else
+          else if(type->Equals(Internal::DotNetTypes::ObjectType))
```
pivotal-jbarrett: Does this check fail for primitive types then? What about the object version of the primitive types, won’t that pass this test and still cause issues?

pivotal-jbarrett:
I think you really have two choices when dealing with unsigned integer. Either treat them a signed in PDX or upscale them to the smaller signed integer. I would suggest treating them as signed in PDX as this is consistent with the C++ serializer.

mmartell:
This fix isn't really about Guids, Sorry for that confusion. Jake thinks Guids used to work. I only know that they aren't currently very useful. You can do puts and the Guid is serialized properly. However, when you do a get it throws an IllegalStateException "... no default constructor".

pivotal-jbarrett:
> This fix isn't really about Guids, Sorry for that confusion. Jake thinks Guids used to work. I only know that they aren't currently very useful. You can do puts and the Guid is serialized properly. However, when you do a get it throws an IllegalStateException "... no default constructor".


I am pretty sure GUID failed for some recursive parameter on serialization. In looking at the current implementation of GUID it looks like the internals may have changed. I could be wrong and confusing it with another type. Either way I think this change is not correct for GUID or primitives.

mmartell:
This fix isn't quite good enough. It properly throws an exception for System.Guid. However it causes the existing ThinClientPdxTests1 to fail due to lists and maps of System.Objects. The auto serializer needs to handle these as System.Objects instead of throwing an exception. Still investigating whether we can deserialize objects without a default ctor (like guid) using Jake's suggestion.

mmartell:
Regarding the handling of unsigned integers, I agree with the suggestion to handle them in a manner consistent with C++.

clicache/src/DataInput.hpp
```diff
@@ -271,10 +286,15 @@ namespace Apache
         List<Object^>^ ReadObjectArray();
 
         /// <summary>
-        /// Read a array of signed byte array from the stream.
+        /// Read an array of unsigned byte arrays from the stream.
```
pivotal-jbarrett: Are you really thinking about changing the behavior of the public API method to return unsigned bytes?

mmartell: No. The intent is merely to add support for currently unsupported types in .NET. Will make sure that I don't change any existing APIs. Just waiting on my superfly to complete. Hopefully will close this PR and open a new one today.

clicache/src/ReflectionBasedAutoSerializer.cpp
mmartell: There are no breaking changes in this PR. Support for new types have been implemented in new APIs.

mmartell:
Changed the title of this PR to better describe the functionality enhancements it provides.

mmartell:
Just noticed I have a bunch of LGTM alerts. Will look at those. pdxcodemonkey actually mentioned these a few days back.

clicache/src/DataInput.hpp
mmartell: This was a good catch. I did have some breaking changes in the API, which have now been fixed. All unsigned types are through new APIs.

clicache/src/impl/PdxLocalWriter.cpp
```diff
@@ -179,23 +179,29 @@ namespace Apache
             return this;
           }
 
-          IPdxWriter^ PdxLocalWriter::WriteByte( String^ fieldName, SByte value )
+          IPdxWriter^ PdxLocalWriter::WriteUnsignedByte( String^ fieldName, Byte value )
```
pdxcodemonkey: this file appears to be indented one space, instead of two?  Except where it contains tabs, of course.  Basically it looks all messed up, and we should probably pick a style and make it consistent.

Actually I guess it is two, but let's lose the tabs and whatever ails line 353.  Also pls resolve/remove your TODO comment on line 188

clicache/src/impl/PdxLocalWriter.hpp
```diff
@@ -83,18 +83,18 @@ namespace Apache
           }
 
           /// <summary>
-          /// Write a byte to the <c>IPdxWriter</c>.
+          /// Write an signed byte to the <c>IPdxWriter</c>.
```
pdxcodemonkey: 'a signed' is correct grammar.  Sorry to be a pain

clicache/src/impl/PdxLocalWriter.hpp
```diff
@@ -263,7 +270,7 @@ namespace Apache
           /// </summary>
           /// <param name="fieldName">The name of the field associated with the value.</param>
           /// <param name="ushortArray">The ushortArray to write.</param>
-          virtual IPdxWriter^ WriteUnsignedShortArray(String^ fieldName, array<System::UInt16>^ ushortArray);
+          virtual IPdxWriter^ WriteUShortArray(String^ fieldName, array<System::UInt16>^ ushortArray);
```
pdxcodemonkey: Why shorten the names of these?  This seems unnecessary....

clicache/src/impl/PdxReaderWithTypeCollector.cpp
```diff
@@ -99,45 +99,45 @@ namespace Apache
...
-            UInt16 retVal = PdxLocalReader::ReadUInt16(fieldName);
+            UInt16 retVal = PdxLocalReader::ReadUShort(fieldName);
             m_dataInput->RewindCursorPdx(position);
             return retVal;
           }
           return 0;
         }
 
-        System::UInt32 PdxReaderWithTypeCollector::ReadUInt32(String^ fieldName)
+        System::UInt32 PdxReaderWithTypeCollector::ReadUInt(String^ fieldName)
```
pdxcodemonkey: Again, this renaming feels gratuitous.  Also I really hate renaming UInt64 -> ULong.  It feels like we're losing important self-documentation.

clicache/src/impl/PdxReaderWithTypeCollector.hpp
```diff
@@ -80,19 +80,19 @@ namespace Apache
           /// Read a 16-bit unsigned integer from the stream.
           /// </summary>
           /// <param name="fieldName">The name of a member field whose value to read.</param>
-          virtual System::UInt16 ReadUInt16(String^ fieldName)override;
+          virtual System::UInt16 ReadUShort(String^ fieldName)override;
```
pdxcodemonkey: Insert standard naming rant here.  This one is more understandable, in that it was inconsistent in the original, but let's not lose any information in the process.

clicache/src/impl/PdxRemoteWriter.hpp
```diff
@@ -220,74 +220,74 @@ namespace Apache
           //virtual IPdxWriter^ writeFile(String fieldName, File file) ;
 
           /// <summary>
-          /// Write an collection to the <c>IPdxWriter</c>.
+          /// Write a boolean array to the <c>IPdxWriter</c>.
```
pdxcodemonkey: Lol, nice change.  Always good to do what in the heck you say you're gonna do

clicache/src/impl/PdxRemoteWriter.hpp
```diff
@@ -220,74 +220,74 @@ namespace Apache
...
           /// </summary>
           /// <param name="fieldName">The name of the field associated with the value.</param>
           /// <param name="boolArray">The boolArray to write.</param>
 					virtual IPdxWriter^ WriteBooleanArray( String^ fieldName, array<Boolean>^ boolArray) override;
 
           /// <summary>
-          /// Write an collection to the <c>IPdxWriter</c>.
+          /// Write an char array to the <c>IPdxWriter</c>.
```
pdxcodemonkey: What the heck is `Char` in this context?  8-bit?  16-bit?  Variable size?  Signed or unsigned?

clicache/src/impl/PdxRemoteWriter.hpp
```diff
@@ -220,74 +220,74 @@ namespace Apache
...
-          /// Write an collection to the <c>IPdxWriter</c>.
+          /// Write an signed byte array to the <c>IPdxWriter</c>.
           /// </summary>
           /// <param name="fieldName">The name of the field associated with the value.</param>
           /// <param name="sbyteArray">The sbyteArray to write.</param>
           virtual IPdxWriter^ WriteSByteArray(String^ fieldName, array<SByte>^ sbyteArray) override;
 
           /// <summary>
-          /// Write an collection to the <c>IPdxWriter</c>.
+          /// Write a 16 bit integer array to the <c>IPdxWriter</c>.
```
pdxcodemonkey: Thank you for adding explicit size remarks in the comments!!!

tests/javaobject/PdxTests/PdxType.java
```diff
@@ -102,21 +103,21 @@ public void init()
   {
     m_char = 'C';
     m_bool = true;
-    m_byte = (byte)0x74;
+    m_byte = (byte)0xa4;
```
pdxcodemonkey: What's the story behind changing these magic numbers?  Was there any significance to their choosing in the first place?  They don't appear to correspond to any kind of MAX_* values or things that test edge cases or anything, so why change 'em?

tests/javaobject/PdxTests/PdxType.java
```diff
@@ -126,20 +127,21 @@ public void init()
     //m_dateTime = new Date( System.currentTimeMillis());
     
     m_int16Array = new short[] { 0x2332, 0x4545 };
-    m_uint16Array = new short[] { 0x3243, 0x3232 };
+    m_uint16Array = new short[] { (short)0xa243, (short)0xffff };
```
pdxcodemonkey: These look like MAX_* constants, are there named constants for them anywhere?

tests/javaobject/PdxTests/PdxType.java
echobravopapa: are we just buying ourselves an extra 48 hours...

clicache/integration-test/ThinClientPdxTests2.cs
```diff
@@ -904,8 +907,18 @@ private void accessPdxInstance()
...
+      PdxType.GenericValCompare(retUInt16ArrayStr, dPdxTypeUShortArrayStr);
+
+      string retUintArrayStr = String.Join(",", ((int[])ret.GetField("m_uint32Array")).Select(p => ((uint)(int)p).ToString()).ToArray());
+      string dPdxTypeUintArrayStr = String.Join(",", dPdxType.Uint32Array.Select(p => p.ToString()).ToArray());
+      PdxType.GenericValCompare(retUintArrayStr, dPdxTypeUintArrayStr);
+
+      string retUlongArrayStr = String.Join(",", ((long[])ret.GetField("m_ulongArray")).Select(p => ((ulong)(long)p).ToString()).ToArray());
+      string dPdxTypeUlongArrayStr = String.Join(",", dPdxType.UlongArray.Select(p => p.ToString()).ToArray());
+      PdxType.GenericValCompare(retUlongArrayStr, dPdxTypeUlongArrayStr);
+
```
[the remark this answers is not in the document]
mmartell: Yah, the full story. Some consider ulong (same as ULONG), which are 64 bits in .NET, to add confusion. I actually like it. It's C++'s long long and unsigned long long that I dislike. Where Microsoft screwed up is making long and unsigned long only 32 bits in Windows: "... The true "war" was for sizeof(long), where Microsoft decided for sizeof(long) == 4 (LLP64) while nearly everyone else decided for sizeof(long) == 8 (LP64)."

There are no unsigned types in Java. Period. So there was no unsigned support for them in Pdx. This PR fixes that. Not just for the primitives, but for the arrays of the primitives.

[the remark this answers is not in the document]
mmartell: Regarding Java interop with unsigned types, I believe the only issue is if you try to use an unsigned type that exceeds the range of the signed type. For example, 16 bit integers go from -32768 to 32767 for signed, and 0 to 65535 for unsigned. So if you put a value of 50K from C# for example, this shows up as a -4350 in Java. So would be an access violation if you tried to use that as an index into an array in your Java code. As a minimum, we should document this in our Java interop docs (if we have such a thing). I believe the main problem we're trying to solve here is to provide support for the .NET unsigned types. People use unsigned types because they are natural (think size_t), not because they double the range.

clicache/src/IPdxWriter.hpp
```diff
@@ -42,10 +42,17 @@ namespace Apache
 				public:
 	        
 					/// <summary>
-					/// Write a byte to the <c>IPdxWriter</c>.
+					/// Write an unsigned byte to the <c>IPdxWriter</c>.
 					/// </summary>
 					/// <param name="fieldName">The name of the field associated with the value.</param>
-					/// <param name="value">The byte to write.</param>
+					/// <param name="value">The unsigned byte to write.</param>
```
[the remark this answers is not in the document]
mmartell: No, this goes with the new API WriteUnsignedByte.

clicache/src/DataInput.cpp
```diff
@@ -1029,6 +1056,22 @@ namespace Apache
...
+        {
+          array<array<SByte>^>^ retVal = gcnew array<array<SByte>^>(len);
+          for (int i = 0; i < len; i++)
+          {
+            retVal[i] = this->ReadSBytes();
+          }
+          return retVal;
+        }
+        else
+          return nullptr;
```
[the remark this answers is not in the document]
mmartell: I agree this is confusing code. I only wanted to add unsigned support to the preexisting pattern in ReadArrayOfByteArrays, and not risk breaking something. If we really want to improve the code here we should open a new ticket.

clicache/src/IPdxInstanceFactory.hpp
```diff
@@ -87,6 +87,18 @@ namespace Apache
...
+        /// Writes the named field with the given value to the serialized form.
+        /// The fields type is <code>unsigned byte</code>.
+        /// <para>Note: Java doesn't have an unsigned types.</para>
+        /// </summary>
+        /// <param name="fieldName"> the name of the field to write</param>
+        /// <param name="value"> the value of the field to write</param>
+        /// <returns> this PdxInstanceFactory</returns>
+        /// <exception cref="IllegalStateException"/> if the named field has already been written</exception>
+          
+        IPdxInstanceFactory^ WriteUnsignedByte(String^ fieldName, Byte value);
```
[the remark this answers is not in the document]
mmartell: This exercise has indeed uncovered a few API docs bugs. I'll be submitting a ticket for that. You might be right about breaking the ABI due to adding new methods to the interface. I'll test that.

clicache/src/IPdxReader.hpp
```diff
@@ -47,6 +47,12 @@ namespace Apache
 					/// </summary>
 					/// <param name="fieldName">The name of a member field whose value to read.</param>
 					SByte ReadByte( String^ fieldName );
+
```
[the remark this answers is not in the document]
mmartell: Agreed. I will see if the additional methods in the IPdxReader, IPdxWriter, and IPdxInstanceFactory require a recompile of the app.

clicache/src/ReflectionBasedAutoSerializer.cpp
```diff
@@ -150,70 +150,100 @@ namespace Apache
               {
                 case FieldType::BOOLEAN:
                   w->WriteBoolean(m_fieldName, (bool)value);
-                break;
+                  break;
```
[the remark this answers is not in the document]
mmartell: Good catch. Looks like the original file had formatting issues. Fixing now.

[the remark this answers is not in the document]
mmartell: Fixed

clicache/src/ReflectionBasedAutoSerializer.cpp
```diff
@@ -320,10 +380,10 @@ namespace Apache
         void ReflectionBasedAutoSerializer::serializeFields(Object^ o,IPdxWriter^ writer )
         {
           Type^ ty = o->GetType();
-         // Log::Debug("ReflectionBasedAutoSerializer::serializeFields classname {0}: objectType {1}", o->GetType()->FullName,o->GetType());
+          Log::Debug("ReflectionBasedAutoSerializer::serializeFields classname {0}: objectType {1}", o->GetType()->FullName,o->GetType());
```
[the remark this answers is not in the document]
mmartell: Log::Debug is our standard in C++/CLI. Don't see a reason to keep these though. Will delete.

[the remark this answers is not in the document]
mmartell: Done.

clicache/src/ReflectionBasedAutoSerializer.hpp
```diff
@@ -38,27 +38,37 @@ namespace Apache
       public enum class FieldType
       {
         BOOLEAN,
+        SBYTE,
```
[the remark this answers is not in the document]
mmartell: Good catch. Can we just add the new enums at the end of the list?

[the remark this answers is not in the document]
mmartell: For now just fixing the formatting.

clicache/src/impl/PdxInstanceImpl.cpp
```diff
@@ -1309,11 +1309,15 @@ namespace Apache
 
           void PdxInstanceImpl::writeField(IPdxWriter^ writer, String^ fieldName, int typeId, Object^ value)
           {
+            Type^ type = value->GetType();
             switch (typeId)
             {
             case PdxFieldTypes::INT:
             {
-              writer->WriteInt(fieldName, (int)value);
+              if (type == Int32::typeid)
```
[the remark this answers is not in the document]
mmartell: Fixed.

clicache/src/impl/PdxLocalReader.cpp
```diff
@@ -115,14 +115,15 @@ namespace Apache
           return m_pdxRemotePreserveData;
         }
 
-        SByte PdxLocalReader::ReadByte( String^ fieldName )
+        Byte PdxLocalReader::ReadUnsignedByte( String^ fieldName )
         {
-					return m_dataInput->ReadSByte();
+					return m_dataInput->ReadByte();
```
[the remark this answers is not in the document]
mmartell: I'm fixing the ones you've caught. I'll create a new ticket for fixing tab issues in C++/CLI.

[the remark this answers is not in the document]
mmartell: Fixed.

clicache/src/impl/PdxLocalReader.cpp
```diff
@@ -308,6 +314,16 @@ namespace Apache
           return m_dataInput->ReadArrayOfByteArrays();
         }
 
+				array<array<SByte>^>^ PdxLocalReader::ReadArrayOfSByteArrays(String^ fieldName )
```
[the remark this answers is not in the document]
mmartell: Fixed. I'm not sure, but what may have happened is that I had turned off clang-format on my machine due to it crashing during the build of clicache. So perhaps some of these formatting issues were prexistent but didn't show up since they hadn't been changed in a long time.

clicache/src/impl/PdxLocalReader.hpp
```diff
@@ -105,19 +105,19 @@ namespace Apache
...
         /// <param name="fieldName">The name of a member field whose value to read.</param>
-        virtual System::UInt32 ReadUInt32( String^ fieldName );
+        virtual System::UInt32 ReadUInt( String^ fieldName );
         
         /// <summary>
         /// Read a 64-bit unsigned integer from the stream.
         /// </summary>
         /// <param name="fieldName">The name of a member field whose value to read.</param>
-        virtual System::UInt64 ReadUInt64( String^ fieldName );
+        virtual System::UInt64 ReadULong( String^ fieldName );
```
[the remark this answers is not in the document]
mmartell: A bit of a religious war here. I was in your camp originally. Using the bitlength suffix avoids any confusion. However, nearly all Microsoft docs use the C# keywords (short, int, long, and their unsigned counterparts (ushor, uint, ulong) instead of the .NET types (Int16, Int32, Int64, etc.). In light of this discovery I switched back to using the C# keywords in the API. I'm good either way, whatever the community prefers.

clicache/src/impl/PdxLocalWriter.cpp
```diff
@@ -136,7 +136,7 @@ namespace Apache
             if(len <= 0xff)
             {
               for(int i = m_offsets->Length-1; i >0 ; i--)
-                m_dataOutput->WriteByte((Byte)m_offsets[i]);
+                m_dataOutput->WriteByte((SByte)m_offsets[i]);
```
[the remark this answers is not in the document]
mmartell: Yup. Good catch.

clicache/src/impl/PdxLocalWriter.cpp
mmartell: Fixed.

clicache/src/impl/PdxLocalWriter.hpp
mmartell: Not a pain. Your diligence is appreciated.

Fixed.

clicache/src/impl/PdxLocalWriter.hpp
mmartell: Discussed above in an earlier conversation. Short, Int, and Long are the C# keywords used in Microsoft docs.

clicache/src/impl/PdxReaderWithTypeCollector.cpp
mmartell: Same as above. Microsoft uses the C# keywords in their API docs.

clicache/src/impl/PdxReaderWithTypeCollector.hpp
mmartell: Same as above. Microsoft uses the C# keywords in their API docs.

clicache/src/impl/PdxRemoteWriter.hpp
mmartell: Char is technically neither signed or unsigned. It has its own rules regarding casts and how const chars are initialized.

tests/javaobject/PdxTests/PdxType.java
mmartell: This indeed warrants some explanation. The PdxType.java is the Java equivalent of PdxType.cs in the clicache tests. (Actually, since java came first it may be better to say PdxType.cs is the C# equivalent of PdxType.java. Note that PdxType.java has a pair of variables for many types (m_byte and m_sbyte, m_int16Array and m_uint16Array, for example). But each member of the pair is a signed type, as it must be since Java has no unsigned types.

Since this PR adds support for unsigned types in the .NET client, it seemed prudent to use unsigned values for the constants which would appear as negative numbers in Java to differentiate them from signed values. This provides for better testing of our unsigned APIs in my opinion.

tests/javaobject/PdxTests/PdxType.java
mmartell: These java variables map to unsigned types on the C# side, where we do use the MAX* constants. I guess you could use whatever the macro for MIN_int16 (or whatever it's called) on the Java side, but that may be confusing.

clicache/src/ReflectionBasedAutoSerializer.hpp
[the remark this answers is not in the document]
mmartell: ABI break has been fixed by moving new enums to the end of the list.

clicache/src/IPdxReader.hpp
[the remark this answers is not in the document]
mmartell: Adding new member functions to the interface doesn't break the ABI. All Pdx tests still pass (without needing to be rempiled) when run against the new Apache.Geode.dll. This seems consistent with what others are saying. It's only changing function signatures and enums (and a few others) that break the ABI in .NET.

clicache/src/IPdxInstanceFactory.hpp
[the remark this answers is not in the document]
pivotal-jbarrett: ABI is fine.

mmartell:
This PR is being replaced with a new PR. The new PR adds support for unsigned types (and also Decimal and Guid) by just enhancing the ReflectionBasedAutoSerializer. The new PR does not add any new APIs to Pdx.

PR #583, opened 2020-03-16
GEODE-7883: Add a test to verify native client handles unnamed pool…

pdxcodemonkey:
…in cache XML.  Previously had a report that this would core dump in old versions.
- Added new test CacheXmlTest.loadCacheWithUnnamedPool to verify
- As expected, test passes.  XML file fails to load, and native client returns a CacheXmlException with text to the effect of "pool tag is missing name attribute"


pdxcodemonkey:

PR #584, opened 2020-03-17
GEODE-7887: Remove reference to nonexistent API from docs

pdxcodemonkey:
- setClientNotificationEnabled is a protected method on an internal class, not part of the public API


davebarnes97:
+1 LGTM
Corresponding mentions in the User Guides (C++ and .NET), which were limited to the upgrade page, have been removed on the develop (aka 1.13) branch of the repo.

PR #593, opened 2020-04-27
GEODE-7944: Fix for QueueConnectionRequest message

jvarenina:
ClientProxyMembershipID correctly encapsulated within QueueConnectionRequest
to avoid "Unable to deserialize membership id java.io.EOFException" on
locator when trying to log recevied ClientProxyMembershipID on debug level


cppcache/src/QueueConnectionRequest.cpp
```diff
@@ -28,10 +28,10 @@ void QueueConnectionRequest::toData(DataOutput& output) const {
   output.writeString(m_serverGp);
   output.write(static_cast<int8_t>(DSCode::FixedIDByte));
   output.write(static_cast<int8_t>(DSCode::ClientProxyMembershipId));
-  uint32_t buffLen = 0;
-  output.writeBytes(reinterpret_cast<uint8_t*>(const_cast<char*>(
-                        m_membershipID.getDSMemberId(buffLen))),
-                    buffLen);
+  uint32_t memIdBufferLength = 0;
+  auto memIdBuffer = m_membershipID.getDSMemberId(memIdBufferLength);
```
pivotal-jbarrett, line 32: Internally the member ID is a std::string. This method is breaking it into a c-string and length. Why not just modify the method to return a `const std::string&`, you get the string and length encapsulated.

cppcache/test/QueueConnectionRequestTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+  return to_hex(out.getBuffer(), out.getBufferLength());
+}
+
+inline std::string int_to_hex_string(const int value) {
+  char hex[10];
+  sprintf(hex, "%x", value);
+  return std::string(hex);
+}
+
+TEST(QueueConnectionRequestTest, testToData) {
```
pivotal-jbarrett, line 52: While I love to see another unit test, it seems like there should have been a failing integration test. Some of the tests are disabled because they never worked. Perhaps there is one that matches this failure.

cppcache/src/QueueConnectionRequest.cpp
pivotal-jbarrett:
```c++
  const auto& dsMemberId = m_membershipID.getDSMemberId();
  output.writeBytes(reinterpret_cast<const int8_t*>(dsMemberId.c_str()), dsMemberId.size());
```

pivotal-jbarrett: The use of `writeBytes` seems suspicious. Have you double checked the Java side for who it writes and reads this string? There are only about 7 ways we encode strings. ;)

jvarenina: Thanks for comments. I agree that it will be good to modify method to return const std::string&.
I will update code the way you suggested. The reason why I was reserved in changing getDSMemberId is because it is used on two more places in native client when serializing membershipID.

jvarenina: Related to writeBytes, I tested this manually to see that everything is de-serialized correctly on locator side.
Also, on two other places in native client the writeBytes is used to serialize ClientProxyMebershipID 
when sent in handshake message towards servers.

When created and serialized in client:
[info 2020/04/21 12:50:19.371347 CEST jakov:14391 140683540789248] Using Native_bfbgddgaaf14391 as random data for ClientProxyMembershipID
[debug 2020/04/21 12:50:19.378174 CEST jakov:14391 140683540789248] GethashKey :0:0:0:0:0:default_GeodeDS:Native_bfbgddgaaf14391 client id: jakov(14391:loner):2:Native_bfbgddgaaf14391:default_GeodeDS


When de-serialized in locator and printed with toString:
Locator received request QueueConnectionRequest{group=, excluded=[], redundant= -1,findDurable=false,proxyId=identity(0.0.0.0(default_GeodeDS:14391:loner):2:Native_bfbgddgaaf14391:default_GeodeDS,connection=1,durableAttributes=DurableClientAttributes[id=31_gem_pool; timeout=300])}

When de-serialized  in servers and printed with toString:
Initializing proxy: CacheClientProxy[identity(0.0.0.0(default_GeodeDS:14391:loner):2:Native_bfbgddgaaf14391:default_GeodeDS,connection=1,durableAttributes=DurableClientAttributes[id=31_gem_pool; timeout=300]); port=34308; primary=true; version=GFE 9.0]

cppcache/test/QueueConnectionRequestTest.cpp
jvarenina: I didn't create integration test, because It seemed to me that non of the functionality is affected with this bug beside debug logging in locator. Native client always sends QueueConnectionRequest with flag findDurable=false, so membershipId is never used for anything functional in locator, it is only logged at debug level.

I have put more details related to this in mails sent towards dev@geode.apache.org, mail Subject: Unable to get behavior described in documentation when using durable native client

cppcache/src/QueueConnectionRequest.cpp
pivotal-jbarrett: > The reason why I was reserved in changing getDSMemberId is because it is used on two more places in native client when serializing membershipID.

Almost always a good rule of thumb but this code has lots of debt. You see from this fix there is the stink of old C-style code here. As a general rule of thumb we have adopted that all dynamic array types should be container wrapped. So you will see a shift from c-string/len pairs to `std::string`, byte*/len pairs to std::vector<byte>, etc. So if you are in an area of code to fix it please take that opportunity to fix it up. See our https://github.com/apache/geode-native/blob/develop/CONTRIBUTING.md guide for more cleanup tasks.

cppcache/test/QueueConnectionRequestTest.cpp
pivotal-jbarrett: Oh, that's right, the bug was just the server side logging null pointer.

cppcache/src/QueueConnectionRequest.cpp
jvarenina: Thanks, I'll have that in mind.

PR #597, opened 2020-05-03
GEODE-8063: Fixes access violation.

jake-at-work:
The external Cache pointer is not valid when destruction has started. Use
internal CacheImpl from RegionInternal.

### DIFF
--- a/cppcache/src/CacheImpl.cpp
+++ b/cppcache/src/CacheImpl.cpp
@@ -494,7 +494,7 @@ std::shared_ptr<Region> CacheImpl::getRegion(const std::string& path) {
     }
   }
 
-  if (region && isPoolInMultiuserMode(region)) {
+  if (region && isPoolInMultiuserMode(*region)) {
     LOGWARN("Pool " + region->getAttributes().getPoolName() +
             " attached with region " + region->getFullPath() +
             " is in multiuser authentication mode. Operations may fail as "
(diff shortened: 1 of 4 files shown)


cppcache/src/CacheImpl.cpp
```diff
@@ -640,13 +640,16 @@ void CacheImpl::readyForEvents() {
...
   if (!poolName.empty()) {
-    auto poolPtr = regionPtr->getCache().getPoolManager().find(poolName);
-    if (poolPtr != nullptr && !poolPtr->isDestroyed()) {
-      return poolPtr->getMultiuserAuthentication();
+    auto pool = static_cast<RegionInternal&>(*region)
+                    .getCacheImpl()
+                    ->getPoolManager()
+                    .find(poolName);
+    if (pool && !pool->isDestroyed()) {
+      return pool->getMultiuserAuthentication();
```
pdxcodemonkey, line 652: Before I approve and merge this, can you explain why a method on CacheImpl is checking whether its pool is in multi-user mode by getting hold of _another_ CacheImpl pointer and extracting the pool from there?  Why can't CacheImpl call getPoolManager() on itself?  This code is super weird without any context...

pdxcodemonkey (changes requested):

cppcache/src/CacheImpl.cpp
pdxcodemonkey: Also is there a reasonable way to write a test to force this race condition or something?

pivotal-jbarrett: It's a really good question. I suppose it could use `this` and some that the `region` is owned by `this` `CacheImpl`.  Even better refactor.

Testing this would require a concurrency test framework we don't have and then magical timing. You could write a flaky test that tries over and over in hopes of hitting it but not hitting it doesn't prove it is fixed. Through analysis of the object lifecycle we can at least guarantee that if we get into this call that `this` `CacheImpl` has not been destructed.

pdxcodemonkey (approved):

PR #604, opened 2020-05-24
GEODE-8183: Update dockerfile versions

moleske:
- Geode from 1.9 to 1.12
- CMake from 3.12.2 to 3.12.4
- Set to noninteractive so that building doesn't require interaction
- Make key=value for LABEL clearer

I'm not really sure how the Travis dockerfile has stayed up to date.  I'm assuming it is part of the release process to update it, but changes never get ported here since this was trying to install 1.9 which you can't download anymore.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,9 +15,10 @@
 # limitations under the License.
 
 FROM ubuntu
-LABEL maintainer Apache Geode <dev@geode.apache.org>
+LABEL maintainer="Apache Geode <dev@geode.apache.org>"
 
 ENV CLANG_VERSION 6.0
+ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && \
         apt-get install -y \
             libc++-dev \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -41,15 +42,15 @@ RUN apt-get update && \
...
+ENV GEODE_VERSION 1.12.0
 RUN wget "https://www.apache.org/dyn/closer.cgi?action=download&filename=geode/${GEODE_VERSION}/apache-geode-${GEODE_VERSION}.tgz" --quiet -O - | \
         tar xzf -
 
 ENV RAT_VERSION 0.13
 RUN wget "https://www.apache.org/dyn/closer.cgi?action=download&filename=creadur/apache-rat-${RAT_VERSION}/apache-rat-${RAT_VERSION}-bin.tar.gz" --quiet -O - | \
         tar xzf -
 
-ENV CMAKE_VERSION 3.12.2
+ENV CMAKE_VERSION 3.12.4
```
pivotal-jbarrett, line 53: Might as well take it to 3.17 or whatever the latest is today.

moleske: I stuck with 3.12 since that is min version we specify in the CMakeLists.txt.  It seemed worth keeping that in sync but maybe it doesn't matter?

PR #606, opened 2020-05-28
GEODE-8181: Fixes default value for statistic-sampling-enabled.

pivotal-jbarrett:
GEODE-8181: Fixes default value for statistic-sampling-enabled.

### DIFF
--- a/cppcache/integration-test/resources/system.properties
+++ b/cppcache/integration-test/resources/system.properties
@@ -14,7 +14,7 @@
 # limitations under the License.
 # All the configurable parameters.
 statistic-sample-rate=700s
-statistic-sampling-enabled=false
+statistic-sampling-enabled=true
 statistic-archive-file=stats.gfs
 log-file=geode-native.log
 #log-level=debug
--- a/cppcache/integration-test/testSystemProperties.cpp
+++ b/cppcache/integration-test/testSystemProperties.cpp
@@ -51,7 +51,7 @@ BEGIN_TEST(DEFAULT)
     ASSERT(
         systemProperties->statisticsSampleInterval() == std::chrono::seconds(1),
         "expected 1");
-    ASSERT(systemProperties->statisticsEnabled() == true, "expected true");
+    ASSERT(systemProperties->statisticsEnabled() == false, "expected false");
     LOG(systemProperties->statisticsArchiveFile());
     auto &&statisticsArchiveFileName =
         systemProperties->statisticsArchiveFile();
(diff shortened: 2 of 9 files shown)


cppcache/integration-test/testSystemProperties.cpp
```diff
@@ -51,7 +51,7 @@ BEGIN_TEST(DEFAULT)
     ASSERT(
         systemProperties->statisticsSampleInterval() == std::chrono::seconds(1),
         "expected 1");
-    ASSERT(systemProperties->statisticsEnabled() == true, "expected true");
+    ASSERT(systemProperties->statisticsEnabled() == false, "expected true");
```
pdxcodemonkey, line 54: The string description should now be "expected false," no?

cppcache/integration-test/testSystemProperties.cpp
```diff
@@ -80,7 +80,7 @@ BEGIN_TEST(NEW_CONFIG)
                std::chrono::seconds(700),
            "expected 700");
 
-    ASSERT(systemProperties->statisticsEnabled() == false, "expected false");
+    ASSERT(systemProperties->statisticsEnabled() == true, "expected false");
```
pdxcodemonkey, line 83: Same here

cppcache/integration-test/resources/system.properties
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 # All the configurable parameters.
 statistic-sample-rate=700s
-statistic-sampling-enabled=false
+statistic-sampling-enabled=true
```
pdxcodemonkey, line 17: I don't understand this change.  Doesn't this change the value to true for _all_ integration tests now?  I thought only the expiration tests required this, i.e. the two places where you're now explicitly enabling them in test code.

cppcache/integration-test/testSystemProperties.cpp
pivotal-jbarrett: Ha! Details!

cppcache/integration-test/resources/system.properties
pivotal-jbarrett: No it’s used by a single integration test.

pdxcodemonkey: Ugh.  I'd ask that we rename this file to reflect that it's test-case-specific, but I think time is better spent just getting rid of the old tests in general.  Thanks for the clarification.

pivotal-jbarrett: Yeah, when I need more time in C++ again after ACE is gone...

PR #608, opened 2020-05-30
GEODE-8207: Enfore No Unknown Pragmas as Error

moleske:
I just used `__sun` and `_MSC_VER` as the check.  Worked for Windows, Mac, and Kubuntu 20.04.  Don't have a Sun based setup to test that one.  I checked the box for "Allow edits by maintainers" so just push up anything you might discover.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,7 +215,6 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
   target_compile_options(_WarningsAsError INTERFACE
     -Werror
     -Wall
-    -Wno-unknown-pragmas #TODO fix
     -Wno-unused-variable #TODO fix
     -Wpedantic
     # -Wshadow TODO fix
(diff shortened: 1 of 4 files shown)


moleske:
@pivotal-jbarrett I switched to definitions in config.h, but also spent a little time trying to figure out the correct way in cmake to detect those pragmas.  I didn't succeed in the latter, kept ending up in docs for checking compiler flags.  The closest I hit to was `get_target_property` but couldn't figure out what target or property I'd really be looking for.  If you have thoughts on detection, I can take another try at using cmake to determine if those pragmas exist (as long as folks don't mind waiting a week in between my attempts based on my current schedule)

pivotal-jbarrett:
> I switched to definitions in config.h

I forgot that those were unprefixed definitions in there. We should fix that some day so they don't collide with others.

> but also spent a little time trying to figure out the correct way in cmake to detect those pragmas. I didn't succeed in the latter, kept ending up in docs for checking compiler flags. The closest I hit to was `get_target_property` but couldn't figure out what target or property I'd really be looking for. If you have thoughts on detection, I can take another try at using cmake to determine if those pragmas exist (as long as folks don't mind waiting a week in between my attempts based on my current schedule)

I believe you want to use `check_cxx_source_compiles`.

PR #613, opened 2020-06-04
GEODE-8227: Updates CMake to 3.16.8.

pivotal-jbarrett:


moleske:
Is this to update the minimum required version also?  Or just to get packer things updated?  The Jira issue doesn't go into detail

pivotal-jbarrett:
> Is this to update the minimum required version also? Or just to get packer things updated? The Jira issue doesn't go into detail

Little of both... The current images have a really old CMake. There are some ACE removal branches that need newer CMake.

moleske:
> Little of both... The current images have a really old CMake. There are some ACE removal branches that need newer CMake.

Thanks for the information.  My request would be to update https://github.com/apache/geode-native/blob/develop/CMakeLists.txt#L16 to new required version of cmake

pivotal-jbarrett:
> Thanks for the information. My request would be to update https://github.com/apache/geode-native/blob/develop/CMakeLists.txt#L16 to new required version of cmake

Chicken and egg problem. If you make that change in this PR the CI fails because the images aren't updated with the new version of CMake. Technically our CMake files are currently compatible with older versions of CMake. Though that will change soon.

moleske:
> Chicken and egg problem. If you make that change in this PR the CI fails because the images aren't updated with the new version of CMake. Technically our CMake files are currently compatible with older versions of CMake. Though that will change soon.

No problem 😄 Just want to make sure stuff comes to eventual consistency

PR #615, opened 2020-06-08
GEODE-8231: remove bucket server location from metadata when server down

albertogpz:


cppcache/integration/test/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ add_executable(cpp-integration-test
   ExampleTest.cpp
   ExpirationTest.cpp
   FunctionExecutionTest.cpp
+  PartitionRegionOpsWithRedundancyAndServerGoesDown.cpp
```
moleske, line 30: I think it general we've been having test files end in `Test` just to make it super obvious when browsing files it is a test.  Also the file reads more like a test case rather than an identifier for a set of tests.  I think the test file being called `PartitionRegionOpsTest.cpp` would work since you use `getPartitionedRegionWithRedundancyServerGoesDown` and putPartitionedRegionWithRedundancyServerGoesDown` as your test names.  We can see if others have thoughts about naming

albertogpz: I agree with you.

cppcache/src/ClientMetadata.cpp
```diff
@@ -114,14 +114,24 @@ const std::string& ClientMetadata::getColocatedWith() {
   return m_colocatedWith;
 }
 
+void ClientMetadata::removeBucketServerLocation(
+    const std::shared_ptr<BucketServerLocation>& serverLocation) {
+  for (auto&& locations : m_bucketServerLocationsList) {
+    for (unsigned int i = 0; i < locations.size(); i++) {
+      if (locations[i]->getEpString() == (serverLocation->getEpString())) {
+        locations.erase(locations.begin() + i);
```
vjr, line 122: Won't erasing an item from `locations` make it invalid in the `for` loop's `locations.size()` ?

albertogpz: Yes it will but given that when an element is erased we right after exit the loop with the break sentence then it does not matter.

vjr: Oh yes, there is a `break` statement, I am blind!

cppcache/integration/test/CMakeLists.txt
mmartell: I tend to always work in an IDE that supports project view (Visual Studio, Xcode), wherein I already know I'm working with a test project. Hence the Test suffix is a bit redundant. Also, we already have 4 tests in the new framework that don't use the Test suffix.

cppcache/integration/test/PartitionRegionOpsTest.cpp
```diff
@@ -0,0 +1,169 @@
...
+std::shared_ptr<Region> setupRegion(Cache& cache,
+                                    const std::shared_ptr<Pool>& pool) {
+  auto region = cache.createRegionFactory(RegionShortcut::PROXY)
+                    .setPoolName(pool->getName())
+                    .create("region");
+
+  return region;
+}
+
+void putEntries(std::shared_ptr<Region> region, int numEntries,
```
alb3rtobr, line 76: I would call this method `putAndGetEntries`

cppcache/integration/test/PartitionRegionOpsWithRedundancyAndServerGoesDown.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  return region;
+}
+
+void putEntries(std::shared_ptr<Region> region, int numEntries,
+                int offsetForValue) {
+  for (int i = 0; i < numEntries; i++) {
+    auto key = CacheableKey::create(i);
+    region->put(key, Cacheable::create(std::to_string(i + offsetForValue)));
+    auto value = region->get(key);
```
mmartell: I'd rename this function to putAndGetEntries. Or perhaps remove the gets, since you already have a getEntries function.

mmartell: Nice test!

cppcache/integration/test/PartitionRegionOpsTest.cpp
albertogpz: I prefer to remove the gets from the method.

cppcache/integration/test/PartitionRegionOpsWithRedundancyAndServerGoesDown.cpp
albertogpz: Thanks. I'll remove the gets from the method.

PR #617, opened 2020-06-11
Update C++ serialization example

davebarnes97:


docs/geode-native-docs-cpp/serialization/cpp-serialization/pdxserializable-interface.html.md.erb
```diff
@@ -27,88 +27,60 @@ A domain class should serialize and de-serialize all its member fields in the sa
...
-       sprintf(idbuf,"PdxObject: [ ID=%d ]",m_id);
-       return CacheableString::create( idbuf );
-}
-```
-
+    ```
+    The fields that are marked as identity fields are used to generate the `hashCode` and equality
+    methods of PdxInstance. Because of this, the identity fields should themselves either be primitives,
+    or implement `hashCode` and equals.  The identity fields should be marked after they are written
+    using a `write*` method.
```
karensmolermiller, line 86: Line 85-86, I'd prefer to see a different wording that directly states that we are ordering the function calls. Perhaps something like:
"Invoke the `markIdentityField` function directly following its `write*` function.

I would do the same as surrounding prose, but I think of C++ as having functions, while Java has methods. Probably no one cares which term you use.

davebarnes97: Good suggestions, @karensmolermiller. I checked in a revision - please give it a second look.

PR #619, opened 2020-06-16
GEODE-8212: Reduce connections to server to get type id

alb3rtobr:


cppcache/src/PdxType.cpp
```diff
@@ -557,11 +557,30 @@ bool PdxType::Equals(std::shared_ptr<PdxType> otherObj) {
...
+
+int32_t PdxType::hashcode() const {
+  std::hash<std::string> strHash;
+  auto result=strHash(this->m_className);
+
+  for (std::vector<std::shared_ptr<PdxFieldType>>::iterator it =
+          m_pdxFieldTypes->begin();
+       it != m_pdxFieldTypes->end(); ++it) {
+    auto pdxPtr = *it;
+    result = result ^ (strHash(pdxPtr->getClassName()) << 1 );
```
pdxcodemonkey, line 579: Do we need to xor in the classname each time through the loop?  It seems like just using the field names in here would be more appropriate

pdxcodemonkey: One thing that is essential we get right here is that ordering of fields should have no bearing on the generated hash.  so { "foo": 7, "bar" 10 } is the same type as { "bar": 12, "foo": 11 }, e.g.  Otherwise we end up with a "type explosion" of zillions of copies of the same thing each getting their own type ID.  If we're lucky, the server side will sort the field names when generating the type ID on that side, and they'll all get the same ID, but then we've still made a ton of unnecessary trips to the server.

alb3rtobr: > Do we need to xor in the classname each time through the loop? It seems like just using the field names in here would be more appropriate

I added it because I thought in the case of two different classes having the same fields (names & types).



> One thing that is essential we get right here is that ordering of fields should have no bearing on the generated hash. so { "foo": 7, "bar" 10 } is the same type as { "bar": 12, "foo": 11 }, e.g. Otherwise we end up with a "type explosion" of zillions of copies of the same thing each getting their own type ID. If we're lucky, the server side will sort the field names when generating the type ID on that side, and they'll all get the same ID, but then we've still made a ton of unnecessary trips to the server.

In the tests I did, I saw that the order of the fields does not change the result of the hash.

pdxcodemonkey: Just to be clear, I didn't mean to suggest we not hash in the classname at all, just outside the for loop.  Also for what it's worth, even though using classname is still correct, the only instance of PDX we're aware of that has this problem, i.e. the classname isn't unique, is for JSON instances, where classname is always "__GEMFIRE_JSON".

pdxcodemonkey: It would be good to have a unit test that shows the hashcode doesn't change if you iterate the fields in a different order...

alb3rtobr: > Just to be clear, I didn't mean to suggest we not hash in the classname at all, just outside the for loop. Also for what it's worth, even though using classname is still correct, the only instance of PDX we're aware of that has this problem, i.e. the classname isn't unique, is for JSON instances, where classname is always "__GEMFIRE_JSON".

oh, I see the confusion: the classname its outside the for loop (line 573). The other classnames you are refering to are the classnames of each field. If we dont include the name of the field and its classname, the hash will be the same for an object with a bool called `foo` and for an object with a string called `foo`.

cppcache/src/PdxTypeRegistry.hpp
```diff
@@ -53,7 +53,20 @@ typedef std::unordered_map<std::shared_ptr<PdxSerializable>,
...
+
+struct PdxTypeHashCode {
+  std::size_t operator()(std::shared_ptr<PdxType> const& pdx) const {
+    return pdx ? pdx->hashcode() : 0;
+  }
+};
+
+struct PdxTypeEqualCmp {
+  bool operator()(std::shared_ptr<PdxType> const& first, std::shared_ptr<PdxType> const& second) const{
+    return first->hashcode() == second->hashcode();
```
pivotal-jbarrett, line 65: Equality of hashcode does not guarantee equality of the object. You should just implement `PdxType::operator==`.

cppcache/src/PdxType.cpp
alb3rtobr: > It would be good to have a unit test that shows the hashcode doesn't change if you iterate the fields in a different order...

sure, Im working on that :)

cppcache/src/PdxTypeRegistry.hpp
```diff
@@ -53,7 +53,20 @@ typedef std::unordered_map<std::shared_ptr<PdxSerializable>,
...
+  }
+};
+
+typedef std::unordered_map<std::shared_ptr<PdxType>, int32_t, PdxTypeHashCode, PdxTypeEqualCmp>
```
pivotal-jbarrett, line 69: If you implement `std::hash<PdxType>` and `PdxType::operator==` then you can use the `dereference_hash<std::shared_ptr<PdxType>>` and `dereference_equal_to<std::shared_ptr< PdxType >>>` here.

cppcache/src/PdxType.cpp
pivotal-jbarrett: > > Just to be clear, I didn't mean to suggest we not hash in the classname at all, just outside the for loop. Also for what it's worth, even though using classname is still correct, the only instance of PDX we're aware of that has this problem, i.e. the classname isn't unique, is for JSON instances, where classname is always "__GEMFIRE_JSON".
> 
> oh, I see the confusion: the classname its outside the for loop (line 573). The other classnames you are refering to are the classnames of each field. If we dont include the name of the field and its classname, the hash will be the same for an object with a bool called `foo` and for an object with a string called `foo`.

I don't think the Java implementation will distinguish one OBJECT from anther OBJECT . I think you just need to has the field name and type and ignore the class name.

cppcache/src/PdxType.cpp
```diff
@@ -557,11 +557,30 @@ bool PdxType::Equals(std::shared_ptr<PdxType> otherObj) {
...
+
+  if ((m_geodeTypeId == 0) && (other.m_geodeTypeId == 0)) {
+    return this->m_className < other.m_className;
+  }
+
+  return false;
+}
+
+int32_t PdxType::hashcode() const {
+  std::hash<std::string> strHash;
```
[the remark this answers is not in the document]
alb3rtobr: > Just a thought here. Is it necessary to hash like Java client does in this case? 
No, we just need a hash function to allow identify `PdxType` object that will be assigned the same id by the server. We dont need to know the hash generated by the server.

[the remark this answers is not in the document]
alb3rtobr: > In fact I would just create a specialization of `std::hash<PdxType>`.

I will check this

cppcache/src/PdxTypeRegistry.hpp
pdxcodemonkey: +1, I like this!

cppcache/src/PdxType.cpp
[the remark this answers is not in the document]
pivotal-jbarrett: See the bottom of cppcache/include/geode/CacheableKey.hpp for how we export `std::hash<CacheableKey>` and you can do the same thing for `std::hash<PdxType>`.

cppcache/src/PdxTypeRegistry.hpp
alb3rtobr: Agree, but I dont want to check if the objects are equals, I just want to check if two `PdxType` objects have the same hash.

Two `PdxType` objects should be equals if they have the same classname and their fields have the same types, names and values.
And two keys of the `pdxTypeToTypeId` map will be equals if they have the same hash, which means they have the same classname, and their fields have the same types and names (values are not checked here).

alb3rtobr: Sorry, I read again, I got your point, two objects could generate the same hash being different :)

alb3rtobr:
Could you help me with the tests? I have not been able to compile them, there is a linker error saying there is an undefined reference to `PdxType` but I dont know how to change the cmake config to fix it. Thanks in advance!

pdxcodemonkey:
Sure, I'll have a look

cppcache/integration/test/PdxJsonTypeTest.cpp
```diff
@@ -142,4 +146,91 @@ TEST(PdxJsonTypeTest, testTwoConsecutiveGets) {
...
+  apache::geode::client::PdxType m_pdxType1(pdxTypeRegistry, gemfireJsonClassName,false);
+  apache::geode::client::PdxType m_pdxType2(pdxTypeRegistry, gemfireJsonClassName,false);
+
+  m_pdxType1.addVariableLengthTypeField("bar0", "string", PdxFieldTypes::STRING);
+  m_pdxType1.addVariableLengthTypeField("bar1", "string", PdxFieldTypes::STRING);
+
+  m_pdxType2.addVariableLengthTypeField("bar2", "string", PdxFieldTypes::STRING);
+  m_pdxType2.addVariableLengthTypeField("bar3", "string", PdxFieldTypes::STRING);
+
+  EXPECT_NE(std::hash<PdxType>(m_pdxType1),std::hash<PdxType>(m_pdxType2));
```
pivotal-jbarrett, line 204: I think you problem might be that `PdxType` wasn't imported into the current namespace like many of the other types. Notice lines 195 and 196 reference it with the fully qualified name and this line does not. Add a `using` statement at line 54 and then you should be good, and you can drop the fully qualified names on 195, 196.

alb3rtobr: Thanks for the suggestion, but it did not solve the problem. I used the fully qualified name in the lines you mention because there is other PdxType class imported used in the namespace.

pivotal-jbarrett: Oh yeah forgot about that test object. It but me a few times too and thought it should be renamed. I assumed you tried the fully qualified name on the line 204 too?

alb3rtobr: Thank to your comment I have realized the assertions are wrong. Thing is that in the assertions I was using `m_pdxType1.hashcode()` instead of `std::hash<PdxType>(m_pdxType1)`, that was a change I did right before adding the commit with the tests to the PR and its causing the compilation to fail in other place different than the linker error I was asking for. I have just added the correct code of the assertions.

pdxcodemonkey: After commenting out the using statement for PdxTest::PdxType, I found the declaration of std::hash wasn't correct in the code (trying to pass 1 param to ctor).  the following change worked for me on line 183 (the first spot I was hitting a compiler gripe):

-   EXPECT_EQ(std::hash<PdxType>(m_pdxType1),std::hash<PdxType>(m_pdxType2));
+   EXPECT_EQ(std::hash<PdxType>{}(m_pdxType1), std::hash<PdxType>{}(m_pdxType2));

I expect the rest are similar...

alb3rtobr: The first linker error is the following:

```
PdxJsonTypeTest.cpp:166: undefined reference to `apache::geode::client::PdxType::PdxType(apache::geode::client::PdxTypeRegistry&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool)'
```

alb3rtobr: > the following change worked for me on line 183 (the first spot I was hitting a compiler gripe):

Nice, its simpler that the code I have just added. Now with the current code, you should see the linker error I was not able to solve:

```
PdxJsonTypeTest.cpp:166: undefined reference to `apache::geode::client::PdxType::PdxType(apache::geode::client::PdxTypeRegistry&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool)'
... 
and more linker errors from this point
```

pdxcodemonkey: And now I think we see why PdxTests::PdxType existed in the first place.  apache::geode::client::PdxType is hidden in the library.  The following fixes linker errors for me, in cppcache/src/PdxType.hpp:
```
- class PdxType
+ class APACHE_GEODE_EXPORT PdxType
```
But now we need to have the more philosophical question about whether or not this is the right thing to do.  If we _are_ going to export _PdxType_, the header needs to move from cppcache/src to cppcache/include/geode.  @pivotal-jbarrett any opinion/suggestion?

cppcache/src/PdxType.cpp
```diff
@@ -568,16 +568,32 @@ bool PdxType::operator<(const PdxType& other) const {
...
+
+  if(this->m_noJavaClass != other.m_noJavaClass){
+    return false;
+  }
+
+  if(this->getTotalFields() != other.getTotalFields()){
+    return false;
+  }
+
+  //TODO: Compare m_pdxFieldTypes & other.m_pdxFieldTypes;
```
pivotal-jbarrett: What's with the new TODO?

cppcache/src/PdxType.cpp
```diff
@@ -568,16 +568,32 @@ bool PdxType::operator<(const PdxType& other) const {
...
+}
+
+size_t PdxType::hashcode() const {
   std::hash<std::string> strHash;
-  auto result=strHash(this->m_className);
+  auto result = strHash(this->m_className);
 
-  for (std::vector<std::shared_ptr<PdxFieldType>>::iterator it =
-          m_pdxFieldTypes->begin();
+  for (auto it = m_pdxFieldTypes->begin();
```
pivotal-jbarrett: Can we use a range `for` loop here.

cppcache/src/PdxType.hpp
```diff
@@ -205,10 +205,25 @@ class PdxType : public internal::DataSerializableInternal,
   // This is for PdxType as key in std map.
   bool operator<(const PdxType& other) const;
```
pivotal-jbarrett, line 206: If we aren't going to use the std::map anymore perhaps we should remove this `operator<`.

cppcache/src/PdxType.hpp
```diff
@@ -205,10 +205,25 @@ class PdxType : public internal::DataSerializableInternal,
...
 }  // namespace apache
 
+namespace std {
+
+template <>
+struct hash<apache::geode::client::PdxType> {
+  typedef apache::geode::client::PdxType argument_type;
+  typedef size_t result_type;
+  result_type operator()(const argument_type& val) const {
+    return val.hashcode();
```
pivotal-jbarrett, line 223: Technically we can take the `hashcode` method off the `PdxType` and line it here. The hash is only used for `std::hash` so let's not split it across the two classes.

cppcache/src/PdxType.cpp
alb3rtobr: I wanted to separate in a different commit the check of the fields, as I was not sure how to do it. I was exploring the possibility to change `m_pdxFieldTypes` to be a set instead of a vector, for easy comparision of values, but that change had a lot of impacts. After that I realized I could use the `m_fieldNameVsPdxType` map for the comparision of the fields. Thinking about the order of the fields, better to compare two maps than two vectors. Next commit after this one removes this TODO

cppcache/src/PdxType.hpp
alb3rtobr: done!

cppcache/src/PdxType.hpp
alb3rtobr: done

cppcache/src/PdxType.cpp
alb3rtobr: done

cppcache/integration/test/PdxJsonTypeTest.cpp
alb3rtobr: Thanks! With this change Im finally able to verify that the tests are fine.

pdxcodemonkey:
As an experiment, I made the following change this morning:

```
TEST(PdxJsonTypeTest, testTwoObjectsWithSameFieldsHaveTheSameHash) {
  //  Cluster cluster{LocatorCount{1}, ServerCount{1}};
  //  cluster.start();
  //  cluster.getGfsh()
  //      .create()
  //      .region()
  //      .withName("region")
  //      .withType("REPLICATE")
  //      .execute();
  //
  //  auto cache = cluster.createCache();
  //
  //  auto properties = std::make_shared<Properties>();
  //  CacheImpl cacheImpl(&cache, properties, false, false, nullptr);

  PdxTypeRegistry pdxTypeRegistry(nullptr);
  (...)
```
Since the new tests don't hit any code that dereferences the type registry, this worked, and the test(s) passed.  This would allow the hash tests to be moved to the _unit_ test suite, where you have access to internal headers, and we wouldn't have to expose `apache::geode::client::PdxType` from the library.  @pivotal-jbarrett , thoughts?

pivotal-jbarrett:
Good catch. I wasn’t paying attention to the fact that was in an integration test. Yes moving the hash testing to a unit test is the correct move.

cppcache/src/PdxType.hpp
```diff
@@ -202,11 +202,38 @@ class PdxType : public internal::DataSerializableInternal,
 
   bool Equals(std::shared_ptr<PdxType> otherObj);
 
-  // This is for PdxType as key in std map.
-  bool operator<(const PdxType& other) const;
+  bool operator==(const PdxType& other) const;
+
+  size_t hashcode() const;
```
pivotal-jbarrett, line 207: Now that `std::hash` has the implementation is it necessary to have this on the class declaration? I don't see it in defined in the cpp file anymore.

cppcache/src/PdxType.cpp
```diff
@@ -556,12 +556,23 @@ bool PdxType::Equals(std::shared_ptr<PdxType> otherObj) {
   return true;
 }
 
-bool PdxType::operator<(const PdxType& other) const {
-  auto typeIdLessThan = this->m_geodeTypeId < other.m_geodeTypeId;
-  auto typeIdsBothZero =
-      (this->m_geodeTypeId == 0) && (other.m_geodeTypeId == 0);
-  auto classnameLessThan = this->m_className < other.m_className;
-  return (typeIdLessThan || (typeIdsBothZero && classnameLessThan));
+bool PdxType::operator==(const PdxType& other) const {
```
pivotal-jbarrett, line 559: How does this differ from the `PdxType::Equals` method? Is there a purpose for the `PdxType::Equals` anymore?

cppcache/test/CMakeLists.txt
```diff
@@ -55,7 +56,7 @@ add_executable(apache-geode_unittests
   util/synchronized_setTest.cpp
   util/TestableRecursiveMutex.hpp
   util/chrono/durationTest.cpp
-)
+        PdxTypeTest.cpp)
```
pivotal-jbarrett, line 59: Looks like we are adding this file twice.

cppcache/src/PdxType.hpp
alb3rtobr: Good catch, im removing it

cppcache/test/CMakeLists.txt
alb3rtobr: Right, CLion added it for me and it added it manually

pdxcodemonkey: We should check and see if `tests/cpp/testobject/PdxType.*` are in use anywhere.  We may get lucky and find that they can be deleted.

pivotal-jbarrett: > We should check and see if `tests/cpp/testobject/PdxType.*` are in use anywhere. We may get lucky and find that they can be deleted.

It's used in several old and some new integration tests.

cppcache/src/PdxType.cpp
```diff
@@ -537,25 +537,6 @@ void PdxType::generatePositionMap() {
   }
 }
 
-bool PdxType::Equals(std::shared_ptr<PdxType> otherObj) {
```
pivotal-jbarrett, line 540: Yay for deleted code!

cppcache/src/PdxType.cpp
alb3rtobr: I think we can get rid of `PdxType::Equals`

pivotal-jbarrett:
@alb3rtobr If you are struggling with formatting make sure your `clang-format` is v6. Unfortunately the versions of `clang-format` disagree on some things.

alb3rtobr:
> @alb3rtobr If you are struggling with formatting make sure your `clang-format` is v6. Unfortunately the versions of `clang-format` disagree on some things.

Thanks! That was the problem!

alb3rtobr:
I run the bechmark test I implemented as:

`$ ./cpp-benchmark --benchmark_filter=PdxTypeBM_createInstance/real_time`

The results in develop branch were:

```
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
PdxTypeBM_createInstance/real_time   69900929 ns     55007314 ns            9
```

While with the content of this PR the result was 6x faster :

```
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
PdxTypeBM_createInstance/real_time   11560270 ns     10884860 ns           50

```

cppcache/benchmark/CMakeLists.txt
```diff
@@ -19,12 +19,15 @@ add_executable(cpp-benchmark
...
-  )
+  PdxTypeBM.cpp)
+
+#add_dependencies(integration-framework)
 
 target_link_libraries(cpp-benchmark
   PUBLIC
     apache-geode-static
     benchmark::benchmark
+    integration-framework
```
pivotal-jbarrett, line 30: Don't add this here. There is an cppcache/integration/benchmark already. Move the benchmark to the integration benchmark directory.

alb3rtobr: done!

alb3rtobr:
Results using the new location of the bm test case:

`$ ./cpp-integration-benchmark --benchmark_filter=PdxTypeBM_createInstance/real_time`

Before this PR:
```
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
PdxTypeBM_createInstance/real_time   80933776 ns     64921880 ns            8
```

After the PR:
```
-----------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-----------------------------------------------------------------------------
PdxTypeBM_createInstance/real_time   15614725 ns     14725105 ns           39
```

cppcache/integration/benchmark/PdxTypeBM.cpp
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <benchmark/benchmark.h>
+
```
pivotal-jbarrett, line 19: This path seems incorrect.

alb3rtobr: Sorry but Im not able to see any error, could you point me to the issue? Is your comment about the first include?

alb3rtobr: Im going to change the second include to `<framework/Cluster.h>`, maybe thats what you are referring to

alb3rtobr:
@pivotal-jbarrett & @pdxcodemonkey thanks for your help with this PR!

PR #620, opened 2020-06-16
GEODE-8262: Replace insensitive terms in dunit framework

pdxcodemonkey:
master/slave --> primary/secondary

@mreddington @dihardman @davebarnes97 @karensmolermiller @moleske


pdxcodemonkey:
Turns out we didn't get all of these.  Also, after a little more consideration, we liked coordinator/worker a little bit better

PR #623, opened 2020-06-25
GEODE-8297: fix tcrmessage timeout

pdxcodemonkey:


cppcache/src/TcrMessage.cpp
```diff
@@ -124,7 +124,7 @@ TcrMessage::TcrMessage()
       m_chunkedResult(nullptr),
       m_keyList(nullptr),
       m_region(nullptr),
-      m_timeout(15 /*DEFAULT_TIMEOUT_SECONDS*/),
+      m_timeout(DEFAULT_TIMEOUT_SECONDS),
```
pivotal-jbarrett, line 127: Could you correct the name of the constant. Given it is now a `std::chrono::duration` the time unit qualifier on the end of the constant name doesn't make sense.

PR #625, opened 2020-07-03
GEODE-8340:  Enforce Switch compiler warnings as errors

moleske:
~~Still running some tests so leaving as draft.  Also want to check formatting with the pr since I don't feel like installing clang 6 on my machine~~

Seems to be passing.  Also seemed reasonable to do these three related compiler warnings as one


cppcache/include/geode/DataInput.hpp
```diff
@@ -313,7 +313,53 @@ class APACHE_GEODE_EXPORT DataInput {
         // empty string
         break;
       // TODO: What's the right response here?
-      default:
+      case internal::DSCode::FixedIDDefault:
```
pdxcodemonkey, line 316: Wow I really really don't like this.  Is there anything we can do to fix the warning _besides_ adding ~65 case statements?  This makes the code less readable, not more IMO.

cppcache/src/ExceptionTypes.cpp
```diff
@@ -297,7 +297,25 @@ const std::string& getThreadLocalExceptionMessage();
       PutAllPartialResultException ex(message);
       throw ex;
     }
-    default: {
+    case GF_NOERR:
```
pdxcodemonkey, line 300: Same as above - adding a zillion case statements doesn't make code clearer.

cppcache/src/TcrMessage.cpp
```diff
@@ -1015,6 +1015,36 @@ void TcrMessage::processChunk(const std::vector<uint8_t>& chunk, int32_t len,
         break;
       }
       // fall-through for other cases
+      if (m_chunkedResult != nullptr) {
```
pdxcodemonkey, line 1018: What's this block of code doing?  I don't immediately see a place where it was removed/moved, so is it new?

moleske: The previous code did a fallthrough if it didn't make it to the `break;` statement on line 999 or 1015 (which is buried in an `if` and `if else`).  This is the same code as on line 1054.  I thought about extracting a function but got lazy

cppcache/include/geode/DataInput.hpp
moleske: I actually like it since it forces you to acknowledge all possible switches for an enum (which is enumerable).  That way you don't accidentally go to the default when adding a new enum.  If the general consensus is we prefer not to change `-Wno-switch-enum` then we can revert that commit.  I could make it worse by adding `break;` statements in between all the cases :-p

pdxcodemonkey: Sorry, I don't think the chance of getting a warning for free is worth the cost of some egregious cluttering of the code.  We're only using 4 of the 61 possible values for a DSCode, so it was kind of an abuse of a switch statement to begin with.  This should probably just be a 4-way if-else block.

moleske: I'll switch it to an if-else block and we'll see if that cleans up better

pdxcodemonkey: Just gonna leave this here, cause I spent the time & effort to figure it out.  Probably slightly better performance than the if-else approach, but I'm still not sure about readability:

```
  template <class CharT = char, class... Tail>
  inline std::basic_string<CharT, Tail...> readString() {
    std::basic_string<CharT, Tail...> value;
    std::map<internal::DSCode, std::function<void(std::string&)> > readers;

    readers.insert(std::make_pair(
        internal::DSCode::CacheableString,
        [=](std::string& val) { this->readJavaModifiedUtf8(val); }));

    readers.insert(
        std::make_pair(internal::DSCode::CacheableStringHuge,
                       [=](std::string& val) { this->readUtf16Huge(val); }));

    readers.insert(
        std::make_pair(internal::DSCode::CacheableASCIIString,
                       [=](std::string& val) { this->readAscii(val); }));

    readers.insert(
        std::make_pair(internal::DSCode::CacheableASCIIStringHuge,
                       [=](std::string& val) { this->readAsciiHuge(val); }));

    auto type = static_cast<internal::DSCode>(read());
    auto it = readers.find(static_cast<internal::DSCode>(read()));

    if (it != readers.end()) {
      it->second(value);
    }
    return value;
  }

```

moleske: I checked the `Allow edits and access to secrets by maintainers` so feel free to add this as commit and see if it works (I'm assuming the commit will go to the branch and trigger the [CI pipeline](https://github.com/moleske/geode-native/actions) I've been playing around with)

moleske: Per some other avenue feedback from @pdxcodemonkey and @mreddington , I'll take some time to change this to a 4 way `if-else if` since it seems this is not much better.  For the latter list in `ExceptionTypes.cpp`, I'll try a map approach (similar to `start_element_map_` and `end_element_map_` in CacheXmlParser

cppcache/src/ExceptionTypes.cpp
moleske: Let me know what you think of the change @pdxcodemonkey .  I have mixed feelings about it...

cppcache/integration-test/testThinClientSecurityAuthorizationMU.cpp
```diff
@@ -160,6 +160,7 @@ void initClientAuth(char UserType) {
     case 'A':
       credentialGeneratorHandler->getAllowedCredentialsForOps(ad, userCreds,
                                                               nullptr);
+      break;
     default:
```
pivotal-jbarrett, line 164: What's with the use of `default` here but in other places it is replaced with explicit `case` statements?

moleske: `default` is used here because this switch statement is checking string characters.  The others are checking an enum.  Since an enum has a well defined set of values, you do not need a default if you have a case statement for each value.  `no-covered-switch-default` and `no-switch-enum` help enforce that

cppcache/src/ExceptionTypes.cpp
moleske: pdxcodemonkey?

(I mostly just wanted to post that gif, no particular rush 😃)

moleske:
@pdxcodemonkey still no rush, but I had to resolve a conflict `cppcache/integration/framework/TestConfig.cpp.in`.  I'd like to merge this is in sooner rather than later.  Let me know if there should be more changes, or if more Geode Jira stories should be made to resolve simplifying things that shouldn't be enums in the future

moleske:
Per conversation with @pdxcodemonkey on the [asf slack](https://the-asf.slack.com/archives/C4RJQAQJF/p1598649633002500?thread_ts=1598634493.001700&cid=C4RJQAQJF), went ahead and create [JIRA-8468](https://issues.apache.org/jira/browse/GEODE-8468) for enhancements to replace the couple of places with large enum switch statements with better control structures.  I put the "GoodForNewContributors" label on it since I think this might be doable with only knowledge of control structures and less of internal of geode native, but feel free to take label off if you think it is more complex.  And also update the description in the Jira if I missed something or it is unclear

PR #628, opened 2020-07-15
GEODE-8344: Add GatewaySenderEventCallbackArgument class

alb3rtobr:
It was observed that when having two sites, with a client connected to each site, there were serialization errors in the clients when handling received events. The cause is that GatewaySenderEventCallbackArgument objects are sent as part of the events, but this class is not known by the C++ client.
This PR allows the C++ clients to deserialize the class.

Using the same setup with using two java clients, I have observed that the GatewaySenderEventCallbackArgument objects are received and deserialized correctly, but it seems they are not used. This is why this PR is not including any other code related with the GatewaySenderEventCallbackArgument in the C++ clients.

### DIFF
--- a/cppcache/integration/test/Order.cpp
+++ b/cppcache/integration/test/Order.cpp
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Order.hpp"
+
+#include <geode/PdxReader.hpp>
+#include <geode/PdxWriter.hpp>
+
+namespace testGeode8344 {
+
+void Order::fromData(PdxReader& pdxReader) {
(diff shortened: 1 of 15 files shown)


alb3rtobr:
Sorry for the late answer, I have been on vacation. I have changed the PR to draft because Im still working on the tests. I have added some functionality to the integration test framework in order to reproduce configure two clusters to reproduce the error, but I still have a problem with a disk access exception when configuring gateway senders.

alb3rtobr:
With the new integration test it is possible to reproduce the problem and ensure it is solved.

These two logs messages can be found on test log:
```
Region::put: region [/region] created key [order], value [OrderID: 2 Product Name: product y Quantity: 37]
Region::put: region [/region] created key [order], value [PDX[28743191,com.example.Order]{name=product y,order_id=2,quantity=37}]
```
Without the GatewaySenderEventCallbackArgument class added to the code, there is a deserialization exception when receiving the second event, so instead of that message, this is one appears:
```
Exception while receiving subscription event for endpoint localhost:33829:: apache::geode::client::IllegalStateException: Unregistered type in deserialization
```
As the event generated in cluster A is not deserialized correctly in cluster B, it is not processed by the dummy cache listener implemented, and the test case will fail with this error:
```
 Expected equality of these values:
  cacheListenerB->getNumEvents()
    Which is: 0
  1
```

cppcache/integration/test/Order.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Order.hpp"
+
+#include <geode/PdxReader.hpp>
+#include <geode/PdxWriter.hpp>
+
+namespace testGeode8344 {
```
moleske, line 23: I would consider picking a namespace name that is more descriptive.  I think a few months from now (or for me, a few weeks) it will not be obvious what Geode 8344 means without having to go look up it up in jira

alb3rtobr:
Hi @moleske @echobravopapa @pivotal-jbarrett , is there anything else missing to approve this PR?

alb3rtobr:
thanks for the reviews! Sorry but I need someone to merge the PR because Im not a committer

PR #629, opened 2020-07-16
GEODE-8364: Change log level at runtime

alb3rtobr:


pdxcodemonkey:
Do you have a use case for this?  I don't suppose it's really harming anything, but I'm not sure I understand why/how you'd want to change this midstream, and it could cause problems down the road for things like log analysis tools.  We've also been planning for a long time to completely replace our logging mechanism with an off-the-shelf solution like spdlog or log4cpp, and this would add a new logging feature that we'd have to replicate with any such change.

alb3rtobr:
It will be useful for us to have this feature in order to get more info when there is a problem with the client. If we detect an issue in the client, we can activate debug log to get more info about what is happening. This will help a lot our engineers to troubleshoot and report problems.

cppcache/include/geode/Cache.hpp
```diff
@@ -239,6 +241,10 @@ class APACHE_GEODE_EXPORT Cache : public GeodeCache {
 
   SystemProperties& getSystemProperties() const override;
 
+  void setLogLevel(LogLevel newLevel);
```
pivotal-jbarrett, line 244: Please add documentation to these new methods.

cppcache/test/CacheTest.cpp
```diff
@@ -61,3 +62,10 @@ TEST(CacheTest, close) {
   EXPECT_THROW(cache.readyForEvents(), CacheClosedException);
   EXPECT_THROW(cache.rootRegions(), CacheClosedException);
 }
+
+TEST(CacheTest, changeLogLevel) {
+  auto cache = CacheFactory{}.set("log-level", "info").create();
+  ASSERT_EQ(cache.getLogLevel(), LogLevel::Info);
```
pivotal-jbarrett, line 68: What happens if you call `cache.getSystemProperties.getLogLevel()` after changing it via `Cache`?

pdxcodemonkey: Exactly - I was thinking about this earlier this morning.  We can still get into the kind of weird situation where the `geode.properties` file says `log-level` is one value but in fact it's another, and the `SystemProperties` object is also in disagreement about `log-level`.  I don't think the first situation is worth worrying about - unless I'm very mistaken, you'll be in this state even when you set `log-level` on the cache factory to a custom value.  The second, however, it seems to me we should be able to prevent.

cppcache/include/geode/Cache.hpp
alb3rtobr: done

cppcache/test/CacheTest.cpp
alb3rtobr: I have changed the `setLogLevel` method to also update the log level on the `SystemProperties` object.

alb3rtobr:
@pivotal-jbarrett could you please check if the changes are ok for you? Thanks

PR #632, opened 2020-08-12
GEODE-8426: Enforce warning no-shorten-64-to-32

moleske:

curious if this is needed since quick test on my machine was compiling (so figured I'd make travis check real quick)

edit - will make a Jira when I bother to log into Jira next time


moleske:
> Looks like it got fixed along the way!
It's nice when that happens!  Only wish I could give the right people credit!

PR #633, opened 2020-08-13
GEODE-8390: Geode Native Client guide - expose "Connection Pools" in nav

davebarnes97:
Updated left-hand nav in both .NET and C++ guides to include "Using Connection Pools" and its subtopics
Made basic changes to text for language-specific books. Thorough update is the subject of another ticket.


davebarnes97:
> Can you please update the file connection-pools/running-connection-pool-code.html in the .NET version only to remove the C++ example? I see 2 examples. One is C++ and the other is .NET.

Leaving C++ in the .NET example was intentional. You’re right that most .NET users would work in C#, but the .NET implementation does support both languages.

PR #634, opened 2020-08-17
GEODE-8398: Add SNI support to .NET API 

mmartell:
This adds a new .NET API to the nativeclient to support SNI.

Notes:

- This ticket usurps GEODE-8156. Since we need the new SetSniProxy API to run the test, we'll use this GEODE-8398 for the pull request.
- The SNI tests are currently disabled on Windows platforms due to issues with running docker in the CI.


clicache/integration-test2/SNITests.cs
```diff
@@ -0,0 +1,127 @@
...
+            d.WaitForExit();
+        }
+
+        public void Dispose()
+        {
+
+            var dockerComposeProc = Process.Start(@"C:\Program Files\docker\docker\resources\bin\docker-compose.exe", "stop");
+            dockerComposeProc.WaitForExit();
+
+            var dockerProc = Process.Start(@"C:\Program Files\docker\docker\resources\bin\docker.exe", "container prune -f");
```
pivotal-jbarrett, line 64: There is a Docker C++ SDK we could use rather than forking an executable.

clicache/integration-test2/SNITests.cs
```diff
@@ -0,0 +1,127 @@
...
+                              .Create<string, string>("jellyfish");
+
+            region.Put("1", "one");
+            var value = region.Get("1");
+
+            Assert.Equal("one", value);
+            cache_.Close();
+        }
+
+        [Fact (Skip = "Docker nut supported in Windows CI")]
```
pivotal-jbarrett, line 112: Perhaps you mean "not" rather than "nut". Also that statement is false, Docker is fully supported in Windows so why are we not skipping it?

clicache/src/PoolFactory.hpp
```diff
@@ -279,6 +279,20 @@ namespace Apache
...
+        /// <remarks>
+        /// Used for connecting via SNI proxy.
+        /// </remarks>
+        /// <param>
+        /// host the host name or ip address that the server is listening on.
+        /// </param>
+        /// <param>
+        /// port the port that the server is listening on.
+        /// </param>
+        PoolFactory^ SetSniProxy(String^ hostname, Int32 port);
```
pivotal-jbarrett, line 294: Deviates from [SNI Proxy RFC](https://cwiki.apache.org/confluence/display/GEODE/Client+side+configuration+for+a+SNI+proxy).

cppcache/include/geode/PoolFactory.hpp
```diff
@@ -426,6 +426,11 @@ class APACHE_GEODE_EXPORT PoolFactory {
    */
   PoolFactory& addServer(const std::string& host, int port);
 
+  /**
+   * Set proxy info for SNI connection.  Used for connecting via SNI proxy.
+   */
+  PoolFactory& setSniProxy(const std::string& hostname, const int port);
```
pivotal-jbarrett, line 432: Deviates from [SNI Proxy RFC](https://cwiki.apache.org/confluence/display/GEODE/Client+side+configuration+for+a+SNI+proxy).

cppcache/integration/test/SNITest.cpp
```diff
@@ -108,41 +112,51 @@ class SNITest : public ::testing::Test {
...
+                    .create("jellyfish");
 
   region->put("1", "one");
 
   cache.close();
 }
 
-TEST_F(SNITest, connectionFailsTest) {
+#if defined(_WIN32)
+TEST_F(SNITest, DISABLE_connectWithoutProxyFails) {
```
pivotal-jbarrett, line 154: Why is this test disabled on Windows?

cppcache/integration/test/sni-test-config/geode-config/truststore_sni.pem
```diff
@@ -0,0 +1,68 @@
+-----BEGIN CERTIFICATE-----
```
pivotal-jbarrett, line 1: Why do we use a pre-baked key and certificates here? I though all the SSL tests generated certs on the fly but maybe I am wrong. In the past this has caused issues with hostname validation.

cppcache/src/PoolAttributes.cpp
```diff
@@ -44,46 +44,13 @@ PoolAttributes::PoolAttributes()
...
-      m_serverGrp(PoolFactory::DEFAULT_SERVER_GROUP) {}
+      m_serverGrp(PoolFactory::DEFAULT_SERVER_GROUP),
+      m_sniProxyPort(0) {}
+
 std::shared_ptr<PoolAttributes> PoolAttributes::clone() {
   return std::make_shared<PoolAttributes>(*this);
 }
 
-/** Return true if all the attributes are equal to those of other. */
-bool PoolAttributes::operator==(const PoolAttributes& other) const {
```
pivotal-jbarrett, line 53: Why did we get rid of this method to compare two pool attributes?

cppcache/src/TcpSslConn.cpp
```diff
@@ -87,15 +97,13 @@ void TcpSslConn::connect() {
...
-  // m_ssl->init();
-
   std::chrono::microseconds waitMicroSeconds = m_waitMilliSeconds;
 
   LOGDEBUG("Connecting SSL socket stream to %s:%d waiting %s micro sec",
            m_addr.get_host_name(), m_addr.get_port_number(),
            to_string(waitMicroSeconds).c_str());
 
-  int32_t retVal = m_ssl->connect(m_addr, waitMicroSeconds);
+  int32_t retVal = m_ssl->connect(m_addr, m_sniHostname, waitMicroSeconds);
```
pivotal-jbarrett, line 106: `auto`

cppcache/src/TcpSslConn.hpp
```diff
@@ -67,14 +81,18 @@ class TcpSslConn : public TcpConn {
              const char* privkeyfile, const char* pemPassword)
       : TcpConn(ipaddr, waitSeconds, maxBuffSizePool),
         m_ssl(nullptr),
+        m_sniPort(0),
+        m_sniHostname(""),
```
pivotal-jbarrett, line 85: Initializing a `std::string` with `""` is redundant.

cppcache/src/ThinClientLocatorHelper.hpp
```diff
@@ -71,6 +74,8 @@ class ThinClientLocatorHelper {
   const ThinClientPoolDM* m_poolDM;
   ThinClientLocatorHelper(const ThinClientLocatorHelper&);
   ThinClientLocatorHelper& operator=(const ThinClientLocatorHelper&);
+  std::string m_sniProxyHost;
```
pivotal-jbarrett, line 77: If the set of member variable is small perhaps its time to make them conform to the style guide. (optional)

sni-test-config/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+version: '3'
+services:
+  geode:
+    container_name: 'geode'
+    image: 'apachegeode/geode'
+    expose:
+      - '10334'
+      - '40404'
+    entrypoint: 'sh'
+    command: ["-c", "while true; do sleep 600; done"]
```
pivotal-jbarrett, line 26: If you aren't using the forever script then why include the script in the check in?

clicache/integration-test2/SNITests.cs
echobravopapa: not that we have found, unfortunately...

clicache/integration-test2/SNITests.cs
echobravopapa: its Docker on GCP that is specifically the issue, but thanks for catching the typo

cppcache/integration/test/SNITest.cpp
echobravopapa: aforementioned Docker issues...

sni-test-config/docker-compose.yml
echobravopapa: good point, that can be removed

cppcache/integration/test/sni-test-config/geode-config/truststore_sni.pem
echobravopapa: you are probably thinking of the server side...

cppcache/src/PoolAttributes.cpp
pdxcodemonkey: At first I refactored it, per our typical practice, because I was in modifying things there already.  Then I checked to see if it was actually in use anywhere, and discovered it was not, so out it went.

cppcache/src/ThinClientLocatorHelper.hpp
pdxcodemonkey: Changes coming to fix this.  We changed the new members to conform to style guide after rebasing on top of your "cryptoimpl removal" change.

clicache/src/PoolFactory.hpp
echobravopapa: Native impl of SNI is consistent with Native's libraries... Java client impl of SNI simply doesn't map to this API.
Our understanding is that the light-weight RFC process exists to avoid issues post implementation etc or any surprises... Native's API for SNI does differ from the Java client, but the core feature etc is in line.

cppcache/include/geode/PoolFactory.hpp
echobravopapa: ""

clicache/integration-test2/SNITests.cs
pivotal-jbarrett: https://github.com/lasote/docker_client

clicache/integration-test2/SNITests.cs
pivotal-jbarrett: I am running Docker on Windows in GCP just fine. Can I help?

clicache/src/PoolFactory.hpp
pivotal-jbarrett: A difference in API is a pretty big surprise to some. Especially when you consider that .NET does expose a Socket concept in the SDK similar to Java. Having a discussion beforehand as to why one API would deviate from another established in an RFC helps reviewers understand the rational for the deviation.

I can only make the assumption that the deviation is a result of the C++ layer not wanting to leak `ACE::Socket` outside the internals. There are alternative ways to do that such that you have an API similar to the Java API that is extensible for future proxy implementations or parameters to the proxy connections itself without exposing the internals of the socket implementation. By not having this conversation we have implemented something that solves today's issue but will need to be deprecated or augmented to support future changes to the proxy strategy.

echobravopapa: not leaking anything from `ACE` out to the public API is important - especially since we are close to removing all of `ACE` from this code base...

pivotal-jbarrett: So why not provide, like the Java API, and interface that can take pre-canned `SocketFactory` classes. At this point the only deviation would be the ability for users to provide their own factories?

cppcache/integration/test/sni-test-config/geode-config/truststore_sni.pem
pivotal-jbarrett: Is there a technical reason we can’t do it for the client too?

echobravopapa: technically possible, just a resource constraint

clicache/integration-test2/SNITests.cs
echobravopapa: thx! we'll take a look at this for future improvements, current deadline isn't going to allow

clicache/src/PoolFactory.hpp
pivotal-jbarrett: Offline conversation: The team is comfortable with the limitations of this API in the effort to get something released and are comfortable with potentially deprecating this API in favor of something more flexible and in line with the RFC when and if expansion of this feature is necessary.

clicache/integration-test2/SNITests.cs
moleske: Still, it is odd to hard code the docker executable paths.

pivotal-jbarrett: Oh good catch. I didn't even notice that initially over the whole, why not use the SDK thing. If the absolute path is necessary, which I doubt it is, then use CMake `find_program` and inject. It should be sufficient to just make sure docker is in the PATH or add it to the PATH in the CTest definition.

cppcache/src/TcpSslConn.cpp
pdxcodemonkey: This code will go away when I finish rebasing this branch from the latest, i.e. bringing in the change to remove cryptoimpl.

cppcache/src/TcpSslConn.hpp
pdxcodemonkey: Fixed

echobravopapa:
@pivotal-jbarrett latest mergable is up...

PR #635, opened 2020-08-19
GEODE-8436: Several threads calling PdxInstanceFactory::create() causes seg fault

alb3rtobr:
A segmentation fault is produced when `PdxInstanceFactory::create()` is called by several threads that are registering the same new pdx type.

The core is produced here:
```
void PdxInstanceImpl::toDataMutable(PdxWriter& writer) {
   auto pt = getPdxType();
   std::vector<std::shared_ptr<PdxFieldType>>* pdxFieldList =
       pt->getPdxFieldTypes();
```
`getPdxType()` returns `nullptr`, so in the next line, there is segmentation fault when calling `pt->getPdxFieldTypes()`.

I have seen the problem appears when `PdxInstanceImpl::getPdxType()` calls `PdxTypeRegistry::getPdxType()` and this method returns `nullptr` due to it is not able to found an entry in its `typeIdToPdxType_` map. But I have verified that at that point, the corresponding entry in `pdxTypeToTypeId_` map exists, so the maps were not aligned.

I checked how they are updated, and I realized that two different locks are being used to update `pdxTypeToTypeId_` map first and `typeIdToPdxType_` after. So it seems the error is caused due to `typeIdToPdxType_` is checked before it is updated.

The test case implemented does not fail every time it is executed (I have not been able to find a configuration to make it fail always) but I have not seen it failing after the change I have introduced in PdxTypeRegistry.

### DIFF
--- a/cppcache/src/PdxTypeRegistry.cpp
+++ b/cppcache/src/PdxTypeRegistry.cpp
@@ -88,7 +88,7 @@ int32_t PdxTypeRegistry::getPDXIdForType(std::shared_ptr<PdxType> nType,
 
     typeId = cache_->getSerializationRegistry()->GetPDXIdForType(pool, nType);
     nType->setTypeId(typeId);
-    pdxTypeToTypeIdMap_.emplace(nType,typeId);
+    pdxTypeToTypeIdMap_.emplace(nType, typeId);
     typeIdToPdxType_.emplace(typeId, nType);
   }
   return typeId;
--- a/cppcache/src/PdxTypeRegistry.cpp
+++ b/cppcache/src/PdxTypeRegistry.cpp
@@ -88,8 +88,8 @@ int32_t PdxTypeRegistry::getPDXIdForType(std::shared_ptr<PdxType> nType,
 
     typeId = cache_->getSerializationRegistry()->GetPDXIdForType(pool, nType);
     nType->setTypeId(typeId);
-    pdxTypeToTypeIdMap_.insert(std::make_pair(nType, typeId));
-    typeIdToPdxType_.insert(std::make_pair(typeId, nType));
+    pdxTypeToTypeIdMap_.emplace(nType,typeId);
+    typeIdToPdxType_.emplace(typeId, nType);
   }
   return typeId;
 }
@@ -118,8 +118,7 @@ void PdxTypeRegistry::clear() {
 void PdxTypeRegistry::addPdxType(int32_t typeId,
                                  std::shared_ptr<PdxType> pdxType) {
   WriteGuard guard(g_readerWriterLock_);
-  std::pair<int32_t, std::shared_ptr<PdxType>> pc(typeId, pdxType);
-  typeIdToPdxType_.insert(pc);
+  typeIdToPdxType_.emplace(typeId, pdxType);
 }
 
 std::shared_ptr<PdxType> PdxTypeRegistry::getPdxType(int32_t typeId) const {
--- a/cppcache/src/PdxTypeRegistry.cpp
+++ b/cppcache/src/PdxTypeRegistry.cpp
@@ -89,8 +89,8 @@ int32_t PdxTypeRegistry::getPDXIdForType(std::shared_ptr<PdxType> nType,
     typeId = cache_->getSerializationRegistry()->GetPDXIdForType(pool, nType);
     nType->setTypeId(typeId);
     pdxTypeToTypeIdMap_.insert(std::make_pair(nType, typeId));
+    typeIdToPdxType_.insert(std::make_pair(typeId, nType));
   }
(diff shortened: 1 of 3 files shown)


alb3rtobr:
Hi @pdxcodemonkey , @pivotal-jbarrett , @moleske . I have not been able to create a test case that fails every time it is executed. Without the fix, the test case I have implemented fails 70-80% of executions. And with the fix, I have not seen it fail yet. Do you think it would be enough to approve the PR?

pivotal-jbarrett:
> Hi @pdxcodemonkey , @pivotal-jbarrett , @moleske . I have not been able to create a test case that fails every time it is executed. Without the fix, the test case I have implemented fails 70-80% of executions. And with the fix, I have not seen it fail yet. Do you think it would be enough to approve the PR?

Yes, I am fine with it. It's pretty clear when you do that math that the original code is wrong from a concurrency standpoint. Good find!

cppcache/src/PdxTypeRegistry.cpp
```diff
@@ -89,8 +89,8 @@ int32_t PdxTypeRegistry::getPDXIdForType(std::shared_ptr<PdxType> nType,
     typeId = cache_->getSerializationRegistry()->GetPDXIdForType(pool, nType);
     nType->setTypeId(typeId);
     pdxTypeToTypeIdMap_.insert(std::make_pair(nType, typeId));
```
pivotal-jbarrett, line 91: Perhaps for readability we dispense with the old `insert` with `std::make_pair` and use `emplace`:
```C++
pdxTypeToTypeIdMap_.emplace(nType, typeId);
typeIdToPdxType_.emplace(typeId, nType);
```

cppcache/src/PdxTypeRegistry.cpp
```diff
@@ -89,8 +89,8 @@ int32_t PdxTypeRegistry::getPDXIdForType(std::shared_ptr<PdxType> nType,
...
+    typeIdToPdxType_.insert(std::make_pair(typeId, nType));
   }
-  addPdxType(typeId, nType);
```
pivotal-jbarrett, line 93: Perhaps it makes sense to remove this method and inline the single use in the `getPDXIdForType` function.

pdxcodemonkey:
Travis CI is failing due to a clang-format issue in the new test file.  Please run `cmake --build . --target all-clangformat` and check in the formatting changes.  Reminder, you will need clangformat v6 for compatibility.

pdxcodemonkey:
I'm also hitting this on Windows builds now, though it appears to have come in with yesterday's PR for GatewaySenderEventCallbackArgument:
`Z:\nativeclient\cppcache\integration\test\WanDeserializationTest.cpp(169): error C3861: 'sleep': identifier not found`
continuing to investigate...

pivotal-jbarrett:
> I'm also hitting this on Windows builds now, though it appears to have come in with yesterday's PR for GatewaySenderEventCallbackArgument:
> `Z:\nativeclient\cppcache\integration\test\WanDeserializationTest.cpp(169): error C3861: 'sleep': identifier not found`
> continuing to investigate...

Not sure how that ever worked given that's not really a standard function.
You want `std::this_thread::sleep_for`.

https://en.cppreference.com/w/cpp/thread/sleep_for

cppcache/src/PdxTypeRegistry.cpp
alb3rtobr: `PdxTypeRegistry::addPdxType()` is used several times in `PdxHelper` class, I cannot be removed. It is also used other time in `PdxTypeRegistry` class, and I think it has sense to keep it for readability: it cannot be replace just by a `typeIdToPdxType_.emplace()` call because it is also needed to acquire the lock, which is done inside `addPdxType()`

cppcache/src/PdxTypeRegistry.cpp
alb3rtobr: done!

alb3rtobr:
> > I'm also hitting this on Windows builds now, though it appears to have come in with yesterday's PR for GatewaySenderEventCallbackArgument:
> > `Z:\nativeclient\cppcache\integration\test\WanDeserializationTest.cpp(169): error C3861: 'sleep': identifier not found`
> > continuing to investigate...
> 
> Not sure how that ever worked given that's not really a standard function.
> You want `std::this_thread::sleep_for`.
> 
> Also, sleeping in a test just makes it flaky.
> 
> https://en.cppreference.com/w/cpp/thread/sleep_for

Ok, Im going to fix in a separate PR.

alb3rtobr:
Just to confirm: the problem for the revert is that the new test case of this PR fails on RHEL platform, or it is that the code changed is making other tests to fail?

PR #636, opened 2020-08-20
Add notes about docker requirement

moleske:

Test `42: SNITest.connectViaProxyTest` requires docker running on your machine, otherwise it fails.  This means docker should be mentioned as a required install now


pivotal-jbarrett:
How about making CMake detect Docker and enable/disable Docker based tests depending on its existence?

moleske:
I think we need both, we should call out that if you really want to run all the tests, you will need docker and we should do some cmake logic to help out

PR #639, opened 2020-08-23
GEODE-8441: AppDomainContext was initialized too late

mmartell:
This fixes a longstanding AppDomain bug that was introduced back in March 2017, wherein the AppDomainContext was being initialized after the cache was created.

I've tested the fix this by utilizing a simple Asp.NET web application running in IIS. Prior to the fix, the exception ("Cannot pass gchandle across AppDomains") appears in the log file. I have seen the exception occur in the following threads: "NC MC Thread" (Manage Connections) and "NC CMDService Thread". With the fix I can no longer generate this exception in IIS or IIS Express.

### DIFF
--- a/clicache/src/CacheFactory.cpp
+++ b/clicache/src/CacheFactory.cpp
@@ -55,6 +55,7 @@ namespace Apache
 
       Cache^ CacheFactory::Create()
       {
+        native::createAppDomainContext = &Apache::Geode::Client::createAppDomainContext;
 				bool pdxIgnoreUnreadFields = false;
         bool pdxReadSerialized = false;
         _GF_MG_EXCEPTION_TRY2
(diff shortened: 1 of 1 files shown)


clicache/src/CacheFactory.cpp
```diff
@@ -70,14 +71,12 @@ namespace Apache
...
           Log::SetLogLevel(static_cast<LogLevel>(native::Log::logLevel( )));
-          native::createAppDomainContext = &Apache::Geode::Client::createAppDomainContext;
-
 
           DistributedSystem::ManagedPostConnect(cache);
           DistributedSystem::registerCliCallback();
           auto&& cacheImpl = CacheRegionHelper::getCacheImpl(nativeCache.get());
           cacheImpl->getSerializationRegistry()->setPdxTypeHandler(new ManagedPdxTypeHandler());
-		  cacheImpl->getSerializationRegistry()->setDataSerializableHandler(new ManagedDataSerializableHandler());
+		      cacheImpl->getSerializationRegistry()->setDataSerializableHandler(new ManagedDataSerializableHandler());
```
pivotal-jbarrett, line 79: Looks like formatting got out of wack with your Visual Studio. Can you back out the change on this lineZ or better yet make the indent match the rest of the file:

mmartell: Fixed to match rest of file.

PR #641, opened 2020-08-26
GEODE-8398: Add sni unittest for dotnet

mmartell:
Since the .NET API is a thin wrapper around the C++ API, all we really need to test in this layer is that the sni parameters are passed properly to the C++ layer.

This PR adds a new unit test in the new google test framework. Also includes new accessor functions to the pool API to get and set the sni parameters.

### DIFF
--- a/clicache/test2/Tests2.cs
+++ b/clicache/test2/Tests2.cs
@@ -127,9 +127,8 @@ namespace Apache.Geode.Client.UnitTests
     public void SetSniProxy()
     {
         PoolFactory poolFactory = _cacheOne.GetPoolFactory()
-                .AddLocator("localhost", 10334);
-
-        poolFactory.SetSniProxy("haproxy", 7777);
+                .AddLocator("localhost", 10334)
+                .SetSniProxy("haproxy", 7777);
 
         Pool pool = poolFactory.Create("testPool");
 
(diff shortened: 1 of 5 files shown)


clicache/test2/Tests2.cs
```diff
@@ -123,7 +123,24 @@ public void RegisterPdxType()
...
     }
 
-    private class DummyPdxSerializer : IPdxSerializer
+    [Fact]
+    public void SetSniProxy()
+    {
+        PoolFactory poolFactory = _cacheOne.GetPoolFactory()
+                .AddLocator("localhost", 10334);
+
+        poolFactory.SetSniProxy("haproxy", 7777);
```
echobravopapa, line 132: I'd like to see this right after `AddLocator` in the typical builder pattern look...

cppcache/src/Pool.cpp
```diff
@@ -45,6 +45,9 @@ std::chrono::milliseconds Pool::getReadTimeout() const {
   return m_attrs->getReadTimeout();
 }
 
+std::string Pool::getSniProxyHost() const { return m_attrs->getSniProxyHost(); }
```
echobravopapa, line 48: formatting looks off here...

clicache/test2/Tests2.cs
mmartell: Good catch.

cppcache/src/Pool.cpp
mmartell: Formatting is consistent with our 80 column limit and other single line functions in this file.

echobravopapa: okay, github just wrapped it, which made it look bad

PR #642, opened 2020-08-26
GEODE-8457: Fix crash when IO_error and single-hop=false

albertogpz:


cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -1428,10 +1428,12 @@ GfErrType ThinClientPoolDM::sendSyncRequest(
           }
           excludeServers.insert(ServerLocation(ep->name()));
           if (error == GF_IOERR) {
-            auto sl = std::make_shared<BucketServerLocation>(ep->name());
-            LOGINFO("Removing bucketServerLocation %s due to GF_IOERR",
-                    sl->toString().c_str());
-            m_clientMetadataService->removeBucketServerLocation(sl);
+            if (m_clientMetadataService != nullptr) {
```
moleske, line 1431: When I see an `if` statement, I usually think there's a possible test case.  One for when the statement is `true`, and one for `false`.  Is it possible to write a test for this and the below `if` statements?

pivotal-jbarrett, line 1431: The one style thing I would point out is that `std::unique_ptr` explicitly converts to `true` if set or `false` if is `nullptr` specifically for the usage in if statements like this.
```c++
if (m_clientMetadataService) {
  m_clientMetadataService->doSomething();
}
```

albertogpz: The m_clientMetadataService member variable is null when single-hop is not enabled.
I have duplicated the test cases I created to verify the fix for GEODE-8231 which had single-hop enabled, to also test the case when single-hop is disabled.

albertogpz: ok

alb3rtobr (approved):

PR #644, opened 2020-08-30
Remove  GemFireVersion.getBuildDate()

moleske:
This [commit](https://github.com/apache/geode/commit/d9d13100ae11e2eba8a6428fad29007388680b43#diff-a377b15d2036a2fc9072c1915c7f0ed4L83) removed getBuildDate() in the java code, causing geode native develop to fail to compile with geode develop

### DIFF
--- a/tests/javaobject/LatestProp.java
+++ b/tests/javaobject/LatestProp.java
@@ -24,7 +24,6 @@ import org.apache.geode.internal.*;
 
 public class LatestProp {
   public static void main(String[] args){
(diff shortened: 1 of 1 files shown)


tests/javaobject/LatestProp.java
```diff
@@ -24,7 +24,6 @@
 
 public class LatestProp {
   public static void main(String[] args){
-    System.out.println("build.date=" + GemFireVersion.getBuildDate());
```
rhoughton-pivot, line 27: Lets remove all of the internal method calls!

PR #645, opened 2020-08-30
GEODE-8469: Enable no-missing-variable-declarations

moleske:

This is in draft because it is working for everything but the following three tests on Windows
```console
14 - testRegionAccessThreadSafe (Timeout)
24 - testThinClientCacheableStringArray (Timeout)
84 - testThinClientInterestList2 (Timeout)
```

I don't have a good Windows machine available to debug, so I need some help.  I think it is not ending the running Geode correctly because if I start these tests, let the run for long enough, then manually go into task manager and kill all the java processes, the tests end and pass.  I'm guessing the work I did in CacheHelper was fine for Mac and Ubuntu, but is insufficient for Windows.

Another interesting side discovery doing this is that this warning actually shows up almost entirely in the old cpp tests and not much in production code (5 files in production code)

I checked the box for `Allow edits and access to secrets by maintainers` so if you figure out the windows thing before me, please push the change!

### DIFF
--- a/cppcache/benchmark/GeodeLoggingBM.cpp
+++ b/cppcache/benchmark/GeodeLoggingBM.cpp
@@ -106,17 +106,17 @@ void GeodeLogToFile(benchmark::State& state) {
   }
 }
 
-static auto LogStringsToConsole = GeodeLogToConsole<GeodeLogStrings>;
-static auto LogIntsToConsole = GeodeLogToConsole<GeodeLogInts>;
-static auto LogComboToConsole = GeodeLogToConsole<GeodeLogCombo>;
+static const auto kLogStringsToConsole = GeodeLogToConsole<GeodeLogStrings>;
+static const auto kLogIntsToConsole = GeodeLogToConsole<GeodeLogInts>;
+static const auto kLogComboToConsole = GeodeLogToConsole<GeodeLogCombo>;
 
-static auto LogStringsToFile = GeodeLogToFile<GeodeLogStrings>;
-static auto LogIntsToFile = GeodeLogToFile<GeodeLogInts>;
-static auto LogComboToFile = GeodeLogToFile<GeodeLogCombo>;
+static const auto kLogStringsToFile = GeodeLogToFile<GeodeLogStrings>;
+static const auto kLogIntsToFile = GeodeLogToFile<GeodeLogInts>;
+static const auto kLogComboToFile = GeodeLogToFile<GeodeLogCombo>;
 
-BENCHMARK(LogStringsToConsole)->Range(8, 8 << 10);
-BENCHMARK(LogIntsToConsole)->Range(8, 8 << 10);
-BENCHMARK(LogComboToConsole)->Range(8, 8 << 10);
-BENCHMARK(LogStringsToFile)->Range(8, 8 << 10);
-BENCHMARK(LogIntsToFile)->Range(8, 8 << 10);
-BENCHMARK(LogComboToFile)->Range(8, 8 << 10);
+BENCHMARK(kLogStringsToConsole)->Range(8, 8 << 10);
+BENCHMARK(kLogIntsToConsole)->Range(8, 8 << 10);
+BENCHMARK(kLogComboToConsole)->Range(8, 8 << 10);
+BENCHMARK(kLogStringsToFile)->Range(8, 8 << 10);
+BENCHMARK(kLogIntsToFile)->Range(8, 8 << 10);
+BENCHMARK(kLogComboToFile)->Range(8, 8 << 10);
--- a/cppcache/benchmark/GeodeLoggingBM.cpp
+++ b/cppcache/benchmark/GeodeLoggingBM.cpp
@@ -35,9 +35,9 @@ using apache::geode::client::internal::geode_hash;
 
 const int STRING_ARRAY_LENGTH = 3;
 
-int g_iteration = 0;
+static int g_iteration = 0;
 
-std::array<const char*, STRING_ARRAY_LENGTH> logStrings{
+static std::array<const char*, STRING_ARRAY_LENGTH> logStrings{
     "Short test string", "Slightly longer test string",
     "Very long string: "
     "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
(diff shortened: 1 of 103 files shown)


cppcache/benchmark/GeodeLoggingBM.cpp
```diff
@@ -106,13 +106,13 @@ void GeodeLogToFile(benchmark::State& state) {
   }
 }
 
-auto LogStringsToConsole = GeodeLogToConsole<GeodeLogStrings>;
-auto LogIntsToConsole = GeodeLogToConsole<GeodeLogInts>;
-auto LogComboToConsole = GeodeLogToConsole<GeodeLogCombo>;
+static auto LogStringsToConsole = GeodeLogToConsole<GeodeLogStrings>;
```
pivotal-jbarrett, line 109: `const` also?

cppcache/integration-test/fw_dunit.cpp
```diff
@@ -71,10 +71,10 @@ using apache::geode::client::testframework::BBNamingContextServer;
 #define __DUNIT_NO_MAIN__
 #include "fw_dunit.hpp"
 
-ACE_TCHAR *g_programName = nullptr;
-uint32_t g_coordinatorPid = 0;
+static ACE_TCHAR *g_programName = nullptr;
```
pivotal-jbarrett, line 74: Yuck! I attempted to take on this beast and gave up on trying to clean up the old tests. Good job!

tests/cpp/security/XmlAuthzCredentialGenerator.hpp
```diff
@@ -55,7 +55,7 @@ const opCodeList::value_type QArr[] = {OP_QUERY, OP_REGISTER_CQ};
 
 const stringList::value_type QRArr[] = {"Portfolios", "Positions"};
 
-const char* PRiUsnm = "%s%d";
+static const char* PRiUsnm = "%s%d";
```
pivotal-jbarrett, line 58: `constexpr` work here?

cppcache/test/CacheXmlParserTest.cpp
```diff
@@ -99,7 +99,7 @@ std::string valid_cache_config_body = R"(<region name = 'Root1' >
     </region>
 </client-cache>)";
 
-std::string invalid_cache_config_body = R"(<region >
+static std::string invalid_cache_config_body = R"(<region >
```
pivotal-jbarrett, line 102: `const` ?

cppcache/src/PdxFieldType.cpp
```diff
@@ -31,7 +31,7 @@ namespace apache {
 namespace geode {
 namespace client {
 
-int32_t fixedTypeSizes[] = {
+static int32_t fixedTypeSizes[] = {
```
pivotal-jbarrett, line 34: `const` and use the `const` naming convention here?

cppcache/integration-test/fw_dunit.cpp
moleske: I think the worst part is since I was letting the compiler tell me what was wrong rather than proactively look for places, I build only a partial percent at a time when it was building the integration-test.  Thanks for the reminder on places that should be able to use `const`!  Putting a link to the [`const` naming style guide](https://google.github.io/styleguide/cppguide#Constant_Names) so I don't forget

tests/cpp/security/XmlAuthzCredentialGenerator.hpp
moleske: It did not.  `ISO C++11 does not allow conversion from string literal to 'char *const'` is the specific error when trying to use it while on a mac. Haven't looked more into at the moment

dgkimura (commented):
Apologies for any unwanted drive-by reviews.

echobravopapa (commented):
@moleske I normally wait for PRs to come out of Draft mode... lmk

moleske:
@echobravopapa as mentioned in the initial blurb up top, I do not have easy access to a useful windows machine to debug, but the PR was generally working.  In that blurb, I mention that I could use some community help on why Windows is being a pain (this time) compared to linux/mac.  I'll take it out of draft just to force your review then :)

cppcache/benchmark/GeodeLoggingBM.cpp
```diff
@@ -35,9 +35,9 @@ using apache::geode::client::internal::geode_hash;
 
 const int STRING_ARRAY_LENGTH = 3;
 
-int g_iteration = 0;
+static int g_iteration = 0;
```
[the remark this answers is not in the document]
moleske: I'll have to look closer at the benchmark code (we have no instructions in the CONTRIBUTING.md file about running benchmarks which seems bad...)

For the old style integration tests, I don't think I'd worry about the paradigm, since that code is not really getting updated, more just ported to the new style tests.  In the actual production code, I don't think we want to use an anonymous namespace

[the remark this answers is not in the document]
dgkimura: Curious, why wouldn't you want production code to use an anonymous namespace?

pdxcodemonkey:
So the following 3 tests are now consistently failing on Windows:
```
The following tests FAILED:
	 14 - testRegionAccessThreadSafe (Timeout)
	 24 - testThinClientCacheableStringArray (Timeout)
	 84 - testThinClientInterestList2 (Timeout)
```
I did some initial investigation thusly:
i. I established a baseline on the develop branch for the test testRegionAccessThreadSafe.  On my EC2 instance, it passes in ~75 seconds
4:00
ii. I then reverted your changes to just testRegionAccessThreadSafe.cpp , rebuilt and re-ran the test, which timed out
4:01
iii.  I reapplied your change to the test, and reverted your changes to just the test framework files /cppcache/integration-test/fw_*.cpp, and re-ran the test, which still timed out
4:02
That's about it for low-hanging fruit, I think.  I think we'll have success with a 'bisect' approach, where we revert subsets of the change until it starts passing again

pdxcodemonkey (changes requested):
See my notes re: failing tests on Windows.  Once we get those all passing, this will be good to go.

moleske:
@pdxcodemonkey do you think it will be easier to rewrite those tests in the new framework rather than find and fix the problem?  It may take more time but long term be more valuable

pdxcodemonkey:
I think we can diagnose the test problem with a couple of (very tedious) hours of debugging, which is a lot better than rewriting the test.  I think we're much closer to tackling the job of rewriting _all_ of the old tests, tho.

pivotal-jbarrett:
> I think we can diagnose the test problem with a couple of (very tedious) hours of debugging, which is a lot better than rewriting the test. I think we're much closer to tackling the job of rewriting _all_ of the old tests, tho.

I would just not apply this check to the legacy tests. They really aren't worth cleaning up.

cppcache/src/ExceptionTypes.cpp
```diff
@@ -30,8 +30,8 @@ namespace client {
 void setThreadLocalExceptionMessage(std::string exMsg);
 const std::string& getThreadLocalExceptionMessage();
 
-std::map<GfErrType,
-         std::function<void(std::string, std::string&, GfErrType, std::string)>>
+static std::map<GfErrType, std::function<void(std::string, std::string&,
+                                              GfErrType, std::string)>>
```
pivotal-jbarrett, line 34: Can we make this conform to the naming convention. 

Will be nice if some day this could be `constexpr`. 
https://www.youtube.com/watch?v=INn3xa4pMfg

pivotal-jbarrett (approved):
Just one minor naming convention issue I saw, otherwise ok.

pdxcodemonkey (approved):

cppcache/src/ExceptionTypes.cpp
moleske: Got it!  I'll have some tests run again just to make sure it is still green

PR #646, opened 2020-08-31
GEODE-8470: Properly stop running containers in SNI

moleske:
-use docker system prune -f to clean up networks and stopped containers

Noticed when trying to `docker rmi` the downloaded images and it said there were still containers with them.

### DIFF
--- a/clicache/integration-test2/SNITests.cs
+++ b/clicache/integration-test2/SNITests.cs
@@ -58,10 +58,10 @@ namespace Apache.Geode.Client.IntegrationTests
         public void Dispose()
         {
 
-            var dockerComposeProc = Process.Start(@"docker-compose.exe", "stop");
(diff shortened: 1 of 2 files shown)


clicache/integration-test2/SNITests.cs
```diff
@@ -58,10 +58,10 @@ public SNITests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
...
 
-            var dockerComposeProc = Process.Start(@"docker-compose.exe", "stop");
+            var dockerComposeProc = Process.Start(@"docker-compose.exe", "-f " + Config.SniConfigPath + "/docker-compose.yml" + " stop");
             dockerComposeProc.WaitForExit();
 
-            var dockerProc = Process.Start(@"docker.exe", "container prune -f");
+            var dockerProc = Process.Start(@"docker.exe", "system prune -f");
             dockerProc.WaitForExit();
 
         }
```
mmartell, line 67: Good change. Didn't know about system.

PR #647, opened 2020-09-01
GEODE-8460: Run SNITests Based on Docker Availability

mmartell:
This PR cleans up our docker based SNITests. In particular:

- SNITests are now enabled during cmake configuration based on whether docker and docker-compose are installed.
- SNITests have been moved to a new acceptance-test folder to keep them separated from tests which don't use docker.


moleske:
I'm excited about this!  It seems like there is a new category of `acceptance-test`.  Not sure when difference between integration and acceptance is (I think it is just does it need docker), but there should be an update to CONTRIBUTING.md to make sure that contributors run tests in the acceptance test directory

mmartell:
> I'm excited about this! It seems like there is a new category of `acceptance-test`. There should be an update to CONTRIBUTING.md to make sure that contributors run tests in the acceptance test directory

Good catch moleske! In fact I think geode-native is sorely lacking in documentation on how to use the new test frameworks to write tests. I'm told geode also is lacking in this area.

moleske:
>  I think geode-native is sorely lacking in documentation on how to use the new test frameworks

For this PR, I think just a one liner mentioning they exist is a good enough.  Agree that it could be better in both places

cppcache/integration/test/CMakeLists.txt
```diff
@@ -97,7 +96,7 @@ configure_file(
   ${CMAKE_CURRENT_SOURCE_DIR}/func_cacheserver2_pool.xml
   ${CMAKE_CURRENT_BINARY_DIR}/func_cacheserver2_pool.xml COPYONLY)
 
- file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/sni-test-config DESTINATION ${CMAKE_CURRENT_BINARY_DIR}) 
+#file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/sni-test-config DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
pdxcodemonkey, line 99: Please remove commented lines

cppcache/integration/test/sni-test-config/docker-compose.yml
```diff
@@ -1,43 +0,0 @@
-#
```
pdxcodemonkey, line 1: Let's get our act together with respect to all these support files for the SNI tests before we merge this.  There's now a geode.properties in _both_ the CLI and the C++ `acceptance-test` directories, are they the same?  If so, they should be in a common area.  You've removed the `sni-test-config` directory from the integration tests, and there's an `sni-test-config` dir at the top level of the project, are the tests pulling everything from this directory?  There is, at least, a file `truststore_sni.pem` in `/sni-test-config/geode-config` and there's also one, from earlier refactoring, in `/ssl_keys/client_keys` - there can be only one!  Also, we're installing SSL certs via INSTALL in CMakeLists.txt.  Please don't do this, we spent a lot of effort removing duplicate SSL certs/keys - tests should always pick these up from the source directory via defined constants in TestConfig.h or Config.cs.

echobravopapa: second that.

mmartell: Good catch on the unnecessary geode.properties in the clicache/acceptance-test folder. Has now been removed. Don't see one in the cppcache/acceptance-test folder.

The cppcache/integration/test/sni-test-config was removed, as all sni tests are pulling from sni-test-config that resides in the root folder.

Good catch on the duplicate truststore_sni.pem. Makes good sense to not duplicate files, particularly cert files. SNI tests have been modified to pull this client side truststore from the ssl_keys/client_keys folder.

cppcache/integration/test/CMakeLists.txt
```diff
@@ -117,7 +114,7 @@ file(GLOB SSL_CERTIFICATES
         "${CMAKE_CURRENT_SOURCE_DIR}/*Keys"
         )
 
-file(INSTALL ${SSL_CERTIFICATES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+#file(INSTALL ${SSL_CERTIFICATES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
pdxcodemonkey, line 117: Still missed this commented out line - please delete.

cppcache/acceptance-test/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+include(GoogleTest)
+gtest_discover_tests(cpp-acceptance-test
+  DISCOVERY_TIMEOUT 60
+)
+
+file(GLOB SSL_CERTIFICATES
+        "${CMAKE_CURRENT_SOURCE_DIR}/*Keys"
+        )
+
+file(INSTALL ${SSL_CERTIFICATES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
pdxcodemonkey, line 79: Why are you still copying certs to the binary dir?  This shouldn't be necessary.

pdxcodemonkey (changes requested):

cppcache/integration/test/CMakeLists.txt
mmartell: Weird. Not sure how that came back.

cppcache/acceptance-test/CMakeLists.txt
mmartell: Good catch. Indeed this isn't necessary.

pdxcodemonkey (approved):
If you think of it before merging, please remove the commented code on line 59 of cppcache/acceptance-test/CMakeLists.txt.  Looks good, nice work on this!

PR #648, opened 2020-09-02
GEODE-8480: Add txmanager check in tx example

alb3rtobr:


pdxcodemonkey:
Did you find an issue that was fixed by this?  What is the repro case?  Can we add a simple test that shows the problem?  The changes look fine, but it's difficult to judge correctness or whether there's a product issue since it doesn't modify any product code.

alb3rtobr:
I was running a test consisting on killing a Geode server while a C++ client with several threads was executing transactions. I observed that most of the times, the client died after killing the server with following error in logs:
```
[error 2020/09/01 10:31:05.977905 CEST alb3rtobr-XPS:7395 139697166743296] Unexpected exception during completing transaction : illegal State
terminate called after throwing an instance of 'apache::geode::client::Exception'
terminate called recursively
  what():  : illegal State
```

I was not sure about what was happening, but I had the feeling that I was doing something wrong, because I would not expect that the client was dying in this scenario. When I was taking a look at the transaction page in the Geode documentation, I realized that in Java, the transaction example includes the call to the `exists()` method ([link to the doc](https://geode.apache.org/docs/guide/112/developing/transactions/directed_example.html)) before performing rollback.

```
if(txManager.exists()) {
    txManager.rollback();
}     
```
I added the call to the function to my C++ client and the problem was solved. After that the client was showing some exceptions when one of the server was killed, but the client was not dying. So I felt that calling to this `exists()`method was something to be done in transactions and that it was missing from the C++ documentation.

pdxcodemonkey (changes requested):
Please add a test that reproduces the problem, if you can.  If it's too much work, file a separate Jira work item for the test and I'll approve/merge this.  Thanks!

alb3rtobr:
Dont worry, I think its not needed to split this in two different tickets, I will try to write a test when I find time. Im moving this PR to draft in the meanwhile.

alb3rtobr:
There you are a test that shows the problem. I have divided the code in two commits. In the first one, the test fails with the following error:

```
$ ctest -R TransactionsTest -j1
Test project /home/alb3rtobr/CLionProjects/Nordix/geode-native/build/cppcache/integration/test
    Start 61: TransactionsTest.ExceptionWhenRollingBackTx
1/1 Test #61: TransactionsTest.ExceptionWhenRollingBackTx ...***Exception: Child aborted 21.41 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =  23.72 sec

The following tests FAILED:
	 61 - TransactionsTest.ExceptionWhenRollingBackTx (Child aborted)
Errors while running CTest
```

But after changing the code to call `transactionManager->exists()` before executing the rollback (the second commit), the test case works fine.

alb3rtobr:
PR rebased to fix conflict with `CMakeLists.txt`

pdxcodemonkey (approved):

PR #649, opened 2020-09-09
GEODE-8449: Cleanup containers before each test

mmartell:
This cleans up any docker containers at the beginning of each test in case of a test failure that leaves docker containers up.

### DIFF
--- a/clicache/acceptance-test/SNITests.cs
+++ b/clicache/acceptance-test/SNITests.cs
@@ -34,7 +34,7 @@ namespace Apache.Geode.Client.IntegrationTests
 
         public SNITests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
         {
-			Dispose();
+			CleanupDocker();
 
             currentWorkingDirectory = Directory.GetCurrentDirectory();
             var clientTruststore = Config.SslClientKeyPath + @"/truststore_sni.pem";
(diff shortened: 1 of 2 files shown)


clicache/acceptance-test/SNITests.cs
```diff
@@ -34,6 +34,8 @@ public class SNITests : TestBase, IDisposable
 
         public SNITests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
         {
+			Dispose();
```
echobravopapa, line 37: does this make sense for C# language?  using the `Dispose` method...?

cppcache/acceptance-test/SNITest.cpp
```diff
@@ -52,6 +52,8 @@ class SNITest : public ::testing::Test {
   ~SNITest() override = default;
 
   void SetUp() override {
+    TearDown();
```
echobravopapa, line 55: same thing here... `gtest` uses `TearDown` as a special method... I wonder if it would be better to have a utility method for both that is called from the constructor in C# and `SetUp` in this test...

echobravopapa (changes requested):
Happy to approve on the functionality, just need to see a strong case for the approach noted in the review...

clicache/acceptance-test/SNITests.cs
pdxcodemonkey: Yeah, please create a utility method to call from Dispose, and call that here.  Calling Dispose directly is legal, but technically misuse of the semantics of IDisposable.

cppcache/acceptance-test/SNITest.cpp
pdxcodemonkey: Yep

clicache/acceptance-test/SNITests.cs
mmartell: Agreed. Though legal, it looks a bit weird to call Dispose directly.

cppcache/acceptance-test/SNITest.cpp
mmartell: Agreed. Employing a utility function as above.

echobravopapa (approved):

cppcache/acceptance-test/SNITest.cpp
moleske: This looks like it was merged with `TearDown()` being called instead of `cleanupDocker()` function...

PR #650, opened 2020-09-10
GEODE-8129: Added a drop proxy test.

mreddington:


mmartell (changes requested):
No longer need the #if defined(_WIN32). SNI tests are enabled/disabled based on docker availability. The acceptance-tests will only run the linux platform in the CI. Please remove.

cppcache/acceptance-test/SNITest.cpp
```diff
@@ -174,4 +174,51 @@ TEST_F(SNITest, connectWithoutProxyFails) {
...
+  auto region = cache.createRegionFactory(RegionShortcut::PROXY)
+                    .setPoolName("pool")
+                    .create("jellyfish");
+
+  auto systemRVal = std::system("docker-compose pause");
+  if (systemRVal == -1) {
+    BOOST_LOG_TRIVIAL(error) << "std::system returned: " << systemRVal;
+  }
+
+  region->put("1", "one");
```
echobravopapa, line 210: you'll want to check/prove that the Put doesn't go through... i.e. its an expected failure and I bet there is an exception thrown...

echobravopapa (changes requested):

cppcache/acceptance-test/SNITest.cpp
mreddington: That's literally impossible. A put or get will never block and never throw due to connectivity issues.

echobravopapa: There won't be any exception or error after the connection times out?

PR #651, opened 2020-09-17
Remove bucketServerLocation if timeout error

alb3rtobr:


alb3rtobr:
@echobravopapa @pdxcodemonkey This is the PR with the change regarding the client metadata update when there is a timeout error that I mentioned in the dev list

alb3rtobr:
@pdxcodemonkey @echobravopapa Have you had the opportunity to check how could I add a test to verify this?

What if I create a separate ticket for the test case and this change is merge in the meanwhile? As after a timeout there is always a IO error, this change seems an improvement in the current situation when that happens, as it is in our case.

alb3rtobr:
I will use a separated ticket and branch for this issue, so im closing this PR. New one is here https://github.com/apache/geode-native/pull/662

PR #656, opened 2020-09-25
GEODE-8530: Fix for coredump during tx commit

jvarenina:
Client is fixed in a way that ignores all regions enlisted within
transaction that client doesn't have configured. This behavior is
aligned with the java client, since same thing is done there.


pdxcodemonkey (approved):
Looks good to me, and the new test passes on all platforms. thanks for the fix!

jvarenina:
Thanks for the review!

PR #658, opened 2020-09-28
GEODE-8129: A test that breaks and reestablishes the proxy connection.

mreddington:


cppcache/acceptance-test/SNITest.cpp
```diff
@@ -174,4 +175,53 @@ TEST_F(SNITest, connectWithoutProxyFails) {
...
+      .addLocator("locator-maeve", 10334)
+      .create("pool");
+
+  auto region = cache.createRegionFactory(RegionShortcut::PROXY)
+                    .setPoolName("pool")
+                    .create("jellyfish");
+
+  auto systemRVal = std::system("docker pause haproxy");
+  if (systemRVal == -1) {
+    BOOST_LOG_TRIVIAL(error) << "std::system returned: " << systemRVal;
```
pdxcodemonkey, line 204: Where does this log statement go?  We're using gtest as our framework, does it have a logging facility that would be more appropriate?  Just curious, before we drop in another pseudo-dependency like this.  Is it an error here if the system call fails?  If so, shouldn't we just `EXPECT_NE(systemRVal, -1)` rather than log and move on?

pdxcodemonkey (changes requested):

mmartell (approved):
Looks good.

cppcache/acceptance-test/SNITest.cpp
mreddington: Honestly I don't know. This test is cobbled out of the existing SNI tests, this is the same structure we use to initiate the docker images - invoke the command, check the result, log the error. I figured when in Rome...

The pseudo-dependency is already there because this is how we start the proxy in the test fixture.

I don't think an expectation is appropriate here. We're not testing whether or not a system command works... You know? I'd love if GTest had an "Incomplete" state for tests that incur an error beyond the scope of the test, in the environment. But then again, we shouldn't be writing big tests like this in the first place with a unit test framework.

cppcache/acceptance-test/SNITest.cpp
```diff
@@ -174,4 +174,51 @@ TEST_F(SNITest, connectWithoutProxyFails) {
...
+                    .create("jellyfish");
+
+  auto systemRVal = std::system("docker-compose pause");
+  if (systemRVal == -1) {
+    BOOST_LOG_TRIVIAL(error) << "std::system returned: " << systemRVal;
+  }
+
+  region->put("1", "one");
+
+  std::this_thread::sleep_for(std::chrono::seconds(17));
```
echobravopapa: the magic of prime numbers?

cppcache/acceptance-test/SNITest.cpp
pdxcodemonkey: Meh - fine, it's just test code.  If we ever need to debug it, we can figure out where BOOST_LOG_TRIVIAL output goes.

pdxcodemonkey (approved):

PR #659, opened 2020-09-29
GEODE-8548 Avoid hostname resolution in TcpConn

gaussianrecurrence:
- Calling get_host_name() in TcpConn::connect involves a system call that
   might take several seconds to complete.
 - So endpoint is stored instead upon TcpConn construction and used later
   instead of making the system call.

### DIFF
--- a/cppcache/src/TcpConn.cpp
+++ b/cppcache/src/TcpConn.cpp
@@ -129,6 +129,7 @@ TcpConn::TcpConn(const std::string& address,
     : stream_(nullptr),
       maxBuffSizePool_(maxBuffSizePool),
       inetAddress_(address.c_str()),
+      endpoint_(address),
       timeout_(waitSeconds) {}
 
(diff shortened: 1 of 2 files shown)


pdxcodemonkey (approved):
Yikes, that's pretty subtle - I had to read it a couple of times to see where the important change was.  Looks good, and glad to see us get rid of this call.  Needless to say, it's super bad to have meaningful code executing inside a LOG* macro.

I'm verifying the build and tests on all platforms, then will merge if everything checks out.

gaussianrecurrence:
> Yikes, that's pretty subtle - I had to read it a couple of times to see where the important change was. Looks good, and glad to see us get rid of this call. Needless to say, it's super bad to have meaningful code executing inside a LOG* macro.
> 
> I'm verifying the build and tests on all platforms, then will merge if everything checks out.

Yes, it was kind of a tricky one to troubleshoot haha. Thanks btw for reviewing the PR :)

PR #661, opened 2020-09-30
GEODE-8512: Add expectDomainClass for PdxInstance

gaussianrecurrence:
- In Java client you can specify whether or not the PdxInstance you
   are trying to create when calling createPdxInstanceFactory
   represents a Java domain class.
 - This commit adds the possibility to also specify expectDomainClass
   for the native C++ client.
 - Take into account that this is necessary in order to use cquery's
   for PdxInstance(s) which classname is __GEMFIRE_JSON.
 - Fix doxygen comments for old createPdxInstanceFactory overloads.


mreddington (approved):

mmartell:
Mario, this looks like a nice piece of new functionality. Your new test has uncovered a bug in our new C++ test framework (https://issues.apache.org/jira/browse/GEODE-8580). Fixing that first. Once that's merged I think we can merge your PR.

gaussianrecurrence:
Thanks for the update :)

cppcache/src/PdxType.cpp
```diff
@@ -53,9 +54,13 @@ PdxType::PdxType(PdxTypeRegistry& pdxTypeRegistry,
       m_isVarLenFieldAdded(false),
       m_remoteToLocalFieldMap(nullptr),
       m_localToRemoteFieldMap(nullptr),
-      m_noJavaClass(false),
+      m_noJavaClass(!expectDomainClass),
```
pivotal-jbarrett, line 57: Let's invert the member variable to be positive and reflect the usage. So if it is the "expect domain class" member now let's rename it so.

cppcache/src/Cache.cpp
```diff
@@ -153,7 +153,12 @@ LogLevel Cache::getLogLevel() { return Log::logLevel(); }
 
 PdxInstanceFactory Cache::createPdxInstanceFactory(
     const std::string& className) const {
-  return m_cacheImpl->createPdxInstanceFactory(className);
+  return createPdxInstanceFactory(className, true);
```
pivotal-jbarrett, line 156: Why does this default to `true`, wouldn't it have been implicitly `false` perviously?

pivotal-jbarrett (changes requested):

cppcache/integration/test/PdxJsonTypeTest.cpp
```diff
@@ -68,6 +68,32 @@ std::shared_ptr<Region> setupRegion(Cache& cache) {
...
+  auto cache = cluster.createCache();
+  auto region = setupRegion(cache);
+  const std::string query_stmt{"SELECT * FROM /region"};
+
+  region->put("non-java-domain-class-entry",
+              cache.createPdxInstanceFactory(gemfireJsonClassName, false)
+                  .writeString("foo", "bar")
+                  .create());
+
+  EXPECT_NO_THROW(gfsh.query(query_stmt).execute());
```
pivotal-jbarrett, line 87: Can we really only test this via `gfsh`? What about writing a Java function to deploy and execute. That would have fewer moving parts and no output text to parse.

pivotal-jbarrett (changes requested):
I think there is also a need for a unit test to verify the message is properly serialized. We have examples for some messages, not sure if there is one already for the original case where it was always false.

pdxcodemonkey:
@gaussianrecurrence Please rebase your branch from the latest `develop` and push.  @mmartell 's fix for GEODE-8580 should fix your test, and this will be good to go after you address Jake's feedback.  Thanks!

cppcache/src/PdxType.cpp
gaussianrecurrence: Good point :)

cppcache/src/Cache.cpp
gaussianrecurrence: By default a Java class is expected, that has always been the case. I get it can be a little bit confusing as flag in the serialized PdxType is inverted.

cppcache/src/CacheImpl.cpp
```diff
@@ -814,14 +814,19 @@ DataInput CacheImpl::createDataInput(const uint8_t* buffer, size_t len,
...
+    const std::string& className, bool expectDomainClass) const {
   this->throwIfClosed();
 
   return PdxInstanceFactory(
-      className, *m_cacheStats, *m_pdxTypeRegistry, *this,
+      className, expectDomainClass, *m_cacheStats, *m_pdxTypeRegistry, *this,
       m_distributedSystem.getSystemProperties().getEnableTimeStatistics());
 }
 
+PdxInstanceFactory CacheImpl::createPdxInstanceFactory(
```
gaussianrecurrence, line 825: I was thinking on removing this overload as this is not part of the public API what do you think?

gaussianrecurrence:
> I think there is also a need for a unit test to verify the message is properly serialized. We have examples for some messages, not sure if there is one already for the original case where it was always false.

I did not think about this. Thanks for the comment, it's a fair point :)

cppcache/integration/test/PdxJsonTypeTest.cpp
gaussianrecurrence: Well, if possible, that would be much better, for sure. Will take a look at it :)

cppcache/integration/test/PdxJsonTypeTest.cpp
```diff
@@ -68,6 +68,32 @@ std::shared_ptr<Region> setupRegion(Cache& cache) {
...
+
+  region->put("java-domain-class-entry",
+              cache.createPdxInstanceFactory(gemfireJsonClassName)
+                  .writeString("foo", "bar")
+                  .create());
+
+  EXPECT_THROW(gfsh.query(query_stmt).execute(), GfshExecuteException);
```
mmartell, line 94: Please check the intended use of this EXPECT_THROW. Even with the GEODE-8580 fix line 94 does not throw. Maybe because the value already exists from the first put? If you put a different object (e.g. "foo2", "bar") on line 91, then line 94 throws, and the test passes.

gaussianrecurrence: Hi @mmartell,

I think there's a singularity in the code. And with this I mean that it seems with those entries this test shows that code fails in failing. TBH I don't have a clue about what's happenning, I've tried changing the entries I write and these are the results:

1.  **FAILED**: [1st: non-java-domain-class-entry => {foo: bar}; 2nd: java-domain-class-entry => {foo: bar}]
2. **PASSED**: [1st: non-java-domain-class-entry => {foo: bar}; 2nd: java-domain-class-entry => {foo2: bar}]
3. **PASSED**: [1st: non-java-domain-class-entry => {foo: bar}; 2nd: java-domain-class-entry => {foo: bar2}]
4. **FAILED**: [1st: non-java-domain-class-entry => {foo: bar}; 2nd: non-java-domain-class-entry => {foo: bar}]
5. **FAILED**: [1st: non-java-domain-class-entry => {foo: bar}; 2nd: non-java-domain-class-entry => {foo2: bar}]
6.  **PASSED**: [1st: entry-1 => {entryName: entry-1}; 2nd: entry-2 => {entryName: entry-2}]
7.  **PASSED**: [1st: entry-1 => {entryName: entry-1}; 2nd: entry-2 => {entryName: entry-1}]

Cases **4 and 5** would somehow make sense to me, but should not be happenning as the PdxType should be different.
But the f... up cases are **1 and 7** as they are equivalent but still 1 fails and 7 not.

If you have any idea about what could be happening, I'd appreciate your ideas.
My feeling here is that there are some issues with hashing functions either on the client or on the server, or even maybe in both. That's the only explanation for case 1 failing and 7 not as of my understanding.

Anyhow, will keep digging on this matter :S

gaussianrecurrence: Ok, so I've been going throught the query engine code of the server and finally located the problem.
As I feared org.apache.geode.cache.query.internal.Bag addAndGetOccurence method uses mapGet on the different objects added, and PdxInstanceImpl hash is not considering noJavaDomainClass attribute and neither is the equals operator therefore all PdxInstanceImpl with the same hash that are equals as of equals operator are not added but instead the previous existing copy inside bag is duplicated.
Now the question, and here is where I need your help, is: "Is this a bug or a feature?" 
Because as of my point of view is a bug, but there might be many more implications than the one I am able to see.

gaussianrecurrence:
- Requested changes have been made.
- Branch has been rebased.
- clang-format passed.
- Compilation runs OK both in Linux and Windows.
- All tests passed both in Linux and Windows (thanks to @moleske for contributing GitHub actions)

CMakeLists.txt
```diff
@@ -254,6 +254,7 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     -Wno-undefined-func-template #TODO fix
     -Wno-float-equal #TODO fix
     -Wno-conversion #TODO fix
+    -Wno-missing-variable-declarations #TODO fix
```
moleske, line 257: Is there a reason this was added back in?  It was recently fixed in https://github.com/apache/geode-native/pull/645 with it only being applied to the [old integration tests](https://github.com/apache/geode-native/blob/develop/cppcache/integration-test/CMakeLists.txt#L110) to save hassle on the old stuff

gaussianrecurrence: Thanks for catching that. I, for some reason, reverted the commit you made by mistake and got squash with revision 2.
I've fixed it thought in revision 3. I am currently executing the pipeline for the latests changes.

cppcache/src/CacheImpl.cpp
pivotal-jbarrett: If it isn't publicly exported yes, remove extra cruft.

cppcache/include/geode/Cache.hpp
```diff
@@ -222,14 +222,25 @@ class APACHE_GEODE_EXPORT Cache : public GeodeCache {
...
+   * Returns a factory that can create a {@link PdxInstance}.
+   * @param className the fully qualified class name that the PdxInstance will
+   * become when it is fully deserialized.
+   * @param expectDomainClass Whether or not created PdxType represents a
+   * Java domain class.
+   * @throws IllegalStateException if the className is nullptr or invalid.
+   * @return the factory
+   */
+  PdxInstanceFactory createPdxInstanceFactory(
+      const std::string& className, bool expectDomainClass) const override;
```
pivotal-jbarrett, line 242: Is the domain class thing actually specific to Java? I could also apply to .NET or other languages where a reflection like behavior exists to instantiate a type by name dynamically. If it is specific to Java then maybe the param name should reflect that. If it isn't specific to Java then maybe the documentation should change.

pivotal-jbarrett (approved):

cppcache/include/geode/Cache.hpp
gaussianrecurrence: Short answer -> Currently, it's specific to Java, so yes, will update the variable name. Thanks for pointing it out
And it's not directly involved with reflection. You can have custom types deserialization and not have reflection. You just need to tell your application how to serialize/de-serialize objects.

PR #663, opened 2020-10-01
GEODE-8569: Add missing method to ClientMessageDecoder

pdxcodemonkey:
- This would be called, and thus throw an exception, for client logs using AuthInitialize
- Also added a new entry to version lookup for regex function dispatcher.

@mreddington @davebarnes97 @dihardman @karensmolermiller @alb3rtobr


mmartell (approved):
Nice enhancement, particularly as cloud computing (which requires auth) gains steam.

PR #664, opened 2020-10-01
GEODE-8519: User Guide - Delete 'preserving-data' topics

davebarnes97:
Housekeeping: Remove unpublished source files that were, nonetheless, visible via Google search.


pdxcodemonkey (approved):
Looks fine from here, just deleting no-longer-used stuff.

PR #667, opened 2020-10-05
GEODE-8531: Fix coredump in MapSegment::remove

gaussianrecurrence:
- Fixed a coredump in MapSegment::remove whenever the the region holding the
   MapSegment object had consistency checks disabled.
 - Added an integration test in RegisterKeysTest TS as it was were this issue
   was detected.


gaussianrecurrence:
**Revision 1** uploaded:

- [x] clang-format executed.

**Linux tests** list:
- [x]  Unit tests passed.
- [ ]  Old integration tests passed.
- [ ]  New integration tests passed.
- [ ]  Acceptance tests passed.

</br>

**Windows tests** list:
- [ ]  Unit tests passed.
- [ ]  Old integration tests passed.
- [ ]  New integration tests passed.
- [ ]  Acceptance tests passed.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -46,9 +46,11 @@ Cache createTestCache() {
       .create();
 }
 
-std::shared_ptr<Region> setupCachingProxyRegion(Cache& cache) {
+std::shared_ptr<Region> setupCachingProxyRegion(Cache& cache,
```
mreddington, line 49: Prefer an overload over a default parameter. This obfuscates the code as documentation and the meaning of all the prior tests. You're explicitly stating concurrency checks are not to be enabled. Why is that an important condition for the other tests? I presume what you wanted was the ability to enable checks for your one test.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -112,6 +114,51 @@ TEST(RegisterKeysTest, RegisterAllWithCachingRegion) {
...
+      .create()
+      .region()
+      .withName("region")
+      .withType("PARTITION")
+      .execute();
+
+  auto producer_cache = createTestCache();
+  auto listener_cache = createTestCache();
+
+  auto producer_region = [&cluster](Cache& cache) {
```
mreddington, line 132: Write it as a function. This looks like celebrating lambdas. A function that is only ever called in one place is inlined by all the major compilers. The lambda here obscures the code; you could have used a function with a useful name that tells me what you're doing, now I have an unnamed function body and I have to deduce it's meaning.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -112,6 +114,51 @@ TEST(RegisterKeysTest, RegisterAllWithCachingRegion) {
...
+    auto poolFactory = cache.getPoolManager().createFactory();
+    cluster.applyLocators(poolFactory);
+    poolFactory.create("default");
+    return setupProxyRegion(cache);
+  }(producer_cache);
+
+  auto listener_region = [&cluster](Cache& cache) {
```
mreddington, line 139: Write it as a function.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -112,6 +114,51 @@ TEST(RegisterKeysTest, RegisterAllWithCachingRegion) {
...
+    auto poolFactory =
+        cache.getPoolManager().createFactory().setSubscriptionEnabled(true);
+    cluster.applyLocators(poolFactory);
+    poolFactory.create("default");
+    auto region = setupCachingProxyRegion(cache, false);
+    region->registerAllKeys();
+    return region;
+  }(listener_cache);
+
+  ASSERT_FALSE(listener_region->containsValueForKey("one"));
```
mreddington, line 149: No need to establish a new, empty region, that cannot possibly have any preloaded data in it is indeed empty.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -112,6 +114,51 @@ TEST(RegisterKeysTest, RegisterAllWithCachingRegion) {
...
+
+  producer_region->put("one", std::make_shared<CacheableInt16>(1));
+  std::this_thread::sleep_for(std::chrono::seconds(5));
```
mreddington, line 152: Correct me if I'm wrong, but this delay should be unnecessary.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -112,6 +114,51 @@ TEST(RegisterKeysTest, RegisterAllWithCachingRegion) {
...
+
+  ASSERT_TRUE(listener_region->containsValueForKey("one"));
+
+  producer_region->destroy("one");
+  std::this_thread::sleep_for(std::chrono::seconds(5));
```
mreddington, line 157: Again, this should be unnecessary.

cppcache/integration/test/RegisterKeysTest.cpp
gaussianrecurrence: You have a good point indeed. Noted :)

cppcache/integration/test/RegisterKeysTest.cpp
gaussianrecurrence: Got it :)

cppcache/integration/test/RegisterKeysTest.cpp
gaussianrecurrence: Yes, indeed, this is overtesting

cppcache/integration/test/RegisterKeysTest.cpp
gaussianrecurrence: See my answer below

cppcache/integration/test/RegisterKeysTest.cpp
gaussianrecurrence: Thing to be tested here is that whenever MapSegment::remove is called with consistency is disabled no coredump occurs. I reproduced the issue with client notifications, that's why I created the test here. The delay is in order to make sure that the listener cache has enough time to receive the destroy event.
---
However I have to say I've suffered having to run testing pipelines for hours for each change, so I was thinking on changing the test to remove the delay and instead make it a timeout.
---
Other choice could be to just try remove without notifications, but I am not sure about this. What do you think?

mreddington: The basis for my doubt about the timeout is because I believe `containsValueForKey` is a server query (the method does some parameter checking, and then it calls `containsValueForKey_remote`), so there's no point in waiting, there is no state change in the client you need to wait for. The destroy is a blocking call and the server should have completed the request and all server-side effects before returning a response and thus unblocking the client call. So you have a producer region and a proxy region to that, both on the same server instance, I don't think consistency is going to be an issue.

gaussianrecurrence: Oh I see the problem. You see, using the client without region consistency is seriously broken because I've seen several coredumps for the following scenarios:

1. Putting a key into a region using the NC.
2. Destroying a key into a region using the NC.
3. Receiving a LOCAL_DESTROY notification in the NC.

The fix for the coredump in this PR is the one for the third case, so that's why I wanted to do it by using notifications. Thing is I was not aware that containsValueForKey queries the server. Therefore what about setting up a cachelistener for the 'listener_region' which actually is a Google Mock and make a call expectation on the afterDestroy function of the catch listener using an approach similar to this: https://stackoverflow.com/questions/10767131/expecting-googlemock-calls-from-another-thread?

The test flow would be:

1. Setup producer_region without cache and subscription to notifications.
2. Setup listener_region with cache, subscription notifications and a cache listener which actually is a google mock.
3. Call registerAllKeys on listener_region.
4. Setup an EXPECT_CALL for the cache listener afterDestroy method, specifying the sync barrier action to occur upon call.
5. Make a put in producer_region
6. Make a destroy on key same key that on step 5 for producer region.
7. Wait for (1, 2 or 5 seconds?) for the semaphore.

The test passes only if wait in the last step does not time out. What do you think?

gaussianrecurrence:
It seems that clang compilation does not like so much my implementation of the Gmock. Will fix it and let you know once ready for review :)

gaussianrecurrence:
Clang unused warning has been solved and I've rebased develop changes into the branch
All the tests are passing.

cppcache/integration/test/RegisterKeysTest.cpp
mreddington: I think we have a 15 year old code base with some dragons that can be slain another day. Really what I was hoping for was removing the wait, and oh look, it works just the same, no problem. If the test breaks without it, and fixing it requires a non-trivial refactor of some unrelated part of the NC, then I'm willing to live with it. Let's move on.

gaussianrecurrence: Well, thing is I've been thinking that, problem here is the problem is reproduced only when a notification is received, that's why we need the wait, because the notification thread is running in another thread.

BUT, after giving it a second thought it came to my mind that maybe it's possible to trick the NC to believe that a notification was received. Let me explore this possibility and will let you know if that's possible.

gaussianrecurrence: Well sadly, is not possible as TcrMessage would need to be exported into the shared library :S

mreddington (approved):

PR #669, opened 2020-10-06
GEODE-8578: Parse partition resolver in RESPONSE_CLIENT_PARTITION_ATTRIBUTES message

alb3rtobr:
I tried the gnmsg tool with a client log file connecting to a cluster that contains a partition region with a partition resolver, which caused an exception.

With this change, it is possible to parse the partition resolver name. Still pending to parse the list of fixed partition attributes, maybe in other ticket...

This is how the message looks like after parsing my log file:
```
{
  "message": {
    "Timestamp": "16:47:55.225776",
    "Connection": "0",
    "Direction": "<---",
    "Type": "RESPONSE_CLIENT_PARTITION_ATTRIBUTES",
    "Length": 80,
    "Parts": 3,
    "TransactionId": -1,
    "SecurityFlag": 0,
    "BucketCount": {
      "Size": 5,
      "IsObject": 1,
      "Data": {
        "DSCode": "CacheableInt32",
        "Value": 113
      }
    },
    "ColocatedWith": {
      "Size": 0,
      "IsObject": 0
    },
    "PartitionResolverName": {
      "Size": 60,
      "IsObject": 1,
      "Data": {
        "DSCode": "CacheableASCIIString",
        "StringLength": 57,
        "Value": "org.apache.geode.cache.util.StringPrefixPartitionResolver"
      }
    }
  }
}
```

### DIFF
--- a/tools/gnmsg/server_messages.py
+++ b/tools/gnmsg/server_messages.py
@@ -36,16 +36,20 @@ def read_bucket_count(message_bytes, offset):
 
 
 def read_partition_attributes(properties, message_bytes, offset):
-    if properties["Parts"] != 2 and properties["Parts"] != 4:
-        raise Exception(
-            "Don't know how to parse a RESPONSE_CLIENT_PARTITION_ATTRIBUTES message with "
-            + properties["Parts"]
-            + " parts (should have 2 or 4 only)."
-        )
-
     (properties["BucketCount"], offset) = read_bucket_count(message_bytes, offset)
     (properties["ColocatedWith"], offset) = parse_key_or_value(message_bytes, offset)
-    # TODO: parse parts 3 and 4 (partition resolver and list of partition attributes), if they exist
+    if properties["Parts"] == 4:
+        (properties["PartitionResolverName"], offset) = parse_key_or_value(message_bytes, offset)
+        # TODO: parse part 4 (list of partition attributes)
+    elif properties["Parts"] == 3:
+        try:
+            (properties["PartitionResolverName"], offset) = parse_key_or_value(message_bytes, offset)
+        except:
+            raise Exception(
+                "Don't know how to parse a RESPONSE_CLIENT_PARTITION_ATTRIBUTES message with "
+                + "3 parts and fpa attribute."
+            )
+        # TODO: parse part 3 if it is not partition resolver but list of partition attributes
 
 
 server_message_parsers = {


tools/gnmsg/server_messages.py
```diff
@@ -36,16 +36,20 @@ def read_bucket_count(message_bytes, offset):
...
+        # TODO: parse part 4 (list of partition attributes)
+    elif properties["Parts"] == 3:
+        try:
+            (properties["PartitionResolverName"], offset) = parse_key_or_value(message_bytes, offset)
+        except:
+            raise Exception(
+                "Don't know how to parse a RESPONSE_CLIENT_PARTITION_ATTRIBUTES message with "
+                + "3 parts and fpa attribute."
+            )
+        # TODO: parse part 3 if it is not partition resolver but list of partition attributes
```
pdxcodemonkey, line 52: Is it possible for the server to send this message with 3 parts?  And with part 3 being attributes, rather than the partition resolver name?  Looking at the C++ code, it's clear geode-native could never parse such a message, so I wonder if we need to here...

pdxcodemonkey (changes requested):

tools/gnmsg/server_messages.py
alb3rtobr: I used this code of `GetClientPartitionAttributesOp` Java class as reference:

```
...
int bucketCount;
String colocatedWith;
String partitionResolverName = null;
Set<FixedPartitionAttributes> fpaSet = null;
bucketCount = (Integer) msg.getPart(0).getObject();
colocatedWith = (String) msg.getPart(1).getObject();
if (msg.getNumberOfParts() == 4) {
  partitionResolverName = (String) msg.getPart(2).getObject();
  fpaSet = (Set<FixedPartitionAttributes>) msg.getPart(3).getObject();
} else if (msg.getNumberOfParts() == 3) {
  Object obj = msg.getPart(2).getObject();
  if (obj instanceof String) {
    partitionResolverName = (String) obj;
  } else {
    fpaSet = (Set<FixedPartitionAttributes>) obj;
  }
...
```

alb3rtobr: > Is it possible for the server to send this message with 3 parts?

It seems so. In my example, I got this exception:

`Exception: Don't know how to parse a RESPONSE_CLIENT_PARTITION_ATTRIBUTES message with 3 parts (should have 2 or 4 only).`

Well, actually what I got first was a `TypeError: must be str, not int` exception because `properties["Parts"]` is not a string:

```
raise Exception(
            "Don't know how to parse a RESPONSE_CLIENT_PARTITION_ATTRIBUTES message with "
            + properties["Parts"]
            + " parts (should have 2 or 4 only)."
        )
```

pdxcodemonkey: Okay, that's what I was wondering - does the Java code do it differently?  And that's clearly a yes.  For the record, the C++ code ignores the possibility of 3 parts entirely, so if this message ever comes back with 3 parts geode-native apparently will get no partition information.  If we were to dig into it, we'd probably find that some version of the server in the dim past used to send the 3-part variant, and the Java parsing code is left for compatibility with that ancient version of the protocol.

pdxcodemonkey: Oh interesting, so you've seen this in the wild.  Let's file another Jira ticket to discuss - it seems to me this might be important for geode-native to handle, and PR comments aren't a great place to iron things out.

Python changes look great, I'll go ahead and approve/merge this.  Thanks again!

pdxcodemonkey (approved):

tools/gnmsg/server_messages.py
alb3rtobr: Thanks, you were too fast :) After your comment I was going to modify the function to leave it like:

```
def read_partition_attributes(properties, message_bytes, offset):
    (properties["BucketCount"], offset) = read_bucket_count(message_bytes, offset)
    (properties["ColocatedWith"], offset) = parse_key_or_value(message_bytes, offset)
    if properties["Parts"] >= 3:
        (properties["PartitionResolverName"], offset) = parse_key_or_value(message_bytes, offset)
    if properties["Parts"] == 4:
        # TODO: parse part 4 (list of partition attributes)
        raise Exception(
                "Don't know how to parse a RESPONSE_CLIENT_PARTITION_ATTRIBUTES message with "
                + "4 parts containing list of partition attributes."
        )
```

pdxcodemonkey: lol no worries - there are a ton of things to add to gnmsg, you can add this refactor next time around :)

PR #673, opened 2020-10-12
GEODE-8604: Fix documentation errors on Mac build

pdxcodemonkey:
MacOS apparently flags more of these than our Linux builds, so this fixes up Mac builds.

@mreddington @davebarnes97 @dihardman


moleske (approved):
Weird that mac clang 11 didn't call these out before 😢 

Nor did clang6 🤔

PR #674, opened 2020-10-14
GEODE-8157: Add DropProxy acceptance test for .NET

mmartell:
This adds a new DropProxy test to the docker based acceptance tests for .NET.

Note: This test will only run on Windows platforms that have docker and docker-compose installed.

### DIFF
--- a/clicache/acceptance-test/SNITests.cs
+++ b/clicache/acceptance-test/SNITests.cs
@@ -35,7 +35,7 @@ namespace Apache.Geode.Client.IntegrationTests
 
         public SNITests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
         {
-			CleanupDocker();
+            CleanupDocker();
 
             currentWorkingDirectory = Directory.GetCurrentDirectory();
             var clientTruststore = Config.SslClientKeyPath + @"/truststore_sni.pem";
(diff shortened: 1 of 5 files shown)


clicache/acceptance-test/SNITests.cs
```diff
@@ -90,7 +91,18 @@ private int ParseProxyPort(string proxyString)
             return Int32.Parse(portNumberString);
         }
 
-        [Fact]
+
+		//private Task PutAsync(IRegion<string, string> region)
```
moleske, line 95: Alas, poor commented code! I knew it, @mmartell : a line of async puts, of most excellent fancy: it hath borne me on its back a thousand times; and now, how abhorred in my imagination it is!

mmartell: Lifted me for at least the remainder of this day you have. 👍

clicache/acceptance-test/SNITests.cs
```diff
@@ -127,5 +133,38 @@ public void ConnectionWithoutProxyFails()
 
             Assert.Throws<NotConnectedException>(() => region.Put("1", "one"));
         }
+
```
echobravopapa, line 136: It would be great to have this test:
- bring up the system/cluster
- do a cache ops (put/get)
- drop the proxy
- attempt cache ops - wait for exception (EXPECTEXCEPTION in gtest or similar...)
- bring proxy back up
- then do verify cache ops (put/get)

echobravopapa (changes requested):

clicache/acceptance-test/SNITests.cs
mmartell: I've updated the test to use timeouts instead of catching exceptions. If you stop the proxy, the socket is closed by the proxy and the client throws a "Not Connected exception with message "No locators available", which can be caught. However, this requires restarting the proxy on the same port, which I'm not sure how to do from C#. 

It seems just as good of a test to merely **pause** the proxy, which suspends the process and stops all network traffic. Then the proxy can merely be **unpaused**, which allows it to resume network communications on the same port. The only caveat to this simpler approach is that since the proxy doesn't actually close the socket, the client continues to retry the operation and never throws an exception.

pdxcodemonkey: This is probably a correct test until such time as we resolve GEODE-8628.  If it turns out NC should be honoring the read timeout value and throwing a timeout exception (what I would've expected), then when we fix it to do so properly, we can update this test to expect throw rather than just checking if the put is complete.  @echobravopapa I'm in favor of resolving this bug and merging the new test, then writing one to actually shut down and restart the proxy and catch the proper exception etc.  To my mind, these are different tests, both valid.  Let us know what you think.

mmartell: Figured out how to restart a stopped proxy on the original port. This allows us to catch the NotConnected exception just like in the Java client test.

mmartell:
This last commit aligns the C++ and the C# SNI tests. Also, the DropProxy tests now mimic the behavior of the Java SNI tests by catching the NotConnectedException whenever the proxy is dropped. Also, much cleaner code now..

echobravopapa (approved):

PR #675, opened 2020-10-14
GEODE-8593: Update geode-native examples to use builder pattern: C++

davebarnes97:
For the C++ native client (does not address .NET)
Modified examples and the docs that refer to them to show pool creation using the builder pattern.
Reduced the Using Connection Pools section by about half to its Configuration-related material (removed load-balancing).
Improved format of the example README files.
Built and ran all examples under MacOS (did not test under Windows).


docs/geode-native-docs-cpp/getting-started/getting-started-nc-client.html.md.erb
```diff
@@ -37,36 +37,37 @@ To connect to a server, your application must follow these steps:
...
-  cacheFactory.set("log-level", "none");          // set cache log-level characteristics
+auto cache = CacheFactory()
+    .set("log-level", "debug")
+    .set("ssl-enabled", "true")
+    .set("ssl-truststore", clientTruststore.string())
+    .create();
 ```
 
-Create a cache and use it to instantiate a `PoolFactory`:
+Create a pool of network connections, and instantiate a region of the desired type:
```
pdxcodemonkey, line 56: This should probably just read "Create a pool of network connections".  The next block is the one that instantiates the region.

pdxcodemonkey (approved):
Looks good, just edit the one description line and this is ready.

docs/geode-native-docs-cpp/getting-started/getting-started-nc-client.html.md.erb
davebarnes97: Good catch - thanks!

PR #676, opened 2020-10-15
GEODE-8613: Removing exclusive access to creation/closing of pool connections

albertogpz:
This PR removes the exclusive access to the creation/closing of pool connections that uses the mutex that is also used to grant exclusive access to the ConnectionQueue.
That way access to the connection queue is not impaired when there are problems creating connections.

### DIFF
--- a/cppcache/src/ThinClientPoolDM.cpp
+++ b/cppcache/src/ThinClientPoolDM.cpp
@@ -510,7 +510,6 @@ void ThinClientPoolDM::restoreMinConnections(std::atomic<bool>& isRunning) {
   int restored = 0;
 
   if (m_poolSize < min) {
-    std::lock_guard<decltype(mutex_)> lock(mutex_);
     while (m_poolSize < min && limit-- && isRunning) {
       TcrConnection* conn = nullptr;
       bool maxConnLimit = false;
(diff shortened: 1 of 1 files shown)


pivotal-jbarrett (changes requested):
I need more time to review this for correctness. I think the locking is required here to both protect concurrent updates to the `m_poolSize` value in various places as well as synchronize the CPU caches of this member.

albertogpz:
> I need more time to review this for correctness. I think the locking is required here to both protect concurrent updates to the `m_poolSize` value in various places as well as synchronize the CPU caches of this member.

Wouldn't that already be guaranteed given that the type of `m_poolSize` is `std::atomic<int32_t>`?

pivotal-jbarrett:
> > I need more time to review this for correctness. I think the locking is required here to both protect concurrent updates to the `m_poolSize` value in various places as well as synchronize the CPU caches of this member.
> 
> Wouldn't that already be guaranteed given that the type of `m_poolSize` is `std::atomic<int32_t>`?

I hadn't looked into the type yet but mostly yes that should be safe. Though if the pool and count are mutated independently there could be some issues with a mismatch. I will poke into this later today.

pivotal-jbarrett (approved):
Looks like we mutate both the poolsize and queue under lock still so I am cool. We might spuriously go into the connection creation methods now but that's ok for this simplification of the locking. This whole chunk of code is begging for a refactor. Perhaps something closer to what we did on the Java side.

mkevo (approved):

PR #678, opened 2020-10-22
GEODE-8601: Replace ExpiryTaskManager ACE implementation by boost::asio one

gaussianrecurrence:
- Replaced ExpiryTaskManager implementation. Now boost is used instead of ACE.
- Several parts of the code have been adapted to work with the new implementation.
- UTs implemented.
- ITs implemented.
- Cleanup unused code.
- Replaced system clock by steady clock as internal clock.


cppcache/include/geode/CacheStatistics.hpp
```diff
@@ -47,9 +47,9 @@ class LocalRegion;
  */
 class APACHE_GEODE_EXPORT CacheStatistics {
  public:
-  typedef std::chrono::system_clock::time_point time_point;
+  using time_point = std::chrono::steady_clock::time_point;
 
-  CacheStatistics() : m_lastAccessTime(0), m_lastModifiedTime(0) {}
+  CacheStatistics() = default;
```
mreddington, line 52: `default` will produce an inlined definition that will be compiled in client code for library code. This is just asking for type pruning problems, which we've seen before. This constructor should instead be moved into the source file. Default is fine for internal implementation, not for published headers.

cppcache/include/geode/CacheStatistics.hpp
```diff
@@ -97,11 +97,11 @@ class APACHE_GEODE_EXPORT CacheStatistics {
   virtual time_point getLastAccessedTime() const;
 
  private:
-  virtual void setLastAccessedTime(time_point lat);
-  virtual void setLastModifiedTime(time_point lmt);
+  virtual void setLastAccessedTime(const time_point& tp);
```
mreddington, line 100: A time_point boils down to maybe a `long long`, is there any pressing reason to pass by reference, and thus cascade a const reference type change to this interface through all the derived implementations in all client code?

cppcache/include/geode/CacheStatistics.hpp
```diff
@@ -97,11 +97,11 @@ class APACHE_GEODE_EXPORT CacheStatistics {
...
+  virtual void setLastModifiedTime(const time_point& tp);
 
-  std::atomic<time_point::duration::rep> m_lastAccessTime;
-  std::atomic<time_point::duration::rep> m_lastModifiedTime;
+  std::atomic<time_point::duration::rep> last_accessed_{0};
```
mreddington, line 103: This change also inlines into client code. Don't get me wrong, the original constructor was just as wrong, it should have just been moved into the source file. Save this sort of thing for internal implementation, not published headers.

cppcache/integration-test/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ add_library(test-cppcache-utils STATIC
 target_link_libraries(test-cppcache-utils
   PRIVATE
     ACE::ACE
+    Boost::boost
```
mreddington, line 28: Do you need all of Boost? You should be able to include just the libraries you need, Boost::asio, for example.

cppcache/integration-test/CMakeLists.txt
```diff
@@ -114,6 +115,7 @@ endif()
   target_link_libraries(${TEST}
     PRIVATE
       ACE::ACE
+      Boost::boost
```
mreddington, line 118: Same.

mreddington (changes requested):

cppcache/include/geode/CacheStatistics.hpp
gaussianrecurrence: Quite subtle catch here. Thanks for pointing it out! :)

cppcache/include/geode/CacheStatistics.hpp
gaussianrecurrence: And also, this might alter ABI signature. Something I had not into account by the time of preparing this PR. Thanks!

pivotal-jbarrett (changes requested):
This yet another awesome move away from ACE. I can appreciate the work put into this because I have tried and failed about 3 times myself. I need a bit more time to review the remaining source but right out the gate I am worried about the `clock` usage and change from `system_clock` to `steady_clock`. While I prefer we work internally in terms of `steady_clock` if any of the time is exchanged between processes or persisted it needs to be in `system_clock`.

cppcache/integration-test/CMakeLists.txt
gaussianrecurrence: After changing CacheStatistics this is not needed anymore. Thanks!

cppcache/integration-test/CMakeLists.txt
gaussianrecurrence: After changing CacheStatistics this is not needed anymore. Thanks!

clicache/src/DistributedSystem.cpp
```diff
@@ -404,9 +404,9 @@ namespace Apache
       void DistributedSystem::HandleMemoryPressure(System::Object^ state)
       {
         // TODO global - Need single memory pressue event running?
```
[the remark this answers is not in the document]
gaussianrecurrence: I am curious about why do we need this method at all. As far as I've seen it's not used anywhere

clicache/src/impl/MemoryPressureTask.hpp
```diff
@@ -30,14 +30,14 @@ namespace Apache
     namespace Client
     {
 
-      class MemoryPressureHandler
-        : public ACE_Event_Handler
+      class MemoryPressureTask
+        : public apache::geode::client::ExpiryTask
```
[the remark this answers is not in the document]
gaussianrecurrence: Thanks for pointing it. Seems that C# part does not have any auto-formatting :S

clicache/src/DistributedSystem.cpp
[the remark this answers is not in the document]
pivotal-jbarrett: It's likely meant to get invoked by .NET garbage collector but if they was never hooked up then I'd agree to remove it.

clicache/src/impl/MemoryPressureTask.hpp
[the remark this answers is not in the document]
pivotal-jbarrett: Worse... this is C++/CLI. As of the last time I tried about 3 years ago there wasn't a formatted they didn't choke on something. Most choked on generics. I should take another look.

We can format C# with clang-format but we have very little of that. I have a todo to hook it up some day.

cppcache/integration-test/CMakeLists.txt
gaussianrecurrence: After carefully looking at this I've noticed that new integration tests also include Boost library.
Thing is symbols are not needed from Boost library for integrations test, but compilation flags from the library.
Thing is in order to ignore clang-tidy checks for boost you need to include it as a system library, which is what's achieved by "linking" the library.

echobravopapa:
@gaussianrecurrence please rebase this branch so the Checks will run again, thx

cppcache/include/geode/CacheStatistics.hpp
gaussianrecurrence: > This yet another awesome move away from ACE. I can appreciate the work put into this because I have tried and failed about 3 times myself. I need a bit more time to review the remaining source but right out the gate I am worried about the `clock` usage and change from `system_clock` to `steady_clock`. While I prefer we work internally in terms of `steady_clock` if any of the time is exchanged between processes or persisted it needs to be in `system_clock`.

After looking into it, CacheStatistics expects that an system_clock::time_point. This might potentially be used by any NC user to get modification and access time. But internally this stats are used to calculate elapsed time rather than a fixed time point. So what I did is to internally store it as steady_clock::time_point and public getters converts it to system_clock::timepoint by calculating the delta between both clocks.

clicache/src/CacheStatistics.cpp
```diff
@@ -32,7 +32,9 @@ namespace Apache
       {
         try
         {
-          return TimeUtils::TimePointToDateTime(m_nativeptr->get()->getLastModifiedTime());
+          auto epoch = m_nativeptr->get()->getLastModifiedTime().time_since_epoch();
```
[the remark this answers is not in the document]
gaussianrecurrence: Explained in my comment below

clicache/src/DistributedSystem.cpp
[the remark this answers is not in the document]
gaussianrecurrence: Removed

clicache/src/impl/MemoryPressureTask.hpp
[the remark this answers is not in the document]
gaussianrecurrence: This was removed as was no plug in, hence not used.

cppcache/src/CacheStatistics.cpp
```diff
@@ -21,20 +21,20 @@ namespace apache {
 namespace geode {
 namespace client {
 
-void CacheStatistics::setLastModifiedTime(time_point lmt) {
-  m_lastModifiedTime = lmt.time_since_epoch().count();
+void CacheStatistics::setLastModifiedTime(const time_point& tp) {
+  last_modified_ = tp.time_since_epoch().count();
```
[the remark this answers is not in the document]
gaussianrecurrence: Explained in my comment below

cppcache/src/EntryExpiryTask.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  entry_->getKeyI(key);
+
+  const auto full_path = region_->getFullPath().c_str();
+  auto key_str = Utils::nullSafeToString(key);
+
+  switch (action_) {
+    case ExpirationAction::INVALIDATE: {
+      LOGDEBUG(
+          "EntryExpiryTask::DoTheExpirationAction INVALIDATE "
```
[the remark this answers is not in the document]
gaussianrecurrence: Done

gaussianrecurrence:
> @gaussianrecurrence please rebase this branch so the Checks will run again, thx

Already rebased it, but until Geode version does not get updated to 1.13.2 it LGTM analysis will keep failing. I am curious, why minor Geode versions tar.gz's are removed once a new one comes out?

pdxcodemonkey:
> > @gaussianrecurrence please rebase this branch so the Checks will run again, thx
> 
> Already rebased it, but until Geode version does not get updated to 1.13.2 it LGTM analysis will keep failing. I am curious, why minor Geode versions tar.gz's are removed once a new one comes out?

No idea why.  This is likely an Apache convention, as so many things are.  I merged the PR to update the Geode version in our Docker images, and it appears LGTM is back in business.

pdxcodemonkey (approved):

PR #680, opened 2020-10-28
GEODE-8666:  Enforce no-non-virtual-dtor

moleske:

tagging @gaussianrecurrence & @mreddington if you'd like to review also

Passing tests
[Ubuntu - clang, units, integration, acceptance, ipv6](https://github.com/moleske/geode-native/actions/runs/332211481)
[Mac - units, integration, acceptance](https://github.com/moleske/geode-native/actions/runs/332211479)
[Windows units, integration, acceptance (cli and cpp)](https://github.com/moleske/geode-native/actions/runs/332211482)

### DIFF
--- a/cppcache/include/geode/AuthenticatedView.hpp
+++ b/cppcache/include/geode/AuthenticatedView.hpp
@@ -90,7 +90,7 @@ class APACHE_GEODE_EXPORT AuthenticatedView : public RegionService {
   /**
    * @brief destructor
    */
-  virtual ~AuthenticatedView() override;
+  ~AuthenticatedView() override;
 
   /**
    * @brief constructors
--- a/cppcache/include/geode/AuthenticatedView.hpp
+++ b/cppcache/include/geode/AuthenticatedView.hpp
@@ -90,7 +90,7 @@ class APACHE_GEODE_EXPORT AuthenticatedView : public RegionService {
   /**
    * @brief destructor
    */
-  virtual ~AuthenticatedView();
+  virtual ~AuthenticatedView() override;
 
   /**
    * @brief constructors
--- a/cppcache/include/geode/CqQuery.hpp
+++ b/cppcache/include/geode/CqQuery.hpp
@@ -53,6 +53,8 @@ class Query;
  */
 class APACHE_GEODE_EXPORT CqQuery {
  public:
+  virtual ~CqQuery() = default;
+
   /**
    * Get the query string provided when a new Query was created from a
    * QueryService.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,7 +244,6 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     -Wno-used-but-marked-unused #TODO fix
     -Wpedantic
     -Weverything
-    -Wno-non-virtual-dtor #TODO fix
     -Wno-missing-prototypes #TODO fix
     -Wno-deprecated #TODO fix
     -Wno-format-nonliteral #TODO fix
--- a/cppcache/include/geode/Cache.hpp
+++ b/cppcache/include/geode/Cache.hpp
@@ -261,7 +261,7 @@ class APACHE_GEODE_EXPORT Cache : public GeodeCache {
   LogLevel getLogLevel();
 
   Cache() = delete;
-  virtual ~Cache();
+  virtual ~Cache() override;
   Cache(const Cache& other) = delete;
   Cache& operator=(const Cache& other) = delete;
   Cache(Cache&& other) noexcept;
--- a/cppcache/include/geode/CqServiceStatistics.hpp
+++ b/cppcache/include/geode/CqServiceStatistics.hpp
@@ -38,6 +38,8 @@ namespace client {
  */
 class APACHE_GEODE_EXPORT CqServiceStatistics {
  public:
+  virtual ~CqServiceStatistics() = default;
+
   /**
    * Get the number of CQs currently active.
    * Active CQs are those which are executing (in running state).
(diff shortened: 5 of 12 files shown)


cppcache/include/geode/AuthenticatedView.hpp
```diff
@@ -90,7 +90,7 @@ class APACHE_GEODE_EXPORT AuthenticatedView : public RegionService {
   /**
    * @brief destructor
    */
-  virtual ~AuthenticatedView();
+  virtual ~AuthenticatedView() override;
```
pivotal-jbarrett, line 93: `virtual` is redundant when `override` is specified. Under some analyzers it will warn or error about it.

pivotal-jbarrett: Given that this is a public header, what impact is there to the ABI?

cppcache/include/geode/CqQuery.hpp
```diff
@@ -53,6 +53,8 @@ class Query;
  */
 class APACHE_GEODE_EXPORT CqQuery {
  public:
+  virtual ~CqQuery() = default;
```
pivotal-jbarrett, line 56: This will for sure change the ABI. I am not sure we can do this on the public headers until the next major.

pivotal-jbarrett (changes requested):

cppcache/include/geode/AuthenticatedView.hpp
gaussianrecurrence: In this case the change is not breaking ABI BC as override is just a keyword introduced in the standard in order to have the compiler checked that the virtual function you are overriding has the same signature as the one on the base class.

gaussianrecurrence:
I am running an [ABI compliance tool](https://lvc.github.io/abi-compliance-checker/) having compile both the develop branch and the PR's branch and oddly it states that changes introduced are ABI backward compatible. I will try to run the same check with Clang-6.0 and onwards so as well as win MSVC.

As stated by @pivotal-jbarrett, exposing base class destructors into vtable should be an ABI breaking change.

pivotal-jbarrett:
> I am running an [ABI compliance tool](https://lvc.github.io/abi-compliance-checker/)

This is awesome. Perhaps we can add this tooling to the Travis checks we run on PRs.

gaussianrecurrence:
> I am running an [ABI compliance tool](https://lvc.github.io/abi-compliance-checker/) having compiled both the develop branch and the PR's branch and oddly it states that changes introduced are ABI backward compatible. I will try to run the same check with Clang-6.0 and onwards so as well as with MSVC.
> 
> As stated by @pivotal-jbarrett, exposing base class destructors into vtable should be an ABI breaking change.

For some reason in both GCC/Clang seems to have no impact. I've opened an [stackoverflow question](https://stackoverflow.com/questions/64576696/exposing-a-base-class-destructor-as-virtual-means-breaking-abi-backward-compatib) just to clarify it, but in the case of MSVC it does. For example for the case of CqQuery vtable is shifted by one position...

cppcache/include/geode/AuthenticatedView.hpp
moleske: whoops, missed that on my pass.  That's what reviews are for

moleske:
Ugh ABI.  I never think about it cause it's not in my head part of the contract for breaking changes (just API changes).  I'll close this after @gaussianrecurrence reports back with any learnings from why the ABI compliance tool says it is fine (when I also agree it is not).  Will also update the JIRA story to reflect that it can't be done yet

Tangential question, what version is native client now?  Geode server is at 1.13 but is that what native client is?  So to release a major version of native client we need to wait til Geode makes a 2.0 release?  That seems not great.  I've never understood why geode native and geode are release with the same versioning scheme...

Also, another tangential question, why are we saying ABI breakage requires a major version change?  Might be worth writing that down somewhere (unless it is already in the wiki, which I never check).  Assuming because it requires consumers of geode native to recompile, but are there other reasons?

pivotal-jbarrett:
> Also, another tangential question, why are we saying ABI breakage requires a major version change? Might be worth writing that down somewhere (unless it is already in the wiki, which I never check). Assuming because it requires consumers of geode native to recompile, but are there other reasons?

This is a semantic versioning thing. An ABI breaking change would require a recompilation of the downstream application. Normally I can drop a minor updated dependency library in place and just restart the downstream application, no recompilation required. An ABI change requires a recompilation, even if the API is unchanged, because the behavior will now be unknown between the downstream application and the library. In this particular case which side of the library boundary generated the destructor and is it in a VTable or not. It is possible that the compiler did the right thing, in seeing that there were virtual methods on class it behind the scenes produced a virtual destructor and exported it as such in the ABI.

pivotal-jbarrett:
> Tangential question, what version is native client now? Geode server is at 1.13 but is that what native client is? So to release a major version of native client we need to wait til Geode makes a 2.0 release? That seems not great. I've never understood why geode native and geode are release with the same versioning scheme...

This is an ASF thing. The geode-native repository is not a separate project or sub-project of Geode. As such its part of the Geode project and is released with the Geode project, currently only as source, at each Geode release. So yes, for a breaking change Geode project needs to release a major. Now this is strict interpretation of the ASF project rules, blah blah blah... Once could probably interpret them differently and version each module produces under the project independently. One could define a semantic version independent of the project version. All great discussion topics for the dev list probably.

gaussianrecurrence:
> I'll close this after @gaussianrecurrence reports back with any learnings from why the ABI compliance tool says it is fine (when I also agree it is not). Will also update the JIRA story to reflect that it can't be done yet

Sadly nobody seems to know the reason. I guess at least from my side will have to remain a mistery. If you happen to ever find out why, please let me know :S

moleske:
Closing this now.  Shame there wasn't a better answer to why the tool wasn't working.  I don't necessarily still agree with why ABI needs to be maintained til a major, but I also don't care enough since I'm currently not paid to work on this (apparently I just really like removing skipped compiler warnings 🤷‍♀️.)  Same for making a 2.0 geode native.  I'll look to set up automation on my branch so that when a decision to make a major version happens, this can be contributed

PR #682, opened 2020-10-29
GEODE-8647: Stop leaking buffer in CLI DataInput 

pdxcodemonkey:
The CLI DataInput object has two ctors, one of which copies the passed-in buffer parameter via new[] and one of which doesn't.  In the event that the former is called, the buffer is leaked when the object is deleted/Disposed.

@mreddington @dihardman @karensmolermiller @davebarnes97

### DIFF
--- a/clicache/src/DataInput.cpp
+++ b/clicache/src/DataInput.cpp
@@ -132,11 +132,10 @@ namespace Apache
               "DataInput.ctor(): given length {0} is zero or greater than "
               "size of buffer {1}", len, buffer->Length));
           }
-          //m_bytes = gcnew array<Byte>(len);
-          //System::Array::Copy(buffer, 0, m_bytes, 0, len);
           _GF_MG_EXCEPTION_TRY2
 
-            _GEODE_NEW(m_buffer, System::Byte[len]);
+          m_ownedBuffer = make_native_unique<System::Byte[]>(len);
+          m_buffer = m_ownedBuffer->get();
           pin_ptr<const Byte> pin_buffer = &buffer[0];
           memcpy(m_buffer, (void*)pin_buffer, len);
           m_nativeptr = gcnew native_conditional_unique_ptr<native::DataInput>(
@@ -878,8 +877,6 @@ namespace Apache
 
       void DataInput::Cleanup()
       {
-        //TODO:
-        //GF_SAFE_DELETE_ARRAY(m_buffer);
       }
 
       void DataInput::ReadDictionary(System::Collections::IDictionary^ dict)
--- a/clicache/src/DataInput.cpp
+++ b/clicache/src/DataInput.cpp
@@ -876,10 +876,6 @@ namespace Apache
         SetBuffer();
       }
 
-      void DataInput::Cleanup()
-      {
-      }
-
       void DataInput::ReadDictionary(System::Collections::IDictionary^ dict)
       {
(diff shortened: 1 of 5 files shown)


clicache/src/DataInput.hpp
```diff
@@ -691,6 +691,7 @@ namespace Apache
         bool m_isRootObjectPdx;
         Apache::Geode::Client::Cache^ m_cache;
         System::Byte* m_buffer;
+        System::Byte* m_allocatedBuffer;
```
mmartell, line 694: My only issue is that your new pointer is really just used as a flag. But still seems clear as to its intend.

mmartell (approved):

clicache/src/DataInput.cpp
```diff
@@ -90,11 +91,13 @@ namespace Apache
         m_ispdxDesrialization = false;
         m_isRootObjectPdx = false;
         m_cache =  cache;
+        m_allocatedBuffer = nullptr;
         if (buffer != nullptr && buffer->Length > 0) {
           _GF_MG_EXCEPTION_TRY2
 
             System::Int32 len = buffer->Length;
           _GEODE_NEW(m_buffer, System::Byte[len]);
```
pivotal-jbarrett, line 99: This should actually be invoking:
```c++
m_buffer = gcnew array<System::Byte>(len);

```

pivotal-jbarrett (changes requested):
Use `std::unique_ptr`.

clicache/src/DataInput.cpp
```diff
@@ -878,8 +877,6 @@ namespace Apache
 
       void DataInput::Cleanup()
       {
```
pivotal-jbarrett, line 879: If this method is empty perhaps we can remove it?

clicache/src/DataInput.hpp
```diff
@@ -663,7 +664,7 @@ namespace Apache
           m_buffer = const_cast<System::Byte*>(nativeptr->currentBufferPosition());
           if ( m_buffer != NULL) {
             m_bufferLength = static_cast<decltype(m_bufferLength)>(nativeptr->getBytesRemaining());
-					}
+          }
```
pivotal-jbarrett, line 667: Some strange formatting mess here.

pivotal-jbarrett (approved):
Just some minor cleanup related stuff otherwise I like the use of smart pointers.

clicache/src/DataInput.hpp
pdxcodemonkey: Appears to be okay now.

clicache/src/DataInput.cpp
pdxcodemonkey: Done

clicache/src/DataInput.cpp
```diff
@@ -93,8 +93,9 @@ namespace Apache
         if (buffer != nullptr && buffer->Length > 0) {
           _GF_MG_EXCEPTION_TRY2
 
-            System::Int32 len = buffer->Length;
-          _GEODE_NEW(m_buffer, System::Byte[len]);
+          System::Int32 len = buffer->Length;
```
pivotal-jbarrett, line 96: auto?

pivotal-jbarrett (approved):

echobravopapa (approved):

PR #684, opened 2020-10-30
GEODE-8678: Add missing API docs

mmartell:
Add missing documentation for Cache::GetPoolManager() and Cache::GetPoolFactory().


moleske (approved):
Seems fine.  I haven't looked at docs for cli side for a long time.  Any reason the `PoolManager` description isn't wrapped in a `<summary></summary>` like `PoolFactory`?  Also looks like `PoolManager` is missing a `@return`

PR #685, opened 2020-11-04
GEODE-8676: Update Bookbindery

davebarnes97:
One change that I think should have been included in the previous PR for this ticket.
Adds a `bundle exec` prefix before `rackup` in the view-docs.sh script. Should improve the user experience.


moleske (approved):
Looks good! Finally got it up and running myself!

yasser3399 (approved):

PR #686, opened 2020-11-05
GEODE-8688: Fix flaky C++ native client integration tests

albertogpz:
The following integration test cases under
integration/test (new integration tests)
ar flaky (do not
fail normally when run locally but fail very often
when run in CI).

- PartitionRegionOpsTest.getPartitionedRegionWithRedundancyServerGoesDownSingleHop
- PartitionRegionOpsTest.putPartitionedRegionWithRedundancyServerGoesDownSingleHop

There were two reasons that can make them fail.

One of them is that sometimes the connections to the server have expired
before the server is restarted and therefore, when traffic is sent
to the restarted server, no errors are found. To fix this,
the pool configuration for the test client
has been changed so that connections do not expire.

The other reason is that sometimes the error in the connection is
found by the ping thread that is invoking the
ThinClientPoolDM::sendRequestToEP() method and in this method,
when the IO error or TIMEOUT error are encountered,
the endpoint is not removed from the metadata (by means of the
removeBucketServerLocation method).
The code has been updated to remove the metadata also in this
case.

With these two changes, the test cases are not flaky anymore.

### DIFF
--- a/cppcache/integration/test/PartitionRegionOpsTest.cpp
+++ b/cppcache/integration/test/PartitionRegionOpsTest.cpp
@@ -146,9 +146,9 @@ void verifyMetadataWasRemovedAtFirstError() {
       }
     }
   }
-  ASSERT_EQ(timeoutErrors, metadataRemovedDueToTimeout);
-  ASSERT_EQ(ioErrors, metadataRemovedDueToIoErr);
-  ASSERT_NE(metadataRemovedDueToTimeout, metadataRemovedDueToIoErr);
+  EXPECT_EQ(timeoutErrors, metadataRemovedDueToTimeout);
+  EXPECT_EQ(ioErrors, metadataRemovedDueToIoErr);
+  EXPECT_NE(metadataRemovedDueToTimeout, metadataRemovedDueToIoErr);
 }
 
 void putPartitionedRegionWithRedundancyServerGoesDown(bool singleHop) {
--- a/cppcache/integration/test/PartitionRegionOpsTest.cpp
+++ b/cppcache/integration/test/PartitionRegionOpsTest.cpp
@@ -74,6 +74,8 @@ std::shared_ptr<Pool> createPool(Cluster& cluster, Cache& cache,
   auto poolFactory = cache.getPoolManager().createFactory();
   cluster.applyLocators(poolFactory);
   poolFactory.setPRSingleHopEnabled(singleHop);
+  poolFactory.setLoadConditioningInterval(std::chrono::milliseconds::zero());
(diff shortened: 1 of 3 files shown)


cppcache/integration/test/PartitionRegionOpsTest.cpp
```diff
@@ -144,6 +146,10 @@ void verifyMetadataWasRemovedAtFirstError() {
       }
     }
   }
+  std::cout << "timeoutErrors: " << timeoutErrors << ", ioErrors: " << ioErrors
+            << ", metadataRemovedDueToTimeout: " << metadataRemovedDueToTimeout
+            << ", metadataRemovedDueToIoErr: " << metadataRemovedDueToIoErr
+            << std::endl;
```
pdxcodemonkey, line 152: If you really want to see this output when running your test, I believe it's best to use std::cerr rather than cout.  This looks like leftover debugging trace stuff to me, tho.

pdxcodemonkey (changes requested):
Looks good, aside from the stray (I think) debug output.

cppcache/integration/test/PartitionRegionOpsTest.cpp
pivotal-jbarrett: Tests should never output anything other than assertion failures. If more information is useful to log then it is useful to assert.

pivotal-jbarrett (changes requested):

pdxcodemonkey (approved):

cppcache/integration/test/PartitionRegionOpsTest.cpp
```diff
@@ -144,9 +146,9 @@ void verifyMetadataWasRemovedAtFirstError() {
       }
     }
   }
-  ASSERT_TRUE((timeoutErrors == metadataRemovedDueToTimeout) &&
-              (ioErrors == metadataRemovedDueToIoErr) &&
-              (metadataRemovedDueToTimeout != metadataRemovedDueToIoErr));
+  ASSERT_EQ(timeoutErrors, metadataRemovedDueToTimeout);
```
pivotal-jbarrett, line 149: Alternatively, if you want to see the results of all three checks and fail if any of them fail you can use `EXPECT_*` macros. There are also more options for expecting and asserting in the `gmock` module.

pivotal-jbarrett (approved):

cppcache/integration/test/PartitionRegionOpsTest.cpp
albertogpz: Thanks, I did not know.

pivotal-jbarrett (approved):

PR #688, opened 2020-11-05
GEODE-8614: Add LowMemoryException exceptions

gaussianrecurrence:
- Added LowMemoryException to be thrown in the client whenever the server
  runs out of memory.
- Added QueryExecutionLowMemoryException to be thrown in the client whenever
  the monitoring queries feature on the server detects that the member is running
  low on memory.
- Added UTs to verity error to exception translation is working.
- Added new integration tests for both exceptions.


gaussianrecurrence:
Maybe @moleske @pivotal-jbarrett @mreddington @pdxcodemonkey you could throw some comments?

Thanks!

cppcache/integration/test/ExceptionTranslationTest.cpp
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+}  // namespace
+
+TEST(ExceptionTranslationTest, testLowMemoryException) {
+  const auto ENTRY_SIZE = 1024U;
+  const auto ENTRIES = 1U << 20U;
+
+  Cluster cluster{LocatorCount{1}, ServerCount{1},
+                  CacheXMLFiles{std::vector<std::string>{
```
pivotal-jbarrett, line 92: Outside the scope of this fix but interesting thought. I wonder if this could take an initializer list to magically construct the `std::vector<std::string>` for readability.

pivotal-jbarrett (approved):

cppcache/src/ExceptionTypes.cpp
```diff
@@ -23,30 +23,61 @@
...
-namespace client {
-
-void setThreadLocalExceptionMessage(std::string exMsg);
-const std::string& getThreadLocalExceptionMessage();
-
-static std::map<GfErrType, std::function<void(std::string, std::string&,
-                                              GfErrType, std::string)>>
-    error_map;
+namespace {
+
```
pivotal-jbarrett, line 27: Wait, why did you pull this out of the namespace and then import all these exceptions?

pivotal-jbarrett (changes requested):

cppcache/src/ExceptionTypes.cpp
gaussianrecurrence: I am glad you noticed. There is a subtle, still pretty good reason for this.
I've made a gist demonstrating my point: https://gist.github.com/gaussianrecurrence/cda4f6c16b2652e09b539ae9c2bf6e92

Summarizing, basically there is a non-written rule that states: "If you implement free functions that are to be used only within the module implementing them, then put them in a anonymous namespace"

That's done in order to avoid naming conflicts.

cppcache/integration/test/ExceptionTranslationTest.cpp
gaussianrecurrence: I changed the cluster initialization so it improves readability. Hopefully that was what you went for with your comment :)

pdxcodemonkey (approved):
Looks good to me.  Especially appreciate the cleanup of get_error_map - so much nicer now, thanks!

pivotal-jbarrett (approved):

PR #689, opened 2020-11-10
GEODE-8593: Update NC examples to use Builder pattern (.NET)

davebarnes97:
Also addresses GEODE-8431: Update Connection Pools topic

Analog of an update performed previously on the geode-native C++ examples and user guide:
- Updated examples to use builder pattern for cache and pool creation
- Updated accompanying text in user guide to match examples
- Pruned Connection Pool doc to focus on pools, not locators and load-balancing


docs/geode-native-docs-cpp/regions/regions.html.md.erb
```diff
@@ -53,20 +53,20 @@ auto cache = CacheFactory()
 
 cache.getPoolManager()
     .createFactory()
-    .addLocator("localhost", 10334)
-    .create("pool");
+    .addLocator("localhost", 40404)
```
mmartell, line 56: I would stick with the port 10334 here, which is the default locator port. 40404 is the default server port, so looks very odd to use as a locator port.

mmartell (commented):
In general I like simplification of our docs by removing sections (particularly for seldom used features). However, we should consider adding a link to an Appendix: Other Features (or similar) or a link to other documents that discuss those features.

davebarnes97:
> In general I like simplification of our docs by removing sections (particularly for seldom used features). However, we should consider adding a link to an Appendix: Other Features (or similar) or a link to other documents that discuss those features.

@mmartell I get it, but I think it's beyond the scope of this PR. This PR does add a "Configuring Pools" topic that did not appear in the previous version, but that was a side-effect of changing the connection pool code in all the examples.

I welcome suggestions regarding topics that should be added. Feel free to contact me directly or open a JIRA ticket with a list so your ideas don't get lost. Thanks!

docs/geode-native-docs-cpp/regions/regions.html.md.erb
davebarnes97: Corrected locator port (40404 -> 10334) fo both C++ and .NET.

mmartell:
Looks good. Thanks for clarifying that this also address GEODE-8431.

PR #690, opened 2020-11-11
GEODE-8693: throw NotConnectedException if server goes down while exe…

albertogpz:
…cute with onServers

When a function is executed by using onServers, if one of the servers goes down while the function is executed, no exception is thrown (despite what the Native C++ API documentation says).
The reason was that when an IO error was detected when calling a function on a server, the FunctionService::execute() function instead of returning the IO error returned no error.
With this change this function now returns the IO error which translates in a NotConnectedException thrown.

### DIFF
--- a/cppcache/integration/test/FunctionExecutionTest.cpp
+++ b/cppcache/integration/test/FunctionExecutionTest.cpp
@@ -183,7 +183,9 @@ const std::vector<std::string> serverResultsToStrings(
     for (decltype(resultArray->size()) i = 0; i < resultArray->size(); i++) {
       auto value =
           std::dynamic_pointer_cast<CacheableString>(resultArray->at(i));
-      resultList.push_back(value->toString());
+      if (value != nullptr) {
+        resultList.push_back(value->toString());
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


cppcache/integration/test/FunctionExecutionTest.cpp
```diff
@@ -251,3 +253,79 @@ TEST(FunctionExecutionTest, OnServersWithReplicatedRegionsInPool) {
...
+
+  for (int i = 0; i < ON_SERVERS_TEST_REGION_ENTRIES_SIZE; i++) {
+    region->put("KEY--" + std::to_string(i), "VALUE--" + std::to_string(i));
+  }
+
+  auto threadAux = std::make_shared<std::thread>(
+      executeTestFunctionOnLoopAndExpectNotConnectedException, pool);
+
+  // Sleep a bit to allow for some successful responses before the exception
+  sleep(5);
```
pdxcodemonkey, line 326: This doesn't build on Windows:
```
C:\nativeclient\cppcache\integration\test\FunctionExecutionTest.cpp(326): error C3861: 'sleep': identifier not found 
```
Looks like we use `std::this_thread::sleep_for` in the rest of the code

cppcache/src/ExecutionImpl.cpp
```diff
@@ -429,33 +429,20 @@ void ExecutionImpl::executeOnAllServers(const std::string& func,
...
-  }
-
-  if (err == GF_AUTHENTICATION_FAILED_EXCEPTION ||
-      err == GF_NOT_AUTHORIZED_EXCEPTION ||
-      err == GF_AUTHENTICATION_REQUIRED_EXCEPTION) {
-    throwExceptionIfError("Execute", err);
-  }
 
   if (err != GF_NOERR) {
+    LOGDEBUG("Execute errorred: %d", err);
```
pdxcodemonkey, line 434: Sorry to be "that guy" about grammar, but errorred isn't really a word.  Should be "erred" or just avoid it altogether and use "failed".  Thx

pdxcodemonkey: Ugh, and it was even like this to begin with - yikes, sorry.

pdxcodemonkey (changes requested):
Looks good, thanks for the fix, and as always a nice new set of tests.  Just need to fix the Windows build and my picky grammar issue.

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -670,11 +670,12 @@ GfErrType ThinClientPoolDM::sendRequestToAllServers(
         err == GF_NOT_AUTHORIZED_EXCEPTION ||
         err == GF_AUTHENTICATION_REQUIRED_EXCEPTION) {
       finalErrorReturn = err;
-    } else if (!(finalErrorReturn == GF_AUTHENTICATION_FAILED_EXCEPTION ||
-                 finalErrorReturn == GF_NOT_AUTHORIZED_EXCEPTION ||
-                 finalErrorReturn ==
-                     GF_AUTHENTICATION_REQUIRED_EXCEPTION))  // returning auth
-                                                             // errors
+    } else if ((err != GF_NOERR) &&
```
echobravopapa, line 673: the check for `!GF_NOERR` looks like a good addition

cppcache/src/ThinClientPoolDM.cpp
```diff
@@ -2414,7 +2415,7 @@ GfErrType FunctionExecution::execute() {
   m_error = m_poolDM->sendRequestToEP(request, reply, m_ep);
   m_error = m_poolDM->handleEPError(m_ep, reply, m_error);
   if (m_error != GF_NOERR) {
-    if (m_error == GF_NOTCON || m_error == GF_IOERR) {
+    if (m_error == GF_NOTCON) {
```
echobravopapa, line 2418: how did you determine we no longer need to check for`GF_IOERR`?

cppcache/src/ExecutionImpl.cpp
albertogpz: No problem. It looked weird to me too. I even thought the author might want to say something like "error red" :-).

cppcache/src/ThinClientPoolDM.cpp
albertogpz: Removing that check is the way to signal the caller of this function (by means of the return in line 2428) that there has been an IO error so that the NotConnectedException is thrown to the client. With the original check, IO errors were ignored.

pdxcodemonkey (approved):

echobravopapa (approved):

PR #691, opened 2020-11-11
GEODE-8700: Add the ability to change PR

gaussianrecurrence:
- Added setPartitionResolver method to region's attribute mutator, so the
   partition resolver can be changed after the region has been created.
 - Added UT to verify that PR is correctly changed.

### DIFF
--- a/cppcache/include/geode/AttributesMutator.hpp
+++ b/cppcache/include/geode/AttributesMutator.hpp
@@ -197,16 +197,6 @@ class APACHE_GEODE_EXPORT AttributesMutator {
    * @param resolver PartitionResolver
    */
   void setPartitionResolver(std::shared_ptr<PartitionResolver> resolver);
-
-  /** Sets cache writer for region. The previous partition resolver will be
-   * replaced with the one created using the factory function provided in
-   * the given library.
-   * @param library_path path of the library containing partition resolver
-   * factory function.
-   * @param function factory function for creating a partition resolver.
-   */
-  void setPartitionResolver(const std::string& library_path,
-                            const std::string& function);
 };
 
 }  // namespace client
--- a/cppcache/include/geode/AttributesMutator.hpp
+++ b/cppcache/include/geode/AttributesMutator.hpp
@@ -40,6 +40,7 @@ class CacheListener;
 class CacheLoader;
 class CacheWriter;
 class Region;
+class PartitionResolver;
 
 /**
  * @class AttributesMutator AttributesMutator.hpp
@@ -190,6 +191,22 @@ class APACHE_GEODE_EXPORT AttributesMutator {
    */
   void setCacheWriter(const std::string& libpath,
                       const std::string& factoryFuncName);
+
+  /** Sets partition resolver writer for region. The previous partition resolver
+   * will be replaced with <code>resolver</code>.
+   * @param resolver PartitionResolver
+   */
+  void setPartitionResolver(std::shared_ptr<PartitionResolver> resolver);
+
+  /** Sets cache writer for region. The previous partition resolver will be
+   * replaced with the one created using the factory function provided in
+   * the given library.
+   * @param library_path path of the library containing partition resolver
+   * factory function.
+   * @param function factory function for creating a partition resolver.
+   */
+  void setPartitionResolver(const std::string& library_path,
+                            const std::string& function);
 };
 
 }  // namespace client
--- a/cppcache/src/AttributesMutator.cpp
+++ b/cppcache/src/AttributesMutator.cpp
@@ -122,12 +122,6 @@ void AttributesMutator::setPartitionResolver(
   auto impl = std::static_pointer_cast<RegionInternal>(m_region);
   impl->adjustPartitionResolver(std::move(resolver));
 }
-
-void AttributesMutator::setPartitionResolver(
-    const std::string& libpath, const std::string& factoryFuncName) {
-  auto impl = std::static_pointer_cast<RegionInternal>(m_region);
-  impl->adjustPartitionResolver(libpath, factoryFuncName);
-}
 }  // namespace client
 }  // namespace geode
 }  // namespace apache
--- a/cppcache/src/AttributesMutator.cpp
+++ b/cppcache/src/AttributesMutator.cpp
@@ -116,6 +116,18 @@ void AttributesMutator::setCacheWriter(const std::string& libpath,
   auto rImpl = std::static_pointer_cast<RegionInternal>(m_region);
   rImpl->adjustCacheWriter(libpath.c_str(), factoryFuncName.c_str());
 }
+
+void AttributesMutator::setPartitionResolver(
+    std::shared_ptr<PartitionResolver> resolver) {
+  auto impl = std::static_pointer_cast<RegionInternal>(m_region);
+  impl->adjustPartitionResolver(std::move(resolver));
+}
+
+void AttributesMutator::setPartitionResolver(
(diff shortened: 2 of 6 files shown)


gaussianrecurrence:
Maybe @moleske @pivotal-jbarrett @mreddington @pdxcodemonkey you could throw some comments?

Thanks!

cppcache/include/geode/AttributesMutator.hpp
```diff
@@ -190,6 +191,22 @@ class APACHE_GEODE_EXPORT AttributesMutator {
...
+  void setPartitionResolver(std::shared_ptr<PartitionResolver> resolver);
+
+  /** Sets cache writer for region. The previous partition resolver will be
+   * replaced with the one created using the factory function provided in
+   * the given library.
+   * @param library_path path of the library containing partition resolver
+   * factory function.
+   * @param function factory function for creating a partition resolver.
+   */
+  void setPartitionResolver(const std::string& library_path,
```
pivotal-jbarrett, line 208: Unless this is explicitly needed by your use case, and I really hope it isn't, let's avoid expanding on this API to load things from arbitrary libraries.

pivotal-jbarrett (changes requested):

cppcache/include/geode/AttributesMutator.hpp
gaussianrecurrence: Indeed, I don't actually need the library one. Good catch :) will remove it!

pivotal-jbarrett:
@gaussianrecurrence on further inspection I have concerns about the methods being added to to the internal region since they only apply to partitioned regions. I am also concerned about the thread safety of changing the resolver after the region is initialized. I have added @pdxcodemonkey as a reviewer because I don't have the bandwidth right now to investigate this. I would also suggest you look at the Java implementation for inspiration. Can the Java version mutate the partition resolver?

gaussianrecurrence:
> @gaussianrecurrence on further inspection I have concerns about the methods being added to to the internal region since they only apply to partitioned regions. I am also concerned about the thread safety of changing the resolver after the region is initialized. I have added @pdxcodemonkey as a reviewer because I don't have the bandwidth right now to investigate this. I would also suggest you look at the Java implementation for inspiration. Can the Java version mutate the partition resolver?

Regarding your concern about adding this to RegionInternal, thing is as I've seen there is no distinction in NC between partitioned regions and no-partitioned regions in terms of modularization, please correct me if I am wrong.

---

Regarding thread safety, it turns out you are right, I originally thought that calling reset/operator= on a shared_ptr was thread-safe, but this is not the case, as I've read. So that shared_ptr will need to be protected by a RW lock. Other than that, is thread safe and let me explain. If **thread A** takes a copy of PR1 and **thread B**, replaces the PR by PR2, any thread requesting the PR will get PR2 after that, so it would happen that one requests are using one PR and others the new one, but that's what's expected.

---

And as for the Java API, I don't think I've seen any implementation allowing to change PR on the fly. Still, I would like to point out that Java has reflection, which currently C++ lacks of. The reason why I bring reflection up is because, the only way right now to specify a PR while declaratively defining a cache is by creating a separate shared library which is not quite much elegant.

My first idea for implementing this was to introduce something like a generic "ExtensionRegistry" in which you could register "Extensions" uniquely identified by a name. But the problem with that approach was that the change was too big and implied several additions to the API, which as of we are all aware, is not a good thing.

So that's why I went for the attributes mutator approach.

pivotal-jbarrett:
I am not sure I understand the motivation. Are you saying that you want to define the region in `cache.xml` and then later mutate it from the API? I am not sure how safe this would be given the general immutable nature of the region objects. What is the motivation to split the configuration this way. Our general recommendation is to use the API over the XML configuration anyway. The server is moving away from the XML configuration to cluster config and API config. The API provides the ability to provide an instantiated by the call instance of the resolver where the API must use "reflection" of sorts to new one up so context is disconnected with the application.

gaussianrecurrence:
> I am not sure I understand the motivation. Are you saying that you want to define the region in `cache.xml` and then later mutate it from the API? I am not sure how safe this would be given the general immutable nature of the region objects. What is the motivation to split the configuration this way. Our general recommendation is to use the API over the XML configuration anyway. The server is moving away from the XML configuration to cluster config and API config. The API provides the ability to provide an instantiated by the call instance of the resolver where the API must use "reflection" of sorts to new one up so context is disconnected with the application.

Use case here is that we would like to have sort of a registry of partition resolvers, which we define in our application and in the cache XML file we specify which one of them we use for each partitioned region. That would be the ideal way of doing it. This way we only need to register all the partition resolvers we want to use.
Also using a shared library is a no go as the application is basically a microservice, hence the less artifacts we have the better.

I was not aware of the fact that the client was also moving away from declarative cache definition. If that's the case, there is no argument I can give you that justifies allowing to mutate the partition resolver.

So being that the case, unless you think there is a possibility to see this implemented and accepted by the community, I guess I am closing the PR?

Thanks for putting some time on this!

pivotal-jbarrett:
> Use case here is that we would like to have sort of a registry of partition resolvers, which we define in our application and in the cache XML file we specify which one of them we use for each partitioned region. That would be the ideal way of doing it. This way we only need to register all the partition resolvers we want to use.
> Also using a shared library is a no go as the application is basically a microservice, hence the less artifacts we have the better.

I get wanting to declare them in the XML so again I don't get why need this PR to make the region resolver mutable. If the resolver is defined in the XML then it should work as designed.

I think what your real issue is that if you have the classes defined in your application how do you declare them in the XML if the XML config expects a shared library name and symbol name for factory. That is something we could easily fix. How about allowing the library name to be optional. If there is no library name then the symbol lookup would happen in the current application space. I think that is possible.

gaussianrecurrence:
> > Use case here is that we would like to have sort of a registry of partition resolvers, which we define in our application and in the cache XML file we specify which one of them we use for each partitioned region. That would be the ideal way of doing it. This way we only need to register all the partition resolvers we want to use.
> > Also using a shared library is a no go as the application is basically a microservice, hence the less artifacts we have the better.
> 
> I get wanting to declare them in the XML so again I don't get why need this PR to make the region resolver mutable. If the resolver is defined in the XML then it should work as designed.
> 
> I think what your real issue is that if you have the classes defined in your application how do you declare them in the XML if the XML config expects a shared library name and symbol name for factory. That is something we could easily fix. How about allowing the library name to be optional. If there is no library name then the symbol lookup would happen in the current application space. I think that is possible.

You are right

> > Use case here is that we would like to have sort of a registry of partition resolvers, which we define in our application and in the cache XML file we specify which one of them we use for each partitioned region. That would be the ideal way of doing it. This way we only need to register all the partition resolvers we want to use.
> > Also using a shared library is a no go as the application is basically a microservice, hence the less artifacts we have the better.
> 
> I get wanting to declare them in the XML so again I don't get why need this PR to make the region resolver mutable. If the resolver is defined in the XML then it should work as designed.
> 
> I think what your real issue is that if you have the classes defined in your application how do you declare them in the XML if the XML config expects a shared library name and symbol name for factory. That is something we could easily fix. How about allowing the library name to be optional. If there is no library name then the symbol lookup would happen in the current application space. I think that is possible.

Regarding mutating the partition resolver, we are on the same point, it's not ideal, and I will abandon this approach.
Regarding looking into the application space, that's a great idea, which I've already tested and it's working. So I will push a change with it.

pivotal-jbarrett:
Closing this PR based on the conversation above. Looking forward to a PR that fixes the factory lookup logic.

PR #693, opened 2020-11-16
GEODE-8442: Exception in server not identified correctly in client

mkevo:
HandleReplay is already commented in other functions as replay is unsupported, this change will not translate the real exception to UnknownException.
User will got correct exception and will know how to handle it.


echobravopapa (changes requested):
Do you have a test that led to this discovery?  also, if these lines are not needed they should be deleted rather than commented out...  
@mkevo did you intend this to be a `DRAFT` pr?

mkevo:
I'm trying to write a test for this, as I reproduce it manually. There are steps to reproduce the issue:

1. Start locator and three servers.
2. Run some puts and then run containsKeyOnServer function with transactions.
3. While containsKeyOnServer is executing kill lead server.
4. You will got UnknownException without this change.

So if you have some idea how to do this containsKeyOnServer and killing server lead in parallel, it will help a lot.
@echobravopapa , yes, I intend this to be a draft.

mkevo:
Just to add that this reproduction steps can lead to the few exceptions, depends on which state of the transaction it fail(It can be MessageException for  CacheTransactionManager::failover or UnknownException for TransactionDataNodeHasDepartedException).
Also the help is needed for writing a test as I need to do containsKeyOnServer and killing server lead in parallel, but don't have an idea how to do this in tests.

mkevo:
Hi @echobravopapa,
I wrote a test, but now have a problem with LGTM check. I saw on other PR that this also happened today. Do you maybe know how to fix it?

moleske:
> I wrote a test, but now have a problem with LGTM check. I saw on other PR that this also happened today. Do you maybe know how to fix it?

Looking a [the output of LGTM](https://lgtm.com/projects/g/apache/geode-native/logs/rev/pr-d830587f41dd9f18339edc0e8866570fd9e44ab6/lang:cpp/stage:Build%20merge_d3f6b1592a01fd440ae19b9cb4772495d3f37649), it seems like it failed to merge with develop.  When I've hit this in the past, a rebase usually solves the problem.

mkevo:
Hi @moleske ,
I tried with rebase, bit the same thing happened.

mkevo:
Hi @moleske, @echobravopapa ,
finally I fix this somehow in a new PR, so I will close this one.
Please take a look on #713

PR #694, opened 2020-11-16
GEODE-8713: .NET user guide - API reference links point to C++ API

davebarnes97:
Added a language-agnostic redirect to which the appropriate API reference can be assigned.
Applied the change to both .NET and C++ user guides, for consistency across-the-board.


karensmolermiller (approved):
Looks good to me.

PR #695, opened 2020-11-17
GEODE-8318: Threading issues when switching from ACE to boost::asio

albertogpz:
Hangs in integration-test cases have been fixed.


mreddington:
Should we abandon this now?

albertogpz:
> Should we abandon this now?

Yes, we should. The commit that was just merged contained all the fixes.

PR #696, opened 2020-11-18
GEODE-8724: Fix compilation using docker image

gaussianrecurrence:
- As latest tag for ubuntu base image was recently change from 18.04 to
   20.04 it seems that some libraries version changed, like for example
   libc++.
 - As it seems with this new version libc++ headers are either not
   anymore located within /usr/include/c++/v1 or the symlink is missing.
 - So in order to solve that a symlink is manually created to ensure
   clang is able to find libc++ headers.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -38,6 +38,7 @@ RUN apt-get update && \
             openjdk-8-jdk \
             wget && \
         rm -rf /var/lib/apt/lists/* && \
+        ln -s /usr/lib/llvm-*/include/c++/v1/ /usr/include/c++/v1 && \
         update-alternatives --install /usr/bin/clang         clang         /usr/bin/clang-${CLANG_VERSION} 999 && \
         update-alternatives --install /usr/bin/clang++       clang++       /usr/bin/clang++-${CLANG_VERSION} 999 && \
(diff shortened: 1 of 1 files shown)


gaussianrecurrence:
As I noticed, yesterday builds were failing, so I hope this helps you.
I am afraid this very build will fail, as until the image is re-uploaded travis will be using the old one.


Regarding versioning, I made a point in the Jira issue I'd recommend you to read :)

pdxcodemonkey (approved):
Looks good, will get this merged and monitor Travis

gaussianrecurrence:
Still I think @dickcav or someone else needs to re-upload the image to docker hub as currently both latest and 1.13.1 tags are failing to compile.

pivotal-jbarrett:
I pushed a new image last night.

onichols-pivotal:
@pivotal-jbarrett how was your image different from the corrected image that Release Manager @dickcav pushed?  Did you push to both `1.13.1` and `latest` tags?  On which branch(es) are the Dockerfile changes commited that you pushed?

dickcav:
Yesterday afternoon is when we did what @gaussianrecurrence requested and merged the change, rebuilt docket and pushed but maybe that wasn't apparent that happened?

pivotal-jbarrett:
When I looked at hub.docker the most recent image pushed were older than the commit. I see the new images are only a few hours old. All should be fine now.

PR #698, opened 2020-11-19
GEODE-8729: Change version of Geode to 1.13.1 in LGTM analysis

albertogpz:


onichols-pivotal (approved):
I think this exact change has already been made on develop

moleske:
@albertogpz - closing since it was [parametrized](https://github.com/apache/geode-native/blob/develop/.lgtm.yml#L5-L10) recently in order to be auto updated when Geode cuts a new release

PR #700, opened 2020-11-20
GEODE-8735: Change factory logic for symbols lookup

gaussianrecurrence:
- In order to allow that factory logic looks into the application space
   for symbols, `library-name` field has been made optional.
 - This removes the need to create shared library for in order to define
   custom factories like cache loaders, cache writers, cache listeners,
   partition resolvers or persistance managers whenever the cache is
   instantiated declaratively.
 - Whenever library-name is not specified a handle to the executable is
   obtained.
 - Take into account that for OS like Linux, Mac... in addition to exporting
   the symbols in the application, you would need to specify `-rdynamic`
   option while compiling.
 - Documentation has been updated.

---

**NOTE**. Take a look into the Jira issue for additional context.


cppcache/src/Utils.cpp
```diff
@@ -173,20 +174,22 @@ const boost::dll::shared_library& getSharedLibrary(
 
   const auto& find = sharedLibraries.find(libraryName);
   if (find == sharedLibraries.end()) {
+    auto path = libraryName.empty() ? boost::dll::program_location()
```
pivotal-jbarrett, line 177: Great timing for me merging the boost::dll changes the other day. This change looks pretty straight forward!

cppcache/src/Utils.cpp
```diff
@@ -173,20 +174,22 @@ const boost::dll::shared_library& getSharedLibrary(
...
+                          path,
                           boost::dll::load_mode::rtld_global |
                               boost::dll::load_mode::rtld_lazy |
                               boost::dll::load_mode::append_decorations |
                               boost::dll::load_mode::search_system_folders))
                   .first->second;
     } catch (const boost::dll::fs::system_error& e) {
-      throw IllegalArgumentException("cannot open library: " + libraryName +
-                                     ": reason: " + e.what());
+      throw IllegalArgumentException("cannot open library: \"" + libraryName +
```
pivotal-jbarrett, line 191: I wonder if path makes more sense here now? Can the `shared_library` construction fail ever in the case of `program_location`?

pivotal-jbarrett (changes requested):
This looks great but needs an integration tests. Can you create one using the new framework that utilizes one of the plugin points from the test executable itself (blank attribute name). You should be able to do this programmatically since api still exposes the library/factory method that the XML configuration uses.

pivotal-jbarrett:
> * Take into account that for OS like Linux, Mac... in addition to exporting
>   the symbols in the application, you would need to specify `-rdynamic`
>   option while compiling.

This might make great example with corresponding CMake configuration.

gaussianrecurrence:
There's one thing I am not sure about if it's the right call and that's having added a modification of GenerateExportHeaders. This is necessary in order to specify the visibility flags to the factory function added to cpp-integration-test for testing purposes. Thing is GenerateExportHeaders does not support executables, so that's the rationale behind ExecutableExportHeader.cmake
Please let me know what you think about it and if you think there might be a better alternative.

gaussianrecurrence:
> > * Take into account that for OS like Linux, Mac... in addition to exporting
> >   the symbols in the application, you would need to specify `-rdynamic`
> >   option while compiling.
> 
> This might make great example with corresponding CMake configuration.

I've found out that CMake support exposing executable symbols by setting ENABLE_EXPORTS=1 property to the executable target. I've used it on IT, so it could serve as an example. If you think it should be stated more explicitly somewhere else, please let me know :)

pivotal-jbarrett (approved):

gaussianrecurrence:
Maybe @pdxcodemonkey @echobravopapa @moleske @mivanac @mkevo could you also review this, so it can be merged? Thanks!

cppcache/integration/test/resources/pr_app_client_cache.xml
```diff
@@ -0,0 +1,29 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<client-cache xmlns="http://geode.apache.org/schema/cache"
+			  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+			  xsi:schemaLocation="http://geode.apache.org/schema/cache http://geode.apache.org/schema/cache/cache-1.0.xsd"
```
pdxcodemonkey, line 21: @pivotal-jbarrett shouldn't the schemaLocation for both of these be "http://geode.apache.org/schema/cpp-cache/cpp-cache-1.0.xsd"?

pdxcodemonkey (changes requested):
Looks good to me, just need an answer on the XML schema.  Thanks!

mkevo (approved):

cppcache/integration/test/resources/pr_app_client_cache.xml
pivotal-jbarrett: No, it's lost of pairs of namespace uri to url mapping.

gaussianrecurrence: Sadly I am not really aware with the XML internals, but just to be clear, should I change the namespace to: https://geode.apache.org/schema/cpp-cache and the schema to https://geode.apache.org/schema/cpp-cache/cpp-cache-1.0.xsd ?

pivotal-jbarrett: Yes, but I think the parser is still non-validating so it won't really matter. I suspect many of our client xml files are bad. The Java server and client will validated and fail of the namespace and location are bad.

moleske (approved):

pdxcodemonkey (approved):

PR #701, opened 2020-11-23
GEODE-8792: Adds support for MacOS Universal binaries.

pivotal-jbarrett:
Specify CMAKE_OSX_ARCHITECTURES with multiple target architectures.

* Updates ACE to 6.5.12
* Updates Boost to 1.75.0
* Updates Benchmark to 1.5.2


dependencies/ACE/CMakeLists.txt
```diff
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-project( ACE VERSION 6.5.9 LANGUAGES NONE )
+project( ACE VERSION 6.5.12 LANGUAGES NONE )
```
pdxcodemonkey, line 16: Ugh - so disappointing that we have to take another update to this one before we can get rid of it altogether.

pdxcodemonkey (approved):
Looks good to me - nobody has an ARM Mac yet anyway, and you've already built it for all our other platforms.  Thanks for the proactive change!  It'll be good not to have this to fuss over later.

dependencies/ACE/CMakeLists.txt
moleske: Well, it might not still work on macOS arm chips because BigSur support was added after 6.5.12 was cut https://github.com/DOCGroup/ACE_TAO/commit/61338c70ac750777c0b8e34f2573c4800e7de30b

PR #704, opened 2020-12-02
GEODE-8754: Fix uninitialized variable in DataInput::ReadInternalObject

mmartell:
This fixes an uninitialized variable in the clicache code (DataInput::ReadInternalObject). Also, added a new serialization test (CompositeClassWithClassAsKey) since there were no existing tests that executed this function.


tests/javaobject/cli/TestClassC.java
```diff
@@ -0,0 +1,114 @@
...
+     Instantiator.register(new Instantiator(javaobject.cli.TestClassC.class, (byte) 102) {
+     public DataSerializable newInstance() {
+        return new TestClassC();
+     }
+   });
+  }
+
+
+
+/*
```
echobravopapa, line 39: dead code, pls remove

clicache/src/DataInput.cpp
```diff
@@ -657,7 +657,7 @@ namespace Apache
           //Log::Debug("DataInput::ReadInternalObject m_cursor " + m_cursor);
           bool findinternal = false;
           int8_t typeId = ReadByte();
-          System::Int64 compId;
+          System::Int64 compId = 0;
```
echobravopapa, line 660: RAII FTW?!?

clicache/integration-test2/SerializationTests.cs
```diff
@@ -510,5 +823,119 @@ public void PutGetCustomSerializableTypes()
                 cache.Close();
             }
         }
-    }
-}
+
+		[Fact]
+		public void CompositeClassWithClassAsKey()
```
echobravopapa, line 828: Good addition.   Does a non-composite class used as a key exercise the same code paths?  if not, be great to add a test for that too...

echobravopapa (changes requested):

clicache/integration-test2/SerializationTests.cs
```diff
@@ -25,7 +26,335 @@
...
+
+		//private unsafe void fillstack(Int64[] val)
+		//{
+  //          return;
+		//}
+		public void FromData(DataInput input)
+		{
+			testclassA.FromData(input);
+
+			// Fill the stack with Order typeid: 66
```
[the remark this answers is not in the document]
mmartell: Fixed.

tests/javaobject/cli/TestClassC.java
mmartell: Fixed dead code in TestClassB.java and TestClassA.java.

clicache/integration-test2/SerializationTests.cs
mmartell: I agree we should have a separate ClassAsKey. It would only use ReadInternalObject if the class being used as the key had a member that was a container (i.e., list, vector, dictionary). I had it that way originally.

clicache/src/DataInput.cpp
mmartell: Yes, it's a good idea to initialize things before you use them!

clicache/integration-test2/SerializationTests.cs
mmartell: Added a new test ClassAsKey.

tests/javaobject/cli/TestClassC.java
```diff
@@ -0,0 +1,94 @@
...
+    Integer year = new Integer(Year);
+
+    int hashcode =
+	  color.hashCode() ^
+	  make.hashCode() ^
+	  year.hashCode();
+    
+    return hashcode;
+  }
+}
```
pivotal-jbarrett, line 94: Missing a few terminating line feeds if you haven't noticed the (-) indicator here on your new files.

pivotal-jbarrett (approved):

tests/javaobject/cli/TestClassC.java
mmartell: Good catch. I'm now aware of the (-) indicator. Final line feed added on the four new files.

echobravopapa (approved):

PR #706, opened 2020-12-08
GEODE-8773: Test that PING and CLOSE_CONNECTION are not implicitly transactional.

mreddington:


pdxcodemonkey:
@dihardman @karensmolermiller @davebarnes97

cppcache/integration/framework/Cluster.cpp
```diff
@@ -110,6 +109,12 @@ void Locator::start() {
                      .withPreferIPv6(cluster_.getUseIPv6())
                      .withJmxManagerStart(true);
 
+  if(cluster_.getLogLevel().empty()) {
```
pivotal-jbarrett, line 112: Why pull this out of the `.withLogLevel` model? Why not have cluster default non-empty value and then always pass cluster value through to `gfsh.withLogLevel()`?

cppcache/integration/test/TransactionsTest.cpp
```diff
@@ -36,6 +40,30 @@ using apache::geode::client::Pool;
 using apache::geode::client::Region;
 using apache::geode::client::RegionShortcut;
 
+std::string getServerLogName() {
```
pivotal-jbarrett, line 43: Using logs to validate tests is really dangerous. Log messages change and aren't part of API. Is there a more concrete way to test this? I would think a unit tests that asserts these two methods set the correct 'no tx' value of `-1` should be sufficient.

cppcache/src/TcrMessage.cpp
```diff
@@ -2096,14 +2096,9 @@ TcrMessagePing::TcrMessagePing(DataOutput* dataOutput, bool decodeAll) {
   m_request->writeInt(static_cast<int32_t>(
       0));  // 17 is fixed message len ...  PING only has a header.
   m_request->writeInt(static_cast<int32_t>(0));  // Number of parts.
-  // int32_t txId = TcrMessage::m_transactionId++;
-  // Setting the txId to 0 for all ping message as it is not being used on the
-  // SERVER side or the
-  // client side.
-  m_request->writeInt(static_cast<int32_t>(0));
+  m_request->writeInt(m_txId);
```
pivotal-jbarrett, line 2099: A ping is not a transactional item. It should send `-1`.

cppcache/src/TcrMessage.cpp
```diff
@@ -2114,8 +2109,7 @@ TcrMessageCloseConnection::TcrMessageCloseConnection(DataOutput* dataOutput,
   m_request->writeInt(m_msgType);
   m_request->writeInt(static_cast<int32_t>(6));
   m_request->writeInt(static_cast<int32_t>(1));  // Number of parts.
-  // int32_t txId = TcrMessage::m_transactionId++;
-  m_request->writeInt(static_cast<int32_t>(0));
+  m_request->writeInt(m_txId);
```
pivotal-jbarrett, line 2112: Not transactional, should send `-1`.

pivotal-jbarrett (changes requested):

pivotal-jbarrett (changes requested):
I take all my previous comments back on the grounds this change should not be made at all. To remain consistent, however odd, with the Java client these messages should continue to set transaction ID to `0`. The fix should be made in the Geode server sources at make BaseCommand.shouldMasqueradeForTx consistent with all other checks against `NOTX`. In all other checks explicit equality is checked, not for greater or less than.

Change
```java
  protected boolean shouldMasqueradeForTx(Message clientMessage,
      ServerConnection serverConnection) {
    return serverConnection.getClientVersion().isNotOlderThan(KnownVersion.GFE_66)
        && clientMessage.getTransactionId() > TXManagerImpl.NOTX;
  }
```
to
```java
  protected boolean shouldMasqueradeForTx(Message clientMessage,
      ServerConnection serverConnection) {
    return serverConnection.getClientVersion().isNotOlderThan(KnownVersion.GFE_66)
        && clientMessage.getTransactionId() != TXManagerImpl.NOTX;
  }
```

pivotal-jbarrett:
> The fix should be made in the Geode server sources at make BaseCommand.shouldMasqueradeForTx consistent with all other checks against `NOTX`. In all other checks explicit equality is checked, not for greater or less than.

Pointed out offline to me, this won't fix the original issue either since `0 != NOTX` is `true` too. 🤦 

I really think we should understand why `0` was used here and on both the C++ and Java versions of these messages.

cppcache/integration/framework/Cluster.cpp
pdxcodemonkey: Switched to a simple unit test, framework changes are gone.

cppcache/integration/test/TransactionsTest.cpp
pdxcodemonkey: Done.

pdxcodemonkey:
Per the larger team, Java side is not using 0 for these, only for certain server-initiated messages.  Java client sets NOTX for PING and CLOSE_CONNECTION, native client should as well.

cppcache/src/TcrMessage.cpp
```diff
@@ -2090,17 +2090,16 @@ TcrMessageReply::TcrMessageReply(bool decodeAll,
 
 TcrMessagePing::TcrMessagePing(DataOutput* dataOutput, bool decodeAll) {
   m_msgType = TcrMessage::PING;
+  // -1 is NOTX constant server-side, PING should *never* be part of a
+  // transaction
+  m_txId = -1;
```
pivotal-jbarrett, line 2095: Isn't this the default? Look at `TcrMessage` default constructor.

cppcache/src/TcrMessage.cpp
```diff
@@ -2090,17 +2090,16 @@ TcrMessageReply::TcrMessageReply(bool decodeAll,
...
   m_request->writeInt(m_msgType);
   m_request->writeInt(static_cast<int32_t>(
       0));  // 17 is fixed message len ...  PING only has a header.
   m_request->writeInt(static_cast<int32_t>(0));  // Number of parts.
-  // int32_t txId = TcrMessage::m_transactionId++;
-  // Setting the txId to 0 for all ping message as it is not being used on the
-  // SERVER side or the
-  // client side.
-  m_request->writeInt(static_cast<int32_t>(0));
+  m_request->writeInt(m_txId);
```
pivotal-jbarrett, line 2102: Look at `TcrMessage::writeHeader` and see that almost all the messages are using this method to do some of the message bits in here. Perhaps it makes sense to normalize this message to match.

cppcache/src/TcrMessage.cpp
```diff
@@ -2109,13 +2108,16 @@ TcrMessagePing::TcrMessagePing(DataOutput* dataOutput, bool decodeAll) {
...
+  // of a transaction
+  m_txId = -1;
   m_decodeAll = decodeAll;
   m_request.reset(dataOutput);
   m_request->writeInt(m_msgType);
   m_request->writeInt(static_cast<int32_t>(6));
   m_request->writeInt(static_cast<int32_t>(1));  // Number of parts.
   // int32_t txId = TcrMessage::m_transactionId++;
-  m_request->writeInt(static_cast<int32_t>(0));
+  m_request->writeInt(m_txId);
```
pivotal-jbarrett, line 2120: See comment about this in ping message.

pivotal-jbarrett (changes requested):

pivotal-jbarrett:
Should we also have a `NOTX` constant like the Java client?

pivotal-jbarrett (approved):

pivotal-jbarrett (approved):

PR #707, opened 2020-12-09
GEODE-8775: Add gfsh execute function support to new C++ test framework

mmartell:
This functionality is currently in the new .NET test framework but missing in the new C++ test framework. This was discovered during the course of adding a new ClassAsKey test for C++ client.


mmartell:
Whoops. Mixing two sets of changes. Closing.

PR #708, opened 2020-12-09
GEODE-8775: Add gfsh execute function support

mmartell:
Adds support to the new C++ test framework for executing the gfsh command: execute function

Note: The new .NET test framework already has this functionality.

### DIFF
--- a/cppcache/integration/framework/Gfsh.cpp
+++ b/cppcache/integration/framework/Gfsh.cpp
@@ -29,6 +29,8 @@ Gfsh::Shutdown Gfsh::shutdown() { return Shutdown{*this}; }
 
 Gfsh::Deploy Gfsh::deploy() { return Deploy(*this); }
 
+Gfsh::ExecuteFunction Gfsh::executeFunction() { return ExecuteFunction(*this); }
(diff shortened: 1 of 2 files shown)


moleske (commented):
I'm a tad confused.  You add some lovely new features to the newer integration tests so you can execute some functions on the server.  But there's no test added to take advantage of these swanky new capabilities.  You mention it was added on the C# side, which I assume needed this ability for some tests.  Why weren't those tests add to the C++ side?  If those tests aren't valuable to add on the C++ side, then I guess just keeping parity with the C# side is good enough...

mmartell:
Good questions moleske. Yes, I have a new ClassAsKey test (which depends on this new gfsh functionality) for C++ ready to submit as soon as this PR gets merged. I chose to keep this PR separate since it really is just a test framework enhancement.

moleske (approved):
Approving assuming @mmartell keeps his promise to open another PR that uses these fancy new abilities.  If he doesn't, guess I'll just go be sad somewhere...

I didn't test the new features out, will wait for that other PR to see them in action.

PR #709, opened 2020-12-09
GEODE-8713: user guide API reference links - fix redirects for Geode website

davebarnes97:
The previous fix failed when publishing to the Geode website (vs cf push to a PWS server).
Modified the website's `.htaccess` file to handle redirects there.
Changed `cppdocs` and `dotnetdocs` placeholders to `cppapiref` and `dotnetapiref`, respectively, to avoid recursion problems.
Updated source files (about* and client-cache-ref*) to use the new placeholders.
Tested locally and pushed to the Geode site to verify that it works in both environments.


karensmolermiller (approved):
LGTM

PR #714, opened 2020-12-16
GEODE-8562: Adds new C++ test for using a class as a key

mmartell:
This adds a new C++ serialization test for using a class as a key.

Notes:
* This test leverages the new gfsh execute function support in the C++ test framework.
* A .NET serialization test for using a class as a key has been added in a previous PR.

### DIFF
--- a/cppcache/integration/test/Position.cpp
+++ b/cppcache/integration/test/Position.cpp
@@ -29,31 +29,14 @@ int32_t Position::cnt = 0;
 
 Position::Position() { init(); }
 
-Position::Position(const char* id, int32_t out) {
+Position::Position(std::string id, int32_t outstandingShares) {
   init();
-  secId = CacheableString::create(id);
-  qty = out - (cnt % 2 == 0 ? 1000 : 100);
+  secId = id;
+  qty = outstandingShares - (cnt % 2 == 0 ? 1000 : 100);
   mktValue = qty * 1.2345998;
-  sharesOutstanding = out;
+  sharesOutstanding = outstandingShares;
   secType = L"a";
   pid = cnt++;
-}
-
-// This constructor is just for some internal data validation test
-Position::Position(int32_t iForExactVal) {
-  init();
-  char* id = new char[iForExactVal + 2];
-  for (int i = 0; i <= iForExactVal; i++) {
-    id[i] = 'a';
-  }
-  id[iForExactVal + 1] = '\0';
-  secId = CacheableString::create(id);
-  delete[] id;
-  qty = (iForExactVal % 2 == 0 ? 1000 : 100);
-  mktValue = qty * 2;
-  sharesOutstanding = iForExactVal;
-  secType = L"a";
-  pid = iForExactVal;
 }
 
 void Position::init() {
@@ -61,7 +44,7 @@ void Position::init() {
   bondRating = nullptr;
   convRatio = 0.0;
   country = nullptr;
-  delta = 0.0;
+  valueGain = 0.0;
   industry = 0;
   issuer = 0;
   mktValue = 0.0;
@@ -77,48 +60,40 @@ void Position::init() {
 
 void Position::toData(apache::geode::client::DataOutput& output) const {
   output.writeInt(avg20DaysVol);
-  output.writeObject(bondRating);
+  output.writeString(bondRating);
   output.writeDouble(convRatio);
-  output.writeObject(country);
-  output.writeDouble(delta);
+  output.writeString(country);
+  output.writeDouble(valueGain);
   output.writeInt(industry);
   output.writeInt(issuer);
   output.writeDouble(mktValue);
   output.writeDouble(qty);
-  output.writeObject(secId);
-  output.writeObject(secLinks);
+  output.writeString(secId);
+  output.writeString(secLinks);
   output.writeUTF(secType);
   output.writeInt(sharesOutstanding);
-  output.writeObject(underlyer);
+  output.writeString(underlyer);
   output.writeInt(volatility);
   output.writeInt(pid);
 }
 
 void Position::fromData(apache::geode::client::DataInput& input) {
   avg20DaysVol = input.readInt64();
-  bondRating = std::dynamic_pointer_cast<CacheableString>(input.readObject());
+  bondRating = std::string(input.readString());
   convRatio = input.readDouble();
-  country = std::dynamic_pointer_cast<CacheableString>(input.readObject());
-  delta = input.readDouble();
+  country = std::string(input.readString());
+  valueGain = input.readDouble();
   industry = input.readInt64();
   issuer = input.readInt64();
   mktValue = input.readDouble();
   qty = input.readDouble();
-  secId = std::dynamic_pointer_cast<CacheableString>(input.readObject());
-  secLinks = std::dynamic_pointer_cast<CacheableString>(input.readObject());
+  secId = std::string(input.readString());
+  secLinks = std::string(input.readString());
   secType = input.readUTF<wchar_t>();
   sharesOutstanding = input.readInt32();
-  underlyer = std::dynamic_pointer_cast<CacheableString>(input.readObject());
+  underlyer = std::string(input.readString());
   volatility = input.readInt64();
   pid = input.readInt32();
-}
-std::string Position::toString() const {
-  char buf[2048];
-  sprintf(buf,
-          "Position Object:[ secId=%s type=%ls sharesOutstanding=%d id=%d ]",
-          secId->toString().c_str(), this->secType.c_str(),
-          this->sharesOutstanding, this->pid);
-  return buf;
 }
 
 }  // namespace DataSerializableTest
--- a/cppcache/integration/test/Position.cpp
+++ b/cppcache/integration/test/Position.cpp
@@ -41,19 +41,19 @@ Position::Position(std::string id, int32_t outstandingShares) {
 
 void Position::init() {
   avg20DaysVol = 0;
-  bondRating = nullptr;
+  bondRating = "";
   convRatio = 0.0;
-  country = nullptr;
+  country = "";
   valueGain = 0.0;
   industry = 0;
   issuer = 0;
   mktValue = 0.0;
   qty = 0.0;
-  secId = nullptr;
-  secLinks = nullptr;
+  secId = "";
+  secLinks = "";
   secType = L"";
   sharesOutstanding = 0;
-  underlyer = nullptr;
+  underlyer = "";
   volatility = 0;
   pid = 0;
 }
--- a/cppcache/integration/test/Position.cpp
+++ b/cppcache/integration/test/Position.cpp
@@ -27,35 +27,26 @@ namespace DataSerializableTest {
 
 int32_t Position::cnt = 0;
 
-Position::Position() { init(); }
+Position::Position() {
+  avg20DaysVol = 0;
+  convRatio = 0.0;
+  valueGain = 0.0;
+  industry = 0;
+  issuer = 0;
+  mktValue = 0.0;
+  qty = 0.0;
+  sharesOutstanding = 0;
+  volatility = 0;
+  pid = 0;
+}
 
-Position::Position(std::string id, int32_t outstandingShares) {
-  init();
+Position::Position(std::string id, int32_t outstandingShares) : Position() {
   secId = id;
   qty = outstandingShares - (cnt % 2 == 0 ? 1000 : 100);
   mktValue = qty * 1.2345998;
   sharesOutstanding = outstandingShares;
   secType = L"a";
   pid = cnt++;
-}
-
-void Position::init() {
-  avg20DaysVol = 0;
-  bondRating = "";
-  convRatio = 0.0;
-  country = "";
-  valueGain = 0.0;
-  industry = 0;
-  issuer = 0;
-  mktValue = 0.0;
-  qty = 0.0;
-  secId = "";
-  secLinks = "";
-  secType = L"";
-  sharesOutstanding = 0;
-  underlyer = "";
-  volatility = 0;
-  pid = 0;
 }
 
 void Position::toData(apache::geode::client::DataOutput& output) const {
--- a/cppcache/integration/test/Position.cpp
+++ b/cppcache/integration/test/Position.cpp
@@ -1,0 +1,124 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "Position.hpp"
+
+#include <wchar.h>
+
+#include <cwchar>
+
+#include <geode/DataInput.hpp>
+#include <geode/DataOutput.hpp>
+
+namespace DataSerializableTest {
+
+int32_t Position::cnt = 0;
+
+Position::Position() { init(); }
+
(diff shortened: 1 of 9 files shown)


cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include <geode/DataInput.hpp>
+#include <geode/DataOutput.hpp>
+
+namespace DataSerializableTest {
+
+int32_t Position::cnt = 0;
+
+Position::Position() { init(); }
+
+Position::Position(const char* id, int32_t out) {
```
pivotal-jbarrett, line 32: This source smells like it was lifted from somewhere without care to bring it up to our high standards. Can you please:
* Use `std::string` rather than `char *`.
* Readable and descriptive variable names. (What is `out`?)
* Inner class members should not be `CacheableX` types.
* Don't use C methods where C++ methods exist, like `sprintf`.

cppcache/integration/test/Position.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+using apache::geode::client::CacheableString;
+using apache::geode::client::DataInput;
+using apache::geode::client::DataOutput;
+using apache::geode::client::DataSerializable;
+
+class Position : public DataSerializable {
+ private:
+  int64_t avg20DaysVol;
+  std::shared_ptr<CacheableString> bondRating;
```
pivotal-jbarrett, line 42: Use `std::string`.

cppcache/integration/test/Position.hpp
```diff
@@ -0,0 +1,100 @@
...
+  double convRatio;
+  std::shared_ptr<CacheableString> country;
+  double delta;
+  int64_t industry;
+  int64_t issuer;
+  double mktValue;
+  double qty;
+  std::shared_ptr<CacheableString> secId;
+  std::shared_ptr<CacheableString> secLinks;
+  // wchar_t* secType;
```
pivotal-jbarrett, line 52: Delete commented code.

cppcache/integration/test/Position.hpp
```diff
@@ -0,0 +1,100 @@
...
+class Position : public DataSerializable {
+ private:
+  int64_t avg20DaysVol;
+  std::shared_ptr<CacheableString> bondRating;
+  double convRatio;
+  std::shared_ptr<CacheableString> country;
+  double delta;
+  int64_t industry;
+  int64_t issuer;
+  double mktValue;
```
pivotal-jbarrett, line 48: Readable fully qualified variable names please.

cppcache/integration/test/Position.hpp
```diff
@@ -0,0 +1,100 @@
...
+  Position();
+  Position(const char* id, int32_t out);
+  // This constructor is just for some internal data validation test
+  explicit Position(int32_t iForExactVal);
+  ~Position() override = default;
+  void toData(DataOutput& output) const override;
+  void fromData(DataInput& input) override;
+  std::string toString() const override;
+
+  size_t objectSize() const override {
```
pivotal-jbarrett, line 75: Unless this or any other test is using usage based eviction we should eliminate the noise of `objectSize`.

cppcache/integration/test/PositionKey.cpp
```diff
@@ -0,0 +1,46 @@
...
+  m_positionId = input.readInt64();
+}
+
+bool PositionKey::operator==(const CacheableKey& other) const {
+  return m_positionId == ((PositionKey&)other).getPositionId();
+}
+
+int PositionKey::hashcode() const {
+  int hash = 11;
+  hash = 31 * hash + (int)m_positionId;
```
pivotal-jbarrett, line 42: This hash function does not match the Java version of this type. This should be fixed or partitioning will be off. Also very odd both just down cast `positionId`. Regardless you should use `static_cast<int32_t>()` not C-style casts.

tests/javaobject/cli/InstantiateDataSerializable.java
```diff
@@ -32,15 +32,16 @@
...
-  Instantiator.register(new Instantiator(javaobject.cli.Position.class, 22) {
+
+  Instantiator.register(new Instantiator(javaobject.cli.PositionKey.class, 21) {
     public DataSerializable newInstance() {
-      return new javaobject.cli.Position();
+      return new javaobject.cli.PositionKey();
     }
   });
 
-  Instantiator.register(new Instantiator(javaobject.cli.PositionKey.class, 77) {
```
pivotal-jbarrett, line 41: Why change the ID from 77 to 21? Have all the older tests expecting 77 been updated?

tests/javaobject/cli/PositionKey.java
```diff
@@ -26,7 +26,7 @@
   private long positionId;
 
   static {
-     Instantiator.register(new Instantiator(javaobject.cli.PositionKey.class, (byte) 77) {
+     Instantiator.register(new Instantiator(javaobject.cli.PositionKey.class, (byte) 21) {
```
pivotal-jbarrett, line 29: I don't know why it casts this literal to a `byte`, the method takes an `int`. Remove the cast.

cppcache/integration/test/PositionKey.hpp
```diff
@@ -0,0 +1,58 @@
...
+ private:
+  int64_t m_positionId;
+
+ public:
+  PositionKey() {}
+  PositionKey(int64_t positionId) { m_positionId = positionId; }
+  ~PositionKey() override = default;
+
+  bool operator==(const CacheableKey& other) const;
+  int32_t hashcode() const;
```
pivotal-jbarrett, line 45: Missing `override`.

pivotal-jbarrett (changes requested):

cppcache/integration/test/Position.cpp
mmartell: You're right. This was just copied from the testobject folder, which is used by the old legacy framework tests. I agree to include it in the new framework tests it needs to be sanitized.

tests/javaobject/cli/InstantiateDataSerializable.java
mmartell: Yes. All tests updated and pass. I had to use something other than 77, as it was causing a conflict (don't remember where). Chose 21 for PositionKey since it was unused and close to Position (22).

cppcache/integration/test/Position.cpp
```diff
@@ -41,19 +41,19 @@ Position::Position(std::string id, int32_t outstandingShares) {
 
 void Position::init() {
```
pivotal-jbarrett, line 42: Good C++ practice would sanitize this to an list of initializations on a common constructor. `std::string` initializes to `""` already.

pivotal-jbarrett (changes requested):

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <cwchar>
+
+#include <geode/DataInput.hpp>
+#include <geode/DataOutput.hpp>
+
+namespace DataSerializableTest {
+
+int32_t Position::cnt = 0;
+
+Position::Position() { init(); }
```
pivotal-jbarrett, line 30: Replace with:
```C++
Position::Position() :
  avg20DaysVol(0),
  convRatio(0.0),
  valueGain(0.0),
  industry(0),
  issuer(0),
  mktValue(0.0),
  qty(0.0),
  sharesOutstanding(0),
  volatility(0),
  pid(0) {}
```

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+Position::Position(std::string id, int32_t outstandingShares) {
```
pivotal-jbarrett, line 32: Replace with:
```C++
Position::Position(std::string id, int32_t outstandingShares) :
  Position(),
  secId(std::move(id)),
  secType(L"a"),
  sharesOutstanding(outstandingShares) {
  qty = outstandingShares - (cnt % 2 == 0 ? 1000 : 100);
  mktValue = qty * 1.2345998;
  pid = cnt++;
}
```

cppcache/integration/test/Position.hpp
```diff
@@ -0,0 +1,81 @@
...
+  double convRatio;
+  std::string country;
+  double valueGain;
+  int64_t industry;
+  int64_t issuer;
+  double mktValue;
+  double qty;
+  std::string secId;
+  std::string secLinks;
+  std::wstring secType;
```
pivotal-jbarrett, line 52: We shouldn't be continuing the use of the platform ambiguous `std::wstring`. Instead we should use `std::u16string` or just `std::string` with utf8 encoding. In either case the stream just needs to serialize the correct string type for the the Java side to deserialize.

pivotal-jbarrett (changes requested):

cppcache/integration/test/Position.hpp
mmartell: Done

cppcache/integration/test/Position.hpp
mmartell: Removed. Didn't realize that.

cppcache/integration/test/Position.hpp
mmartell: Done

cppcache/integration/test/PositionKey.cpp
mmartell: Now they match, and switched to static_cast.

tests/javaobject/cli/PositionKey.java
mmartell: Done.

cppcache/integration/test/PositionKey.hpp
mmartell: Done.

cppcache/integration/test/Position.cpp
mmartell: Good catch. Done.

cppcache/integration/test/Position.cpp
mmartell: Done.

cppcache/integration/test/Position.hpp
mmartell: Agreed. Just ratholed on this issue for 30 minutes. Not sure this test is the right place to be testing string types. I've changed this variable to just be std::string, and serializing with readString/writeString. The corresponding Position.java code uses String and readUTF/writeUTF for all of the string variables. Doesn't break any tests.

cppcache/integration/test/Position.cpp
mmartell: Switched to this style.

cppcache/integration/test/Position.hpp
pivotal-jbarrett: You should continue to use `writeUTF` even with `std::string` if the Java side uses `writeUTF`. This method has a very specific way of encoding strings that is non-standard. While it won't affect this test it could encourage us towards doing the wrong things in the future or worse encouraging a consumer to.

mmartell:
pivotal-jbarrett, thanks for the detailed review!

pivotal-jbarrett (changes requested):

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,89 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "Position.hpp"
+
+#include <wchar.h>
+
+#include <cwchar>
```
pivotal-jbarrett, line 21: Do we need `wchar.h` and `cwchar` anymore?

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,89 @@
...
+  output.writeDouble(convRatio);
+  output.writeString(country);
+  output.writeDouble(valueGain);
+  output.writeInt(industry);
+  output.writeInt(issuer);
+  output.writeDouble(mktValue);
+  output.writeDouble(qty);
+  output.writeString(secId);
+  output.writeString(secLinks);
+  output.writeString(secType);
```
pivotal-jbarrett, line 63: Needs to remain `writeUTF` to match Java side.

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,89 @@
...
+  bondRating = std::string(input.readString());
+  convRatio = input.readDouble();
+  country = std::string(input.readString());
+  valueGain = input.readDouble();
+  industry = input.readInt64();
+  issuer = input.readInt64();
+  mktValue = input.readDouble();
+  qty = input.readDouble();
+  secId = std::string(input.readString());
+  secLinks = std::string(input.readString());
```
pivotal-jbarrett, line 81: These reconstruction of `std::string` from `readString` seem redundant.

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,89 @@
...
+  secType = input.readString();
```
pivotal-jbarrett, line 82: Needs to remain `readUTF` to match Java side.

cppcache/integration/test/PositionKey.cpp
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "PositionKey.hpp"
+
+#include <wchar.h>
```
pivotal-jbarrett, line 19: Headers?

pivotal-jbarrett (changes requested):

cppcache/integration/test/Position.cpp
mmartell: Good catch.

cppcache/integration/test/Position.cpp
mmartell: Good catch.

cppcache/integration/test/PositionKey.cpp
mmartell: Good catch.

cppcache/integration/test/Position.cpp
mmartell: Good catch. You're kind of like an optimizing compiler.

cppcache/integration/test/PositionKey.hpp
```diff
@@ -0,0 +1,58 @@
...
+namespace DataSerializableTest {
+
+using apache::geode::client::CacheableKey;
+using apache::geode::client::DataInput;
+using apache::geode::client::DataOutput;
+using apache::geode::client::DataSerializable;
+
+class PositionKey : public DataSerializable, public CacheableKey {
+ private:
+  int64_t m_positionId;
```
pivotal-jbarrett, line 37: Since this is a new class, even if it is copied from some older sources, it should conform to our naming conventions.
```C++
int64_t positionId_;
```

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "Position.hpp"
+
+#include <wchar.h>
```
pivotal-jbarrett, line 19: Still need these wide char includes?

pivotal-jbarrett (changes requested):

cppcache/integration/test/Position.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+int32_t Position::cnt = 0;
+
+Position::Position()
+    : avg20DaysVol(0),
+      convRatio(0.0),
+      valueGain(0.0),
+      industry(0),
+      issuer(0),
+      mktValue(0.0),
```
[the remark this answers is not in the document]
mmartell: Good catch.

cppcache/integration/test/PositionKey.hpp
mmartell: Missed that.

cppcache/integration/test/Position.cpp
mmartell: Dang. Thought I removed those already. Good catch.

pivotal-jbarrett (approved):

PR #718, opened 2021-01-10
GEODE-8806: Refactor Logger code

pdxcodemonkey:
Log rolling wasn't working in all (any?) cases, due to poorly-written manual path manipulation code etc.

- Convert all ACE code and manual path/file logic to boost::filesystem
- Move logging integration test to unit tests, because it really is/was a unit test
- Add more unit tests for initialization and various log-rolling scenarios

@mreddington @dihardman @karensmolermiller @davebarnes97


cppcache/integration-test/fw_helper.hpp
```diff
@@ -131,10 +131,7 @@ class TestException {
         m_filename(const_cast<char*>(filename)) {}
 
   void print() {
-    char buf[256];
-    apache::geode::client::Log::formatLogLine(
-        buf, apache::geode::client::LogLevel::Error);
-    fprintf(stdout, "--->%sTestException: %s in %s at line %d<---\n", buf,
+    fprintf(stdout, "--->%sTestException: %s in %s at line %d<---\n", apache::geode::client::Log::formatLogLine(apache::geode::client::LogLevel::Error).c_str(),
```
pivotal-jbarrett, line 134: Why is the logic of this method inverted? Before the Log::formatLogLine formatted the message, now it only formats an empty string? Very confused.

cppcache/integration/test/AuthInitializeTest.cpp
```diff
@@ -68,7 +68,7 @@ const int32_t CQ_PLUS_AUTH_TEST_REGION_ENTRY_COUNT = 100000;
 
 Cache createCache(std::shared_ptr<SimpleAuthInitialize> auth) {
   auto cache = CacheFactory()
-                   .set("log-level", "debug")
+                   .set("log-level", "none")
```
pivotal-jbarrett, line 71: Since this test is explicitly setting a log file I suspect it might be slurping the logs for its test. If it is, bad but we shouldn't set to `none` here and break the test. If it isn't slurping the logs then remove the log `log-file` lines too.

cppcache/integration/test/CqPlusAuthInitializeTest.cpp
```diff
@@ -64,8 +64,7 @@ const int32_t CQ_PLUS_AUTH_TEST_REGION_ENTRY_COUNT = 50000;
 
 Cache createCache(std::shared_ptr<SimpleAuthInitialize> auth) {
   auto cache = CacheFactory()
-                   .set("log-level", "debug")
-                   .set("log-file", "geode_native.log")
+                   .set("log-level", "none")
```
pivotal-jbarrett, line 67: Thank you for this!!!!

cppcache/test/LoggingTest.cpp
```diff
@@ -0,0 +1,759 @@
+/*
```
pivotal-jbarrett, line 1: Yay for new test!

cppcache/test/LoggingTest.cpp
```diff
@@ -0,0 +1,759 @@
...
+  }
+
+  static void findRolledFiles(
+      const std::string& logFilePath,
+      std::map<int32_t, boost::filesystem::path>& rolledFiles) {
+    const auto basePath =
+        boost::filesystem::absolute(boost::filesystem::path(logFilePath)) /
+        testLogFileName;
+    const auto filterstring = basePath.stem().string() + "-(\\d+)\\.log$";
+    const boost::regex my_filter(filterstring);
```
pivotal-jbarrett, line 173: Why `boost::regex` and not `std::regex`?

cppcache/test/LoggingTest.cpp
```diff
@@ -0,0 +1,759 @@
...
+          auto index = std::atoi(
+              std::string(testMatches[1].first, testMatches[1].second).c_str());
+          rolledFiles[index] = i->path();
+        }
+      }
+    }
+  }
+
+  void testLogFnError() {
+#ifdef WIN32
```
pivotal-jbarrett, line 196: Is this because VC++ uses a different format for `__FUNCTION__`? It might be more appropriate to check for MSVC not WIN32 because if someone used Clang on Windows this might not hold true?

What does it look like if we switched to C++11 [`__func__`](https://en.cppreference.com/w/c/language/function_definition)?

pivotal-jbarrett (changes requested):
Looks awesome but have a few questions.

cppcache/integration-test/fw_helper.hpp
pdxcodemonkey: formatLogLine used to take a char* buffer and do icky snprintf stuff, now it _returns_ a formatted std::string.  The name is kind of misleading now, maybe it should be something like getLogLinePrefix?

cppcache/test/LoggingTest.cpp
pdxcodemonkey: Uhhh, because I didn't know that was a thing and I just assumed.  I'll switch to std::, that's much better, obvs.

cppcache/test/LoggingTest.cpp
pdxcodemonkey: One of the few times I actually prefer what Microsoft has done.  __FUNCTION__ in VS is "ClassName::FunctionName", other platforms it just gives you FunctionName, which is often insufficient.  Will look into `__func__` and switch the ifdef to MSVC.

cppcache/integration/test/CqPlusAuthInitializeTest.cpp
pdxcodemonkey: OMG I couldn't figure out why this stupid test was timing out in my pipeline, then noticed _a test that does 50k puts_ was logging at `debug` level.  Good grief.  So yeah, I set all the tests to `none`, except there's one that _is_ slurping logs that I had to leave at `debug` :(.

cppcache/integration/test/AuthInitializeTest.cpp
pdxcodemonkey: I don't think it is, will just clean this up the rest of the way.

cppcache/test/LoggingTest.cpp
pivotal-jbarrett: If the née C++ 11 symbol is consistent between platforms you won't need the ifdef.

cppcache/integration/test/AuthInitializeTest.cpp
pivotal-jbarrett: There was 1 or 2 others like it.

cppcache/test/LoggingTest.cpp
pivotal-jbarrett: On closer inspection, I don't think this LogFn stuff is even used. It looks like the only use is in the tests. I say delete it.

pivotal-jbarrett (changes requested):

cppcache/test/LoggingTest.cpp
pdxcodemonkey: Removed it.

cppcache/integration/test/AuthInitializeTest.cpp
pdxcodemonkey: Got em.

cppcache/test/LoggingTest.cpp
pdxcodemonkey: Done

cppcache/src/Log.cpp
```diff
@@ -502,178 +351,77 @@ LogLevel Log::charsToLevel(const std::string& chars) {
...
-  return buf;
-}
+  std::snprintf(timebuf, 15, ".%06" PRId64 " ",
+                static_cast<int64_t>(microseconds.count()));
+  msg << timebuf << " ";
 
-void Log::put(LogLevel level, const std::string& msg) {
-  put(level, msg.c_str());
+  std::strftime(timebuf, MINBUFSIZE, "%Z ", &tm_val);
+  msg << timebuf << " " << boost::asio::ip::host_name() << ":"
```
pivotal-jbarrett, line 373: I am a bit concerned this may result in a system call the resolver for every log message. The previous code cached this. I suggest a `::hostname()` method with a static resolved hostname using this method (Meyer's Singleton). The likely hood of the hostname changing over the life of a client, while not 0, is pretty darn close to 0.

cppcache/src/Log.cpp
```diff
@@ -502,178 +351,77 @@ LogLevel Log::charsToLevel(const std::string& chars) {
...
+  char timebuf[MINBUFSIZE];
 
-  std::snprintf(pbuf, 300, "%s:%d %" PRIu64 "] ", g_uname.nodename, g_pid,
-                hacks::aceThreadId(ACE_OS::thr_self()));
+  std::strftime(timebuf, MINBUFSIZE, "%Y/%m/%d %H:%M:%S", &tm_val);
+  msg << timebuf;
 
-  return buf;
-}
+  std::snprintf(timebuf, 15, ".%06" PRId64 " ",
```
pivotal-jbarrett, line 368: You should be able to do with with std stream formatting directives and avoid the extra calls to `snprintf`.

pivotal-jbarrett (changes requested):

cppcache/src/Log.cpp
pdxcodemonkey: Cached g_hostName string.  Didn't read the fine print here and implement an accessor, but I think we're okay as is.

cppcache/src/Log.cpp
pdxcodemonkey: Done.

cppcache/integration/test/CacheXmlTest.cpp
```diff
@@ -44,8 +44,7 @@ apache::geode::client::Cache createCacheUsingXmlConfig(
 
   CacheFactory cacheFactory;
 
-  auto cache = cacheFactory.set("log-level", "debug")
-                   .set("log-file", "geode_native.log")
+  auto cache = cacheFactory.set("log-level", "none")
                    .set("statistic-sampling-enabled", "false")
                    .set("cache-xml-file", xmlFile.c_str())
```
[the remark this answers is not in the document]
pdxcodemonkey: Fixed

clicache/src/Log.cpp
```diff
@@ -70,12 +70,6 @@ namespace Apache
           static_cast<apache::geode::client::LogLevel>(level));
       }
 
-      bool Log::Enabled(LogLevel level)
```
pivotal-jbarrett, line 73: Doesn't this change the public API of the .NET library?

cppcache/src/LocalRegion.cpp
```diff
@@ -717,7 +717,7 @@ void LocalRegion::registerEntryExpiryTask(
       new EntryExpiryHandler(rptr, entry, getEntryExpirationAction(), duration);
   auto id = rptr->getCacheImpl()->getExpiryTaskManager().scheduleExpiryTask(
       handler, duration, std::chrono::seconds::zero());
-  if (Log::finestEnabled()) {
+  if (Log::logLevel() >= LogLevel::Finest) {
```
pivotal-jbarrett, line 720: I am sort of indifferent on this change but the original was more closely aligned with the Java based logger API. This change requires you to know that log levels have order.

cppcache/src/statistics/GeodeStatisticsFactory.cpp
```diff
@@ -61,8 +61,9 @@ GeodeStatisticsFactory::~GeodeStatisticsFactory() {
     statsTypeMap.clear();
 
   } catch (const Exception& ex) {
-    Log::warningCatch("~GeodeStatisticsFactory swallowing Geode exception", ex);
-
+    LOGWARN(std::string(
```
pivotal-jbarrett, line 64: What is the rational for removing the a method that encapsulated the deconstruction of an exception being logged?

cppcache/src/util/Log.hpp
```diff
@@ -182,315 +183,69 @@ class APACHE_GEODE_EXPORT Log {
...
-#define LOGFINER                                  \
-  if (::apache::geode::client::LogLevel::Finer <= \
-      ::apache::geode::client::Log::logLevel())   \
-  ::apache::geode::client::LogVarargs::finer
-
-#define LOGFINEST                                  \
-  if (::apache::geode::client::LogLevel::Finest <= \
-      ::apache::geode::client::Log::logLevel())    \
-  ::apache::geode::client::LogVarargs::finest
+#define LOGDEBUG(...)                                                         \
```
pivotal-jbarrett, line 247: This change results in the loss of the LogLevel check to avoid the cost of executing functions that may make up the arguments to this log message. Now all arguments will be resolved and the log message executed regardless of the log message actually being loggable.

pivotal-jbarrett (changes requested):

clicache/src/Log.cpp
pdxcodemonkey: Maybe?  I'll have a look.  Easy enough to fix, if so.

cppcache/src/LocalRegion.cpp
pdxcodemonkey: Put `Log::enabled(LogLevel level)` back in, and am using that now.

clicache/src/Log.cpp
pdxcodemonkey: CLI Log class is actually `private`, but I think it makes sense to have an `Enabled()` method anyway, so it's back now.

cppcache/src/LocalRegion.cpp
pdxcodemonkey: Added enabled() back, and using that now.  No need for warningEnabled, finestEnabled, etc helper functions tho.

cppcache/src/util/Log.hpp
pdxcodemonkey: Right.  Added the level check to the new versions of the macros, and removed it from the `Log::log` methods.

pivotal-jbarrett (approved):

PR #719, opened 2021-01-13
GEODE-8831: Fix performance degradation after changing to boost::asio

albertogpz:
After replacing ACE_SOCK with boost::asio a
significant performance degradation has been observed
in send and receive operations.

The cause has been some log debug calls added to the
TcpConn::send and TcpConn::receive that took as
input a stringstream previously created.

After changing these logs to not use the
stringstream, the performance of the
native client has returned to the original
figures.

### DIFF
--- a/cppcache/src/TcpConn.cpp
+++ b/cppcache/src/TcpConn.cpp
@@ -163,40 +163,40 @@ TcpConn::TcpConn(const std::string ipaddr,
 }
 
 TcpConn::~TcpConn() {
-  std::stringstream ss;
-
   try {
-    ss << "Disconnected " << socket_.local_endpoint() << " -> "
-       << socket_.remote_endpoint();
-
+    LOGFINE("Disconnected %s:%u -> %s:%u",
+            socket_.local_endpoint().address().to_string().c_str(),
+            socket_.local_endpoint().port(),
+            socket_.remote_endpoint().address().to_string().c_str(),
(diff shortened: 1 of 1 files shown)


cppcache/src/TcpConn.cpp
```diff
@@ -163,40 +163,40 @@ TcpConn::TcpConn(const std::string ipaddr,
...
   return receive(buff, len, timeout, true);
 }
 
 size_t TcpConn::receive_nothrowiftimeout(char *buff, const size_t len,
                                          std::chrono::milliseconds timeout) {
-  std::stringstream ss;
-  ss << "Receiving an unknown number of bytes from "
-     << socket_.remote_endpoint() << " -> " << socket_.local_endpoint();
-  LOGDEBUG(ss.str());
+  LOGDEBUG("Receiving an unknown number of bytes from %s:%d -> %s:%d",
```
pivotal-jbarrett, line 195: In the future, if you want to use more complex logging you should wrap with `if (Log::enabled(LogLevel))` block. So the original would be:
```C++
if (Log::enabled(Debug)) {
  std::stringstream ss;
  ss << "Receiving " << len << " bytes from " << socket_.remote_endpoint()
    << " -> " << socket_.local_endpoint();
  Log::log(Debug, ss.str());
}
 ```

pivotal-jbarrett (approved):

PR #723, opened 2021-01-16
GEODE-8842: Limit no-used-but-marked-unused to new integration tests

moleske:
- SCOPED_TRACE in gtest causes this error, so only allow in that scope

Also sorted the tests listed because it bothered me that it was really close to sorted but wasn't

### DIFF
--- a/cppcache/integration/test/CMakeLists.txt
+++ b/cppcache/integration/test/CMakeLists.txt
@@ -16,6 +16,7 @@
 add_executable(cpp-integration-test
   AuthInitializeTest.cpp
   BasicIPv6Test.cpp
+  CacheWriterTest.cpp
   CacheXmlTest.cpp
   CleanIdleConnections.cpp
   ClientTransactionXATest.cpp
@@ -30,12 +31,13 @@ add_executable(cpp-integration-test
   FunctionExecutionTest.cpp
   LRUEvictionTest.cpp
   LocatorRequestsTest.cpp
+  Order.cpp
+  Order.hpp
   PartitionRegionOpsTest.cpp
+  PdxInstanceFactoryTest.cpp
   PdxInstanceTest.cpp
   PdxJsonTypeTest.cpp
   PdxSerializerTest.cpp
-  Order.cpp
-  Order.hpp
   Position.cpp
   Position.hpp
   PositionKey.cpp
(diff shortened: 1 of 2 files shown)


cppcache/integration/test/CMakeLists.txt
```diff
@@ -53,11 +55,15 @@ add_executable(cpp-integration-test
...
-  WanDeserializationTest.cpp
-  PdxInstanceFactoryTest.cpp
-  CacheWriterTest.cpp
   TransactionsTest.cpp
+  WanDeserializationTest.cpp
 )
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  target_compile_options(cpp-integration-test
+    PUBLIC
+      -Wno-used-but-marked-unused
```
pivotal-jbarrett, line 64: Is this only in a single test? Perhaps we can just add it as a `#pragma` to that file or eve suppress the single line. Worst case we could add this as an option to just that tests cpp file options.

pivotal-jbarrett (changes requested):

cppcache/integration/test/CMakeLists.txt
moleske: It is used twice in two test files, [CommitConflictExceptionTest Line 61](https://github.com/apache/geode-native/blob/6122245ddf0c55fd1e07ae1daae25d67db7889be/cppcache/integration/test/CommitConflictExceptionTest.cpp#L61), [CommitConflictExceptionTest Line 91](https://github.com/apache/geode-native/blob/6122245ddf0c55fd1e07ae1daae25d67db7889be/cppcache/integration/test/CommitConflictExceptionTest.cpp#L91), [ExampleTest Line 102](https://github.com/apache/geode-native/blob/6122245ddf0c55fd1e07ae1daae25d67db7889be/cppcache/integration/test/ExampleTest.cpp#L118), and [ExampleTest Line 102](https://github.com/apache/geode-native/blob/6122245ddf0c55fd1e07ae1daae25d67db7889be/cppcache/integration/test/ExampleTest.cpp#L118).

Since it was used four times across two files I went for cmake to suppress it for just tests.  I'm cool with a pragma if folks feel that's better, I don't have strong feelings.

pivotal-jbarrett (approved):

PR #724, opened 2021-01-18
GEODE-8845: Add missing public attribute to .NET Cacheables

mmartell:
Several of the provided .NET Cacheable types were missing the public attribute.

### DIFF
--- a/clicache/src/CacheableHashMap.hpp
+++ b/clicache/src/CacheableHashMap.hpp
@@ -44,7 +44,7 @@ namespace Apache
       /// that can serve as a distributable object for caching. This class
       /// extends .NET generic <c>Dictionary</c> class.
       /// </summary>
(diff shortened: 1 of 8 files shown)


pivotal-jbarrett (changes requested):
I am pretty certain this are meant to be internal only to bridge from .NET container types to C++ layer's wrappers for the Java container types. I don't see any reason these types should be leaked publicly. 

GEODE-8845 references the Cacheable* being public. They should not be public either. 

Can you explain in the ticket why these internal types should be public?

pivotal-jbarrett:
Use `Objects.Hash` from https://github.com/apache/geode-native/pull/740

mmartell:
Modified the ClassAsKey example to use new Objects.Hash() API. No longer need to expose Cacheables to get matching hash codes between client and server

PR #725, opened 2021-01-19
GEODE-8562: Add new C# classaskey Example 

mmartell:
This adds a new C# example that illustrates how to use a class as a key.


davebarnes97 (approved):
Fixed minor typos and format issues.
This approval is for docs only -- I did not run the example

PR #726, opened 2021-01-20
GEODE-8847: Change from boost::asio asynchronous name resolution to s…

albertogpz:
…ynchronous

When creating a TCP connection and name resolution is needed, the
synchronous version of the boost::asio::ip::tcp::resolver class
is used instead of the asynchronous one.

The reason is that the asynchronous one does not offer any advantage
over the synchronous one because even if a timeout is set when
using the async one it is later required by the io_service to wait for the
thread launched by the asynchronous call to finish.

### DIFF
--- a/cppcache/src/TcpConn.cpp
+++ b/cppcache/src/TcpConn.cpp
@@ -113,16 +113,11 @@ TcpConn::TcpConn(const std::string ipaddr,
 TcpConn::TcpConn(const std::string host, uint16_t port,
                  std::chrono::microseconds timeout, int32_t maxBuffSizePool)
     : socket_{io_context_} {
-  auto beforeResolvePoint = std::chrono::system_clock::now();
-  auto results = resolve(host, port, timeout);
-  auto elapsedTime = std::chrono::duration<double, std::micro>(
-      std::chrono::system_clock::now() - beforeResolvePoint);
+  auto results = resolve(host, port);
 
   // We must connect first so we have a valid file descriptor to set options
   // on.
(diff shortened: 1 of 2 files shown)


cppcache/src/TcpConn.cpp
```diff
@@ -113,16 +113,11 @@ TcpConn::TcpConn(const std::string ipaddr,
 TcpConn::TcpConn(const std::string host, uint16_t port,
                  std::chrono::microseconds timeout, int32_t maxBuffSizePool)
     : socket_{io_context_} {
-  auto beforeResolvePoint = std::chrono::system_clock::now();
-  auto results = resolve(host, port, timeout);
-  auto elapsedTime = std::chrono::duration<double, std::micro>(
-      std::chrono::system_clock::now() - beforeResolvePoint);
+  auto results = resolve(host, port);
```
pivotal-jbarrett, line 116: I believe it is in intentional that the resolve of the host is also part of the connection timeout. Now it looks like resolve could take an indeterminate amount of time followed by only connect limited by timeout.

pivotal-jbarrett (changes requested):

cppcache/src/TcpConn.cpp
pivotal-jbarrett: Ok, looking at the original organization of the code using ACE the resolve step was not included in the timeout. The resolve happened on the construction of the `TcpConn` when it initialized the `ACE_INET_Addr` member. The timeout was only used for the connect phase. This change seems consistent with that change.

pivotal-jbarrett (approved):

PR #727, opened 2021-01-21
GEODE-8868: Add thread_id parameter to gnmsg command line

pdxcodemonkey:
- filter messages by thread, so we can see which responses correspond to
  which requests
- also decoded the CONTAINS_KEY message in the course of other debugging


echobravopapa (approved):
Nice addition!

PR #717, opened 2021-01-08
GEODE-8820: Fix Windows build for latest ACE

mmartell:
This change allows the Windows SDK version for ACE to be specified using CMAKE_SYSTEM_VERSION, which can be set appropriately for the target machine.

Our CI uses Windows SDK version 10.0.16299.0,, so passing -DCMAKE_SYSTEM_VERSION=10.0.16299.0 to the cmake config step fixes our CI build failure on Windows. Note, this also requires a change to our CI code which generates the cmake command.

PR #728, opened 2021-01-22
Fix LGTM Usless Variable Assigment in C# files

moleske:
Draft cause there's a lot of them.  Also opening this pr so that lgtm can run through it.

Also if anyone is reading this, my Windows CI seems to have become sad because of

[`error MSB8036: The Windows SDK version 10.0.14393 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution"`](https://github.com/moleske/geode-native/runs/1691403365?check_suite_focus=true#step:11:270)

in building ACE.  Did something change that I missed?  Was using Visual Studio 2017, should be trying to add that SDK version in Visual Studio installer?


pdxcodemonkey:
-DCMAKE_SYSTEM_VERSION=10.0.16299.0, see Mike's PR here: https://github.com/apache/geode-native/pull/717

PR #731, opened 2021-01-27
GEODE-8874: Fix for transaction commit failure

jvarenina:
Issue: Transaction commit fail when m_transactionID integer overflows to
negative value, because server interprets negative value as
non-transactional message.

Fix: Client is impacted to reset transactionID to one when INT_32MAX
value is reached in order to avoid overflow.

### DIFF
--- a/cppcache/src/TXId.cpp
+++ b/cppcache/src/TXId.cpp
@@ -30,11 +30,18 @@ namespace client {
 std::atomic<int32_t> TXId::m_transactionId(0);
 
 TXId::TXId() {
-  // If m_transactionId has reached maximum value, then set
-  // it to zero in order to avoid overflow to negative value.
-  auto maxValueTXId = INT32_MAX;
-  m_transactionId.compare_exchange_strong(maxValueTXId, 0);
-  m_TXId = ++m_transactionId;
+  // If m_transactionId has reached maximum value, then start again
+  // from zero, and increment by one until first unused value is found.
+  // This is done to avoid overflow of m_transactionId to negative value.
+  auto maxValueTXId = std::numeric_limits<int>::max();
+  auto current = m_transactionId.load();
+  decltype(current) next;
+  do {
+    next = std::numeric_limits<decltype(current)>::max() == current
+               ? 1
+               : current + 1;
+  } while (!m_transactionId.compare_exchange_strong(current, next));
+  m_TXId = next;
 }
 
 TXId& TXId::operator=(const TXId& other) {
--- a/cppcache/src/TXId.cpp
+++ b/cppcache/src/TXId.cpp
@@ -33,7 +33,6 @@ TXId::TXId() {
   // If m_transactionId has reached maximum value, then start again
   // from zero, and increment by one until first unused value is found.
   // This is done to avoid overflow of m_transactionId to negative value.
-  auto maxValueTXId = std::numeric_limits<int>::max();
   auto current = m_transactionId.load();
   decltype(current) next;
   do {
--- a/cppcache/test/TXIdTest.cpp
+++ b/cppcache/test/TXIdTest.cpp
@@ -24,11 +24,11 @@ namespace geode {
 namespace client {
 
 TEST(TXIdTest, testIncrementTransactionID) {
-  auto tx = new TXId();
-  ASSERT_EQ(1, tx->getId());
+  TXId tx1;
+  ASSERT_EQ(1, tx1.getId());
 
-  tx = new TXId();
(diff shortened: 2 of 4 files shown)


cppcache/src/TXId.cpp
```diff
@@ -27,15 +27,27 @@ namespace apache {
...
-std::atomic<int32_t> TXId::m_transactionId(1);
+std::atomic<int32_t> TXId::m_transactionId(0);
 
-TXId::TXId() : m_TXId(m_transactionId++) {}
+TXId::TXId() {
+  // If m_transactionId has reached maximum value, then set
+  // it to zero in order to avoid overflow to negative value.
+  auto maxValueTXId = INT32_MAX;
+  m_transactionId.compare_exchange_strong(maxValueTXId, 0);
+  m_TXId = ++m_transactionId;
```
pivotal-jbarrett, line 37: While highly unlikely it is possible for `m_transactionId` to be at max value again between lines 36 and 37. This whole operation should be done in a loop where you fetch the current, increment if less than max or reset to 0, then compare and set the new value. If failure to set, then try again.

```c++
auto current = m_transactionId.load();
decltype(current) next;
do {
    next = std::numeric_limits<decltype(current)>::max() == current ? 1 : current + 1;
} while(!m_transactionId.compare_exchange_strong(current, next)); 

m_TXId = next;
```

cppcache/src/TXId.cpp
```diff
@@ -27,15 +27,27 @@ namespace apache {
...
 namespace client {
 
-std::atomic<int32_t> TXId::m_transactionId(1);
+std::atomic<int32_t> TXId::m_transactionId(0);
 
-TXId::TXId() : m_TXId(m_transactionId++) {}
+TXId::TXId() {
+  // If m_transactionId has reached maximum value, then set
+  // it to zero in order to avoid overflow to negative value.
+  auto maxValueTXId = INT32_MAX;
```
pivotal-jbarrett, line 35:
```c++
auto maxValueTXId = std::numeric_limits<int>::max();
```
https://en.cppreference.com/w/cpp/types/numeric_limits

pivotal-jbarrett (changes requested):

cppcache/src/TXId.cpp
jvarenina: You are right. I have updated solution as you suggested.

cppcache/src/TXId.cpp
jvarenina: Thanks, done.

cppcache/src/TXId.cpp
```diff
@@ -27,15 +27,34 @@ namespace apache {
...
 
-std::atomic<int32_t> TXId::m_transactionId(1);
+std::atomic<int32_t> TXId::m_transactionId(0);
 
-TXId::TXId() : m_TXId(m_transactionId++) {}
+TXId::TXId() {
+  // If m_transactionId has reached maximum value, then start again
+  // from zero, and increment by one until first unused value is found.
+  // This is done to avoid overflow of m_transactionId to negative value.
+  auto maxValueTXId = std::numeric_limits<int>::max();
```
pivotal-jbarrett, line 36: Unused.

pivotal-jbarrett (changes requested):

pivotal-jbarrett (approved):

cppcache/test/TXIdTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <gtest/gtest.h>
+
+namespace apache {
+namespace geode {
+namespace client {
+
+TEST(TXIdTest, testIncrementTransactionID) {
+  auto tx = new TXId();
+  ASSERT_EQ(1, tx->getId());
+
```
pdxcodemonkey, line 29: need `delete(tx);` ?

pdxcodemonkey (changes requested):

cppcache/test/TXIdTest.cpp
pivotal-jbarrett: Use `std::unique_ptr` if this needs to even be a pointer. Probably just as testable as a stack allocated variable.

pivotal-jbarrett (changes requested):

cppcache/test/TXIdTest.cpp
jvarenina: Thanks. I have changed it to stack allocated variables.

pivotal-jbarrett (approved):

PR #732, opened 2021-01-28
GEODE-8884: Update ACE to 7.0.0

moleske:

I didn't really want to update ACE, but I really wanted to be able to [build on MacOS 11 (Big Sur)](https://github.com/moleske/geode-native/runs/1773671597?check_suite_focus=true) so 🤷‍♀️ 

Seems to work in my tests for the existing operating systems, but will let folks with fancy pipelines tell me I'm wrong

### DIFF
--- a/dependencies/ACE/CMakeLists.txt
+++ b/dependencies/ACE/CMakeLists.txt
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-project( ACE VERSION 6.5.12 LANGUAGES NONE )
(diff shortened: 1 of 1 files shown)


echobravopapa (dismissed):
Maybe this is worth a dev list discussion... for the most part the desire is to remove all usages of ACE, going to a new version has any value add... aside from compiling on BigSur

mkevo (approved):
I agree with @moleske that we need to stay up to date as it is still in the use. In the past we also updated this to have the newest version with all fixes without dev list discussion.

pdxcodemonkey:
Started up a "fancy pipeline", will approve/merge if all looks good.

PR #733, opened 2021-01-29
GEODE-8887: Refactor EventIdTSS class

pdxcodemonkey:
- Use Meyers singleton pattern for global singleton (threadId), and
  thread local singleton EventIdTSS
- rename all member variables in EventId and EventIdTSS classes
- clean up logic slightly around global threadId atomic var
- rename EventId getter methods for clarity

### DIFF
--- a/cppcache/src/EventId.cpp
+++ b/cppcache/src/EventId.cpp
@@ -16,9 +16,8 @@
  */
 #include "EventId.hpp"
 
-#include <inttypes.h>
-
 #include <atomic>
+#include <cstdint>
 #include <cstring>
 
 #include <geode/DataInput.hpp>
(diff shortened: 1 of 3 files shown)


cppcache/src/EventId.cpp
```diff
@@ -16,83 +16,99 @@
...
+};
 
+class EventIdTSS {
  public:
-  // this should get called just once per thread due to first access to TSS
-  EventIdTSS() {
-    m_eidThrTSS = ++s_eidThrId;
-    m_eidSeqTSS = 0;
+  static EventIdTSS& instance() {
+    thread_local EventIdTSS idTss_;
```
pivotal-jbarrett, line 46: Can we do even better on these types and variable names.

cppcache/src/EventId.hpp
```diff
@@ -20,12 +20,16 @@
 #ifndef GEODE_EVENTID_H_
 #define GEODE_EVENTID_H_
 
+#include <inttypes.h>
```
pivotal-jbarrett, line 23: Use <cstdint>.

https://en.cppreference.com/w/cpp/header/cstdint

pivotal-jbarrett (changes requested):

mreddington (approved):

pivotal-jbarrett (approved):

PR #734, opened 2021-01-30
Geode 8892 loosen log file name rules

pdxcodemonkey:
Instead of using `boost::filesystem::portable_file_name` to police log file names, try/catch boost and std exceptions when creating log file and report an error via IllegalArgumentException.


pivotal-jbarrett (approved):
YES!!

PR #735, opened 2021-01-30
GEODE-8900: Bump sqlite and sqlite-netfx to latest

moleske:
Since I was updating ACE, figured I should check the rest of the dependencies.  Seems good on the [operating systems that I tested on](https://github.com/moleske/geode-native/pull/34/checks).

As an aside, has anyone looked at trying to use [dependabot ](https://github.com/dependabot/dependabot-core) to create a CMake dependency updater?  I've seen a few people fork and run their own dependabot to keep dependencies up to date in a build system that dependabot didn't support.  I haven't looked too closely cause I didn't really want to muck with ruby.

### DIFF
--- a/dependencies/sqlite-netFx/CMakeLists.txt
+++ b/dependencies/sqlite-netFx/CMakeLists.txt
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-project( sqlite-netFx VERSION 1.0.102.0 LANGUAGES NONE )
+project( sqlite-netFx VERSION 1.0.113.0 LANGUAGES NONE )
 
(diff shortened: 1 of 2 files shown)


pivotal-jbarrett:
> As an aside, has anyone looked at trying to use [dependabot ](https://github.com/dependabot/dependabot-core) to create a CMake dependency updater?  I've seen a few people fork and run their own dependabot to keep dependencies up to date in a build system that dependabot didn't support.  I haven't looked too closely cause I didn't really want to muck with ruby.

I have been playing with Conan. ACE is the tough one to get working. I put it on the shelf until ACE is gone. So close!!

pdxcodemonkey (approved):

PR #736, opened 2021-01-30
GEODE-8872: added API impacts to native client

mivanac:
Native client impacts for ticket GEODE-8872

### DIFF
--- a/clicache/src/PoolFactory.cpp
+++ b/clicache/src/PoolFactory.cpp
@@ -425,6 +425,26 @@ namespace Apache
 	   }
 
 
+      PoolFactory^ PoolFactory::SetRequestLocatorInternalAddress( bool requestInternal )
+      {
+			  _GF_MG_EXCEPTION_TRY2/* due to auto replace */
+
+          try
+          {
+            m_nativeptr->get()->setRequestLocatorInternalAddress( requestInternal );
+          }
+          finally
+          {
+            GC::KeepAlive(m_nativeptr);
+          }
+
+			  _GF_MG_EXCEPTION_CATCH_ALL2/* due to auto replace */
+          return this;
+      }
+
+
+
+
 		  PoolFactory^ PoolFactory::Reset()
       {
 			  _GF_MG_EXCEPTION_TRY2/* due to auto replace */
--- a/clicache/src/PoolFactory.cpp
+++ b/clicache/src/PoolFactory.cpp
@@ -36,445 +36,443 @@ namespace Apache
       using namespace System;
 
       PoolFactory^ PoolFactory::SetFreeConnectionTimeout( TimeSpan connectionTimeout )
-		  {
-			  _GF_MG_EXCEPTION_TRY2/* due to auto replace */
-
-			  try
-			  {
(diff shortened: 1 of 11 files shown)


clicache/src/PoolFactory.cpp
```diff
@@ -425,6 +425,26 @@ namespace Apache
 	   }
 
 
+      PoolFactory^ PoolFactory::SetRequestLocatorInternalAddress( bool requestInternal )
+      {
+			  _GF_MG_EXCEPTION_TRY2/* due to auto replace */
```
pivotal-jbarrett, line 430: C++/CLI formatting looks a bit off.

cppcache/src/ThinClientLocatorHelper.cpp
```diff
@@ -123,7 +124,8 @@ std::shared_ptr<Serializable> ThinClientLocatorHelper::sendRequest(
     auto conn = createConnection(location);
     auto data =
         m_poolDM->getConnectionManager().getCacheImpl()->createDataOutput();
-    data.writeInt(static_cast<int32_t>(1001));  // GOSSIPVERSION
+    data.writeInt(static_cast<int32_t>(1002));  // GOSSIPVERSION
```
pivotal-jbarrett, line 127: Let's move this GOSSIPVERSION to a constant somewhere

cppcache/src/ThinClientLocatorHelper.cpp
```diff
@@ -123,7 +124,8 @@ std::shared_ptr<Serializable> ThinClientLocatorHelper::sendRequest(
...
+    data.writeInt(static_cast<int16_t>(125));   // GEODE 1.14.0
```
pivotal-jbarrett, line 128: Use `Version::getOrdinal()`. You will need to update the ordinal to 1.14.0. This also means any protocol changes from 1.0.0, the current value, to 1.14.0 need to be implemented.

cppcache/src/LocatorListRequest.hpp
```diff
@@ -34,13 +34,16 @@ class Serializable;
 class LocatorListRequest : public ServerLocationRequest {
  private:
   std::string m_servergroup;
+  bool m_requestInternalAddress;
 
  public:
-  explicit LocatorListRequest(const std::string& servergroup = "");
+  explicit LocatorListRequest(const std::string& servergroup = "",
```
pivotal-jbarrett, line 40: I suspect this signature has some really strange undefined behavior and now might be a time to fix it. We should probably go a step further and fix the default value assigned to a `std::string` reference. I think the keyword `explicit` can be dropped now since there is more than one argument. Maybe we default the first argument can convert but.. 🤷  What does clang-tidy say?

pivotal-jbarrett (changes requested):

pdxcodemonkey:
@mivanac what's the current status of this one?  Looks like Jake's last concern still needs to be addressed, and CI jobs haven't been run.  Pushing an empty commit should wake up CI...

mivanac:
This PR is currently paused, since PR for java (server) impacts is currently blocked.

PR #741, opened 2021-02-04
GEODE-8921: Enable "extra qualification" warning in Windows builds

pdxcodemonkey:
- This would bite developers from time to time, because code with extra
  qualifications will build on Windows, but all other platforms will
  flag it with a warning and break the build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,9 +280,9 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
 
   target_compile_options(_WarningsAsError INTERFACE
     /WX
-	/wd4996
-  /we4596
-	)
+    /wd4996
+    /we4596
+  )
 
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /ignore:4099")
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /ignore:4099")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,6 +281,7 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   target_compile_options(_WarningsAsError INTERFACE
     /WX
 	/wd4996
+  /we4596
 	)
 
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /ignore:4099")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,7 +281,6 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   target_compile_options(_WarningsAsError INTERFACE
     /WX
 	/wd4996
-  /we4596
 	)
 
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /ignore:4099")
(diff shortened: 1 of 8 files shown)


echobravopapa (approved):
Leave it to `cli` ...   makes sense in a `c++` context, one would expect to see that in an implementation file...

moleske (approved):
Looks good, though don't have my windows machine setup to actually verify.

There's some interesting indentation in CMakeLists.txt and PdxHelper.hpp if you end up making other edits to those file

cppcache/integration/benchmark/RegionBM.cpp
```diff
@@ -19,6 +19,17 @@
 #include <framework/Cluster.h>
 #include <framework/Gfsh.h>
 
+// Disable warning for "extra qualifications" here.  One of the boost log
+// headers triggers this warning.  Note: use of disable pragma here is
+// intentional - attempts to use push/pop as you ordinarily should just
+// yielded a gripe from the MS tools that "warning number '4596' is not a
+// valid compiler warning". re-enabling the warning after the include
+// fails in the same way, so just leave it disabled for the rest of the
+// file.  This is safe, since the warning can only trigger inside a class
```
moleske, line 28: There's a class defined on line 51?  Or did you mean something else by this comment?

clicache/src/impl/PdxHelper.hpp
```diff
@@ -46,15 +46,15 @@ namespace Apache
...
+          static IPdxSerializable^ DeserializePdx(DataInput^ dataInput, bool forceDeserialize, int typeId, int length, const native::SerializationRegistry* serializationRegistry);
 
           literal Byte PdxHeader = 8;
 
           static Int32 ReadInt32(System::Byte* offsetPosition);
 
           static Int32 ReadInt16(System::Byte* offsetPosition);
 
-					static Int32 PdxHelper::ReadUInt16(System::Byte* offsetPosition);
+					static Int32 ReadUInt16(System::Byte* offsetPosition);
```
moleske, line 57: Still weird indentation if you decide it's worth it

CMakeLists.txt
```diff
@@ -281,6 +281,7 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   target_compile_options(_WarningsAsError INTERFACE
     /WX
 	/wd4996
+  /we4596
```
moleske, line 284: I was going to write indentation, but I actually am coming around to indenting every line with different indentation

moleske (approved):
Seems reasonable to me, just one clarification and some indentation I mentioned before

clicache/src/impl/PdxHelper.hpp
echobravopapa: please fix this - its a pain that just continues if we don't keep it tidy...

CMakeLists.txt
echobravopapa: 🙄 🤦

echobravopapa (approved):

cppcache/integration/benchmark/RegionBM.cpp
pdxcodemonkey: What I'm saying is there's _no_ class defined in this file, cause that's a thing typically done in headers.  Thus, leaving the warning disabled for the rest of the .cpp file has no effect.

clicache/src/impl/PdxHelper.hpp
pdxcodemonkey: Meh - tabs.  Fixed.

PR #743, opened 2021-02-09
GEODE-8929: fix 'leaked disk space' issue in logger

pdxcodemonkey:
logger wasn't calculating size of existing logs at initialization, so stopping and restarting an app without deleting logs would cause the logger to 'grow' used disk space by up to `log-disk-space-limit` each iteration.

Also added appropriate tests, and fixed a default filename issue.

Co-authored-by: Matthew Reddington <mreddington@vmware.com>

### DIFF
--- a/cppcache/src/Log.cpp
+++ b/cppcache/src/Log.cpp
@@ -138,6 +138,19 @@ void Log::removeOldestRolledLogFile() {
     boost::filesystem::remove(fileToRemove);
     g_rollFiles.erase(index);
     g_spaceUsed -= fileSize;
+  } else {
+    throw IllegalStateException(
+        "Failed to free sufficient disk space for logs");
+  }
+}
+
+void Log::calculateUsedDiskSpace() {
+  g_spaceUsed = 0;
+  if (boost::filesystem::exists(g_fullpath)) {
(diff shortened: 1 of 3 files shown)


cppcache/src/Log.cpp
```diff
@@ -105,8 +105,8 @@ void Log::validateSizeLimits(int64_t fileSizeLimit, int64_t diskSpaceLimit) {
 void Log::init(LogLevel level, const char* logFileName, int32_t logFileLimit,
                int64_t logDiskSpaceLimit) {
   auto logFileNameString =
-      logFileName ? std::string(logFileName) : std::string();
-  init(level, std::string(logFileNameString), logFileLimit, logDiskSpaceLimit);
+      logFileName ? std::string(logFileName) : std::string("geode-native.log");
```
echobravopapa, line 108: is there a test checking the repeated use of the default log file name?

echobravopapa (changes requested):
I think there is a test case needed and Travis is red...

pivotal-jbarrett:
We really need to get out of the business of writing our own logger! ;)

echobravopapa (approved):

PR #747, opened 2021-02-19
GEODE-8949: Fix broken build inconsistent-missing-override

moleske:
Was failing to build on Mac

PR https://github.com/apache/geode-native/pull/746 caused a failure on Mac only it seems due to `SetUp` function from gTest not using override correctly.  Since the body was commented out and `scrubTestLogFiles` is used in `TearDown`, I just deleted the function to solve the problem

### DIFF
--- a/cppcache/test/LoggingTest.cpp
+++ b/cppcache/test/LoggingTest.cpp
@@ -94,6 +94,8 @@ class LoggingTest : public testing::Test {
     }
   }
 
+  void SetUp() override { scrubTestLogFiles(); }
+
   void TearDown() override { scrubTestLogFiles(); }
 
  public:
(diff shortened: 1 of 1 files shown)


echobravopapa (approved):
Less Code is Better Code ™️

pdxcodemonkey (changes requested):
This needs to be there, and the call to `scrubTestLogFiles` uncommented, in order to ensure tests function correctly.  Sorry, I missed that the body of this function had been commented out.  It should never have gone into `develop` like that.  Likewise, I am liable to miss Mac build stuff from time to time, because I'm working primarily on a Windows box these days and MacOS isn't part of our CI.

mmartell (approved):

mmartell:
Looks like PR pipeline has issues. Although this PR had 2 approvals, shouldn't have been able to merge due to failing Travis (clang-analyzer problem with ACE). Investigating.

PR #752, opened 2021-02-24
Remove obsolete Travis related files

mmartell:
With our new CI we no longer need to use Travis. So the two Travis files below can be deleted:

- .travis.yml
- docker/Dockerfile


pivotal-jbarrett:
Can you make sure you merge with the JIRA ticket for the new CI.

PR #754, opened 2021-02-26
GEODE-8949: Versioning improvements

pivotal-jbarrett:
* Download the latest matching version of Geode Java bits from Maven repository.
* Adds semver and Geode artifact version configuration.
* Adds minor version bump job.
* Triggers packer image builds on packer source changes.
* Removes Geode from packer images.
* Uploads last Geode to heavy lifters.


ci/lib/templates.lib.txt
```diff
@@ -90,7 +90,7 @@ run-clang-tidy -j $(nproc) -quiet
 
 (@ def clang_format_bash_task(): -@)
 set -ueo pipefail
-export GEODE_HOME=$(pwd)/geode/expanded/apache-geode-$(cat geode/version)
+export GEODE_HOME=$(pwd)/geode
```
pdxcodemonkey, line 93: Why set GEODE_HOME here, and use -DGEODE_ROOT in cmake configure line below?  It'd be great to do this consistently, probably using cmake variable in both cases.

pivotal-jbarrett: Sort of out of scope for this PR. I agree making them consistent would be nice.

pdxcodemonkey (approved):

PR #758, opened 2021-03-06
GEODE-9008: associate tid with connection when parsing message about

pdxcodemonkey:
"receiving header"
- This association wasn't happening if the connection on this thread
  hadn't already sent a message.  When that happened, we'd drop the
  trace with the message body and mess up the state machine.


mreddington (approved):

mmartell (approved):
Looking more robust all the time.

PR #761, opened 2021-03-09
GEODE-8679: switch to spdlog

pdxcodemonkey:


CMakeLists.txt
```diff
@@ -285,8 +285,9 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /ignore:4099")
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /ignore:4099")
 
-  # Enables multiprocess compiles
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+  # /MP Enables multiprocess compiles
+  # SPDLOG_NO_TLS disables use of tls code which clashes with managed build on Windows
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP /DSPDLOG_NO_TLS")
```
pivotal-jbarrett, line 290: This sort for definition specific to a dependency should be defined in that dependency.

cppcache/integration-test/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ target_link_libraries(test-cppcache-utils
   PUBLIC
     apache-geode
     framework
+    SPDLog::SPDLog
```
pivotal-jbarrett, line 33: Why are the integration test dependent on the `SPDLog` library? It's use should be hidden at the integration level. The library should be statically compiled into the `apache-geode` library.

cppcache/src/Log.cpp
```diff
@@ -14,70 +14,41 @@
...
-LogLevel Log::s_logLevel = LogLevel::Default;
-
-/*****************************************************************************/
+const int __1K__ = 1024;
+const int __1M__ = __1K__ * __1K__;
+const int __1G__ = __1K__ * __1M__;
+const int LOG_SCRATCH_BUFFER_SIZE = 16 * __1K__;
 
-LogLevel Log::logLevel() { return s_logLevel; }
+static std::recursive_mutex g_logMutex;
```
pivotal-jbarrett, line 45: Why still global variables in logging?

cppcache/src/Log.cpp
```diff
@@ -107,334 +78,280 @@ void Log::init(LogLevel level, const char* logFileName, int32_t logFileLimit,
...
+    } else if (logDiskSpaceLimit && logFileSizeLimit > logDiskSpaceLimit) {
+      IllegalArgumentException ex(
+          "File size limit must be smaller than disk space limit for "
+          "logging.");
+      throw ex;
+    }
 
-    g_fullpath =
-        boost::filesystem::absolute(boost::filesystem::path(logFileName));
+    int32_t adjustedFileSizeLimit;
```
pivotal-jbarrett, line 204: spdlog has its own log rotation, why are we implementing our own?

cppcache/test/LoggingTest.cpp
```diff
@@ -315,18 +315,14 @@ TEST_F(LoggingTest, logInit) {
   apache::geode::client::Log::close();
   boost::filesystem::remove("LoggingTest (#).log");
 
-  // Init with invalid filename
+#ifdef WIN32
```
pivotal-jbarrett, line 318: Strikes me that this should be its own test.

cppcache/test/LoggingTest.cpp
```diff
@@ -103,7 +103,7 @@ class LoggingTest : public testing::Test {
                                  const char* filename, int32_t rollIndex) {
     auto baseName = boost::filesystem::path(filename).stem().string();
     auto rolledPath =
-        logdir / boost::filesystem::path(baseName + "-" +
+        logdir / boost::filesystem::path(baseName + "." +
```
pivotal-jbarrett, line 106: By nature, any test that using the filesystem can't be a unit test. This test should be moved to integration/test or use some in-memory log sink.

dependencies/spdlog/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+target_include_directories(${PROJECT_NAME} SYSTEM INTERFACE
+  $<BUILD_INTERFACE:${INSTALL_DIR}/include>
+)
+
+target_link_libraries(${PROJECT_NAME} INTERFACE
+  spdlog
+)
+
+add_dependencies(${PROJECT_NAME} ${PROJECT_NAME}-extern)
+add_library(SPDLog::SPDLog ALIAS spdlog)
```
pivotal-jbarrett, line 50: Given the project name, library name, namespace is "spdlog", I think this should probably be `spdlog::spdlog`. The CMake naming convention isn't super strong here but tends towards "project":"library" matching the case of the respective component.

pivotal-jbarrett (changes requested):

cppcache/src/Log.cpp
pdxcodemonkey: We're not... really.  Our two existing properties are file size and disk space, specified in MB.  spdlog uses two properties, file size and max number of files, specified in bytes.  So we calculate limits like we did before, then translate disk space to a max number of files and give those limits to the spdlog file sink.  There's no logic anymore in our code to check file size, roll logs, delete old files, etc.

cppcache/integration-test/CMakeLists.txt
pdxcodemonkey: This may or may not be vestigial - I'll have to check.  spdlog is a header-only library, at least as we're using it, so this dependency just means we need to include one of its headers.  But again, I'm not sure that's the case any more, so I will try to remove this ref.

CMakeLists.txt
pdxcodemonkey: Kind of looks like someone's bringing in a spdlog header file that shouldn't be.  Will investigate.

pdxcodemonkey: Removed.

cppcache/src/Log.cpp
```diff
@@ -45,9 +45,21 @@ const int LOG_SCRATCH_BUFFER_SIZE = 16 * __1K__;
 static std::recursive_mutex g_logMutex;
 static std::shared_ptr<spdlog::logger> currentLogger;
 static LogLevel currentLevel = LogLevel::None;
+static std::string logFilePath;
+static int32_t adjustedFileSizeLimit;
+static int32_t maxFiles;
 
 const std::shared_ptr<spdlog::logger>& getCurrentLogger() {
-  return currentLogger;
+  if (logFilePath.empty()) {
```
pivotal-jbarrett, line 53: In the Java bits we landed on `-`, consistent with lots of utilities, as the indicator to use `stdout` logging.

pivotal-jbarrett (changes requested):

cppcache/test/LoggingTest.cpp
pdxcodemonkey: Integration tests should only be using the public API and dynamically linking to geode-native.  The logger is not a public API.  I'm unsure how I would go about testing file size and disk space limits using an in-memory sink - did you have something in mind?

cppcache/integration-test/CMakeLists.txt
pdxcodemonkey: removed

cppcache/integration-test/LibraryCallbacks.cpp
```diff
@@ -29,7 +29,7 @@ void dummyFunc() {}
 }  // namespace test
 
 #define SLEEP(x) std::this_thread::sleep_for(std::chrono::milliseconds(x))
-#define LOG LOGDEBUG
+#define LOG LOG_DEBUG
```
mreddington, line 32: Our tests shouldn't be using the NC library logger, there's just no need or justification for it. The test harness should be responsible for all reporting. If the mocks and stated expectations and assertions are inadequate to correctly deduce the nature of a failed test, that indicates the test is inadequate. The test harness provides all the facilities necessary to express expected sequences of operations and additional information. Each expectation and and assertion returns a stream object for logging additional information. For example:

`EXPECT_EQ(foo, bar) << "I BLAME ZOIDBERG!";`

There are other facilities for logging additional information, like the `Gfsh` command output, that will only appear in the case of a failure. The test harness should be quiet, with a clean indication of success, and fail with a complete report. The harness knows how to do it. I recommend you map `LOG` to a no-op, and we begin removing all logging from tests.

dependencies/spdlog/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+project( spdlog VERSION 1.3.1 LANGUAGES NONE )
+
+set( SHA256 db6986d0141546d4fba5220944cc1f251bd8afdfc434bda173b4b0b6406e3cd0 )
+
+include(GNUInstallDirs)
+include(ExternalProject)
+
+ExternalProject_Add( ${PROJECT_NAME}-extern
+  URL "https://github.com/gabime/spdlog/archive/v${PROJECT_VERSION}.zip"
```
mreddington, line 24: Is there a way we can track the tip of spdlog? Is that desirable?

cppcache/integration-test/LibraryCallbacks.cpp
pdxcodemonkey: Agree 100% in principle, but I think this is beyond the scope of the current PR.  To give you an idea of the size of this task, here is the result of "find in files" for `LOG`, whole-word only, so I'd only catch test code that was using some variation of this hokey define:

`Matching lines: 2609    Matching files: 112    Total files searched: 1167`

That's a whooooole lot of logging to go remove from code that is targeted for deprecation.

cppcache/src/Log.cpp
pdxcodemonkey: I was able to eliminate most of these, but several are still needed due to the original implementation of our logging.  To wit, we need a mutex, the shared pointer to the logger, the current level, and a global singleton to initialize the logger at startup.

cppcache/src/Log.cpp
pdxcodemonkey: Modified the code to accept this as an explicit request for `stdout` logging.

dependencies/spdlog/CMakeLists.txt
pdxcodemonkey: Dunno, but for sure we're already way out of date.  Current version is 1.8.5, will update.

cppcache/test/LoggingTest.cpp
pdxcodemonkey: Moved file/disk space init and filename init tests to their own test cases.

PR #762, opened 2021-03-11
GEODE-9026: Fix RegisterKeysTest race condition

gaussianrecurrence:
- Sometimes it could happen that in RegisterAllWithConsistencyDisabled
   test the call to afterDestroy happens before the expectance is set.
   THis leads to the test sporiously failing.
 - Expect call has been moved after the listener instantiation to avoid
   this race condition.


rhoughton-pivot:
The CI change to need Concourse statuses to pass is causing an issue with this PR. We'll work to fix the jobs ASAP so that the PR is mergable.

pdxcodemonkey (approved):
Looks good, and all tests are passing.  Ship it.

PR #763, opened 2021-03-11
GEODE-8973: Fix calls dup on exception handling

gaussianrecurrence:
- Whenever calling throwExceptionIfError macro it can happen that calls
   are duplicated. Hence the macro has been replaced by a inlined
   function.


pdxcodemonkey (approved):
Looks good to me.

PR #768, opened 2021-03-16
Remove log line fixed in GEODE-8994

pdxcodemonkey:
- One of our integration tests has started failing on Ubuntu after making this
  change, just attempting to determine if log line is the cause.


pdxcodemonkey:
Test issue seen with the prior fix is almost certainly due to the issue fixed in https://github.com/apache/geode-native/pull/762.  Closing this, and will rebase the earlier PR.

PR #769, opened 2021-03-17
GEODE-9026: Fix race cond in RegisterKeysTest #2

gaussianrecurrence:
- As it seems there is still a very rare race condition that might
   cause the test to fail.
 - If the listener is called before wait_for is called, the test fails
   due to timeout.
 - A flag and predicate has been added to tackle this race condition.


pivotal-jbarrett (approved):

gaussianrecurrence:
Sorry for the mess caused by this test. I've run in-local all the new integration-tests (with -j 5) several times. With this fix, it seems to be working fine. So I hope it causes no more trouble.

PR #774, opened 2021-03-23
GEODE-9058: Remove ACE_OS references

gaussianrecurrence:
- All references to ACE_OS have been removed in the source and replaced
   by a suitable boost alternative, an STL alternative or in few cases,
   a custom alternative.
 - Note that most of the impacts are in the old integration tests.
 - Boost::iostream added to old integration tests.
 - IMPORTANT. There are two cases in which ACE_OS references remain:
    * ExpiryTaskManager. This reference is left there as there is
      already an ongoing PR to replace the whole implementation by
      boost::asio one. PR #678
    * DistributedSystemImpl. There is a call to uname, which I've been
      unable to find a suitable alternative. The approach on this needs
      to be discussed.

### DIFF
--- a/cppcache/src/DistributedSystemImpl.cpp
+++ b/cppcache/src/DistributedSystemImpl.cpp
@@ -79,11 +79,8 @@ void DistributedSystemImpl::logSystemInformation() const {
   LOGCONFIG("Source revision: %s", PRODUCT_SOURCE_REVISION);
   LOGCONFIG("Source repository: %s", PRODUCT_SOURCE_REPOSITORY);
 
-  ACE_utsname u;
-  ACE_OS::uname(&u);
-  LOGCONFIG(
-      "Running on: SystemName=%s Machine=%s Host=%s Release=%s Version=%s",
-      u.sysname, u.machine, u.nodename, u.release, u.version);
+  auto sysinfo = Utils::getSystemInfo();
+  LOGCONFIG("Running on: %s", sysinfo.c_str());
   LOGCONFIG("Current directory: %s",
             boost::filesystem::current_path().string().c_str());
   LOGCONFIG("Current value of PATH: %s", Utils::getEnv("PATH").c_str());
--- a/cppcache/CMakeLists.txt
+++ b/cppcache/CMakeLists.txt
@@ -24,6 +24,8 @@ if (CMAKE_USE_PTHREADS_INIT)
   check_function_exists("pthread_setname_np" HAVE_pthread_setname_np)
 endif()
 
+check_function_exists(uname HAVE_uname)
+
 # Search OpenSSL
 find_package(OpenSSL COMPONENTS Crypto)
 
--- a/cppcache/integration-test/BBNamingContext.cpp
+++ b/cppcache/integration-test/BBNamingContext.cpp
@@ -47,7 +47,7 @@ static int hashcode(char *str) {
 }
 
 static int getRandomNum() {
-  char *testName = ACE_OS::getenv("TESTNAME");
+  char *testName = std::getenv("TESTNAME");
 
   int seed = hashcode(testName) + 11;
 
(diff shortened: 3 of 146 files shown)


pivotal-jbarrett:
I you have have mostly successful in removing ACE from the old integration tests then keep going but I don't feel like it is a priority. Having ACE as a dependency for that framework doesn't keep me awake at night or keep us from making meaningful progress in the main sources and new test framework.

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -81,9 +81,9 @@ void DistributedSystemImpl::logSystemInformation() const {
 
   ACE_utsname u;
   ACE_OS::uname(&u);
-  LOGCONFIG(
-      "Running on: SystemName=%s Machine=%s Host=%s Release=%s Version=%s",
-      u.sysname, u.machine, u.nodename, u.release, u.version);
+  LOGCONFIG("Running on: SystemName=" BOOST_PLATFORM
```
pivotal-jbarrett, line 84: A really important distinction here is that `BOOST_PLATFORM` is resolved at compile time, while `ACE_OS::name` is resolved at runtime. What are you thinking for the other components here? Roll our own system independent solution like `ACE_OS::uname`? When I looked into this I couldn't find a different path.

pivotal-jbarrett (approved):

cppcache/src/DistributedSystemImpl.cpp
gaussianrecurrence: Thanks for pointing that out. Well as you say. Writing our own solution seems like the only way :S

gaussianrecurrence:
> I you have have mostly successful in removing ACE from the old integration tests then keep going but I don't feel like it is a priority. Having ACE as a dependency for that framework doesn't keep me awake at night or keep us from making meaningful progress in the main sources and new test framework.

I am quite close to remove all references. I am currently solving Windows segmentations. Once solved, solving uname and few minors comments will be the remainings.

cppcache/src/DistributedSystemImpl.cpp
pdxcodemonkey: Unless I'm missing something, this makes a slight difference in the log output at CONFIG level and above, but otherwise has no effect on behavior of the client, since the data in question isn't used.  So how much do we care about, say, not having as much detail in Windows logs?

gaussianrecurrence: You've got a point. I did not want to change the output, but it's true that this kind of info is not much relevant and if needed for troubleshooting it can be asked to the issue reporter. If you are OK with removing the call it would make my life a whole lot more easy and future code more maintable. What you think @pivotal-jbarrett ?

pivotal-jbarrett: It is not uncommon for the person collecting the logs to have little knowledge of the host details so I suspect support staff will not be happy if the are working with a customer who can't tell them what OS and version a log came from.  In general I agree though that this is ephemeral information. It is also pretty easy to create the platform independent. You can look to the ACE implementation for hints on how to do the same.

gaussianrecurrence: That's understandable. Yes, I was planning on re-using ACE implementation, so I'll stick to that!

gaussianrecurrence: There it is. I've written an implementation for at least all the supported platforms. I've left out some old Windows versions, like Windows 95/98/Me, but I really doubt someone manages to run geode on those OS.

pivotal-jbarrett: Well, considering those are 32bit OSs, out of support and not compatible with our minimum runtime, I think we are ok!

pivotal-jbarrett (approved):

PR #776, opened 2021-03-27
GEODE-9078: Remove ACE mutexes

ghost:
- Removed all ACE mutexes.
 - Some mutexes have been left out of this change, like fwklib ones,
   as PR #777 is removing most of the test/fwklib code.


pdxcodemonkey:
@gaussianrecurrence Please rebase onto `develop` and we will review.  We merged a couple of old outstanding PRs this morning, one of which was _huge_, so there are conflicts to resolve.

gaussianrecurrence:
> @gaussianrecurrence Please rebase onto `develop` and we will review. We merged a couple of old outstanding PRs this morning, one of which was _huge_, so there are conflicts to resolve.

Will do. However note that there is one thing to look into:
```
TODO. Review mutex usage in PdxType as it is only used to read, so
either it's not necessary or we are missing write locks.
```

Anyone happens to know why is there a RW mutex to protect just read access? Seems quite odd to me

gaussianrecurrence:
> @gaussianrecurrence Please rebase onto `develop` and we will review. We merged a couple of old outstanding PRs this morning, one of which was _huge_, so there are conflicts to resolve.

Will do. However note that there is one thing to look into:
```
TODO. Review mutex usage in PdxType as it is only used to read, so
either it's not necessary or we are missing write locks.
```

Anyone happens to know why is there a RW mutex to protect just read access? Seems quite odd to me :S

cppcache/src/ReadWriteLock.cpp
```diff
@@ -23,28 +23,28 @@ namespace apache {
...
-  } while (!exitCondition);
+  } while (!exit_cond);
 }
 
-TryWriteGuard::TryWriteGuard(ACE_RW_Thread_Mutex& lock,
-                             const volatile bool& exitCondition)
-    : lock_(lock), isAcquired_(false) {
+TryWriteGuard::TryWriteGuard(boost::shared_mutex& mutex,
+                             const volatile bool& exit_cond)
+    : mutex_{mutex}, locked_{false} {
```
pivotal-jbarrett, line 40: What do you think about replacing this hot spin try lock with a condition variable that is signaled. Almost all these are around some shutdown flag, which doesn't change frequently so if the lock is contended we are just eating CPU for nothing.

cppcache/src/RemoteQuery.cpp
```diff
@@ -125,7 +125,7 @@ GfErrType RemoteQuery::executeNoThrow(
     ThinClientBaseDM* tcdm, std::shared_ptr<CacheableVector> paramList) {
   LOGFINEST("%s: executing query: %s", func, m_queryString.c_str());
 
-  TryReadGuard guard(m_queryService->getLock(), m_queryService->invalid());
+  TryReadGuard guard(m_queryService->getMutex(), m_queryService->invalid());
```
pivotal-jbarrett, line 128: This always felt very wrong to me. It's not just that locks across components is a bad smell but that we leave the leak the locking logic too. What if there was an `tryAcquireReadLock` type method that returned a guard object?

jake-at-work (changes requested):
You are my hero!

Just a few things to consider.

pivotal-jbarrett:
> ```
> TODO. Review mutex usage in PdxType as it is only used to read, so
> either it's not necessary or we are missing write locks.
> ```
> 
> Anyone happens to know why is there a RW mutex to protect just read access? Seems quite odd to me :S

I vaguely recall this. I think at the time I didn't have time to research why it only ever appears to acquire the read lock. If it only acquires a read lock then it's rather useless and should be removed. It's also possible that there should be a write lock somewhere.

cppcache/src/AdminRegion.hpp
```diff
@@ -48,7 +48,7 @@ class AdminRegion : public std::enable_shared_from_this<AdminRegion> {
   ThinClientBaseDM* m_distMngr;
   std::string m_fullPath;
   TcrConnectionManager* m_connectionMgr;
-  ACE_RW_Thread_Mutex m_rwLock;
+  boost::shared_mutex mutex_;
```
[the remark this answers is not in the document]
gaussianrecurrence: I am trying to keep changes as minimal as possible. I have to confess every time I see something wrong, like this my OCD triggers and I am tempted to change it, but as I said I am trying to hold back hehe. However if you consider this is not a big change, I will update all the member variable names to match the current standard.

cppcache/src/ReadWriteLock.cpp
gaussianrecurrence: You definetly have a point. This will require quite some changes, but I'll make it work.

cppcache/src/RemoteQuery.cpp
gaussianrecurrence: Totally agree. Will do that.

cppcache/src/ReadWriteLock.cpp
pivotal-jbarrett: We don't necessarily have to do it immediately. You could keep this in your back pocket as a future cleanup of this helper. I have wanted to fix this one for a while but just keep punting it myself. Your steps to make it boost based make it that much closer to reality though. Thanks!

gaussianrecurrence:
> > ```
> > TODO. Review mutex usage in PdxType as it is only used to read, so
> > either it's not necessary or we are missing write locks.
> > ```
> > 
> > 
> > Anyone happens to know why is there a RW mutex to protect just read access? Seems quite odd to me :S
> 
> I vaguely recall this. I think at the time I didn't have time to research why it only ever appears to acquire the read lock. If it only acquires a read lock then it's rather useless and should be removed. It's also possible that there should be a write lock somewhere.

What about

gaussianrecurrence:
Sorry for the spam. I closed the PR by mistake. Luckily I was able to re-open it.
Regarding changes requested by you @pivotal-jbarrett, I finally made all of them as suggested.

pivotal-jbarrett:
@gaussianrecurrence Big sorry! I think I have lead you down a habit hole. It looks like we never called out our deviations from the Google C++ Style Guide. I for one actually don't mind the `_` separator model in C++ but we had been preserving the `camelCase` model. We just adopted the trailing `_` for member variables. We have a few other deviations would should probably have documented too.

https://github.com/apache/geode-native/blob/develop/CONTRIBUTING.md#style

Even our examples avoid this because they are all single words. 

Let's see what other members of the team think on moving on fully to the Google `_` style?

@pdxcodemonkey @mmartell @mreddington @echobravopapa @mivanac

gaussianrecurrence:
I

> @gaussianrecurrence Big sorry! I think I have lead you down a habit hole. It looks like we never called out our deviations from the Google C++ Style Guide. I for one actually don't mind the `_` separator model in C++ but we had been preserving the `camelCase` model. We just adopted the trailing `_` for member variables. We have a few other deviations would should probably have documented too.
> 
> https://github.com/apache/geode-native/blob/develop/CONTRIBUTING.md#style
> 
> Even our examples avoid this because they are all single words.
> 
> Let's see what other members of the team think on moving on fully to the Google `_` style?
> 
> @pdxcodemonkey @mmartell @mreddington @echobravopapa @mivanac

I guess you mention it because of the make_lock methods right? I did follow another existing example to name the member function. However if you prefer to stick to the agreed camelCase notation, I can change it, no problem.

pivotal-jbarrett:
> I guess you mention it because of the make_lock methods right? I did follow another existing example to name the member function. However if you prefer to stick to the agreed camelCase notation, I can change it, no problem.

No, camelCase on method names are in the Style Guide. I mean the member and local variable names.

pivotal-jbarrett:
@gaussianrecurrence The sub group of people working on geode-native has historically been a VERY small group and we did a very poor job communicating outward and recording some of these variations. I wonder if for the sake of not holding up this PR if you should just blackout that variable renaming commit, sorry, and we start an open discussion on the dev@geode list on the C++ style guide. We should discuss deviations we want and how far we should go to try to refactor code to fit it. Then we should record the consensus in the CONTRIBUTING.md.

gaussianrecurrence:
> @gaussianrecurrence The sub group of people working on geode-native has historically been a VERY small group and we did a very poor job communicating outward and recording some of these variations. I wonder if for the sake of not holding up this PR if you should just blackout that variable renaming commit, sorry, and we start an open discussion on the dev@geode list on the C++ style guide. We should discuss deviations we want and how far we should go to try to refactor code to fit it. Then we should record the consensus in the CONTRIBUTING.md.

Yes, I also think that would be the fastest approach for this PR. Will create a new PR with that change and solving Windows compilation

cppcache/src/RemoteQuery.cpp
```diff
@@ -125,11 +127,11 @@ GfErrType RemoteQuery::executeNoThrow(
     ThinClientBaseDM* tcdm, std::shared_ptr<CacheableVector> paramList) {
   LOGFINEST("%s: executing query: %s", func, m_queryString.c_str());
 
-  TryReadGuard guard(m_queryService->getLock(), m_queryService->invalid());
-
+  boost::shared_lock<boost::shared_mutex> guard{m_queryService->getMutex()};
```
pivotal-jbarrett, line 130: Were you going to change these patterns to return a `shared_lock` rather than the `shared_mutex`?

jake-at-work (approved):
Other than the strange pattern we have about reaching into classes for their mutex this seems ready.

cppcache/src/RemoteQuery.cpp
gaussianrecurrence: > Were you going to change these patterns to return a `shared_lock` rather than the `shared_mutex`?

I tried to change it, but I found a really strange issue in the .Net part:
   - An exception with message "The string binding is invalid" was being thrown due to the fact that
   the Apache.Geode DLL has mixed code and compiles with CLR enabled.
   That's something dis-recommended by Boost because under certain
   conditions could cause problems with boost static variables.
   In this case the issue was related to one of the boost::exception
   static variables.
   This issue was solved by removing header inclusion introduced in
   revision 3.
 - Due to above issue, exposing the lock rather than the mutex is not
   possible for now, until a further refactor is made to .Net code, so
   mutex/lock exposition from revision 3 has been reverted.

So I had to revert that particular change. I am creating an issue so we can tackle this issue.

pivotal-jbarrett: .NET 😢 , got it, say no more!

gaussianrecurrence:
Regarding the pending topic we had related to having only read locks in PdxType, I went through the possible cases in which PdxType state is written and where is read I came to the following conclusion (Sorry for the long read):
## Cases in which PdxType state is accessed

- **toData**
- **(fixed|variable)LengthFieldPosition**
- **get(RemoteToLocal|LocalToRemote)Map**. This it uses lazy initialization, that's why is considered here.
- **isContains**
- **clone**
- **is(Local|Remote)TypeContains**
- **mergeVersion**
- **operator==**

If PdxType requires exlusive access for those cases where its state is mutated, then in all of the above cases, a shared access lock will be required.

## Cases in which PdxType state is mutated

- **fromData**. This is only used while requesting a PdxType to the server, so no other thread will access this type. Hence, this function does not require exclusive access.
- **add(Fixed|Variable)LengthTypeField** are only used while creating the PdxType, so it won't require exclusive access.
- **InitializeType** is called always after a PdxType has been created or fetched from the server, so it won't require exclusive access.
- **init(LocalToRemote|RemoteToLocal)** is called in get(LocalToRemote|RemoteToLocal)Map and InitializeType. Given InitializeType does not require exclusive access, whether or not init(LocalToRemote|RemoteToLocal) requires exclusive access depends on get(LocalToRemote|RemoteToLocal)Map.
- **get(RemoteToLocal|LocalToRemote)Map** uses lazy initialization, meaning that if init(LocalToRemote|RemoteToLocal) was previously called, it won't mutate the state, and it's used by:
  - **PdxRemoteWriter::PdxRemoteWriter**
  - **PdxRemoteWriter::writeUnreadFields**
  - **PdxLocalReader::PdxLocalReader**
  - **PdxLocalReader::getPreservedData**
  - **PdxRemoteReader::PdxRemoteReader** (by inheritance)
  - **PdxReaderWithTypeCollector::PdxReaderWithTypeCollector** (by inheritance)

### **get(RemoteToLocal|LocalToRemote)Map** follow up

So, regarding **PdxRemoteWriter**, the PdxType is either initialized to:
- A merged PdxType which is initialized in all cases.
- A local PdxType which is initialized in all cases

Regarding **PdxLocalReader**, **PdxRemoteReader**, **PdxReaderWithTypeCollector**, it uses a PdxType instance which might be fetched from the PdxTypeRegistry, so the exclusive access depends on whether all the PdxTypes added to the PdxTypeRegistry are initialized.
All the calls to PdxTypeRegistry::addPdxType are in PdxHelper and in all cases PdxTypes are initialized.

Consequently, I can conclude that **get(RemoteToLocal|LocalToRemote)Map** does not require exlusive access.

With all of the above, I'd say there is no need to use locks in PdxTypes, so I will remove the mutex and shared lock from that class.
I guess probably those locks were left in the code by mistake after some refactor. I sadly could not trace the exact commit, as the change seems to have been done before open sourcing the native client.

gaussianrecurrence:
Is there anything else here to do? Maybe @pdxcodemonkey, you want to review it? Or can it be merged already? Thanks! :)

PR #783, opened 2021-04-13
Revert fix for GEODE-8231 

pdxcodemonkey:
"remove bucket server location from metadata when …server down (#615)"

This reverts commit 85951d7e23a9cb493fde16d09a56c785d09d6ba5.


cppcache/src/ClientMetadata.cpp
```diff
@@ -114,24 +114,14 @@ const std::string& ClientMetadata::getColocatedWith() {
...
-        break;
-      }
-    }
-  }
-}
-
 void ClientMetadata::getServerLocation(
     int bucketId, bool tryPrimary,
     std::shared_ptr<BucketServerLocation>& serverLocation, int8_t& version) {
+  // ReadGuard guard (m_readWriteLock);
```
echobravopapa, line 120: can we lose this dead code while we're in here... thx

cppcache/src/ClientMetadata.cpp
```diff
@@ -153,15 +143,30 @@ void ClientMetadata::getServerLocation(
...
-    LOGFINER("returning random & m_bucketServerLocationsList size is: %zu",
-             m_bucketServerLocationsList.size());
     RandGen randgen;
     serverLocation = m_bucketServerLocationsList[bucketId].at(randgen(
         static_cast<int>(m_bucketServerLocationsList[bucketId].size())));
   }
   // return m_bucketServerLocationsList[bucketId].at(0);
 }
 
+/*
```
echobravopapa, line 153: here too

echobravopapa (dismissed):

cppcache/src/ClientMetadata.cpp
echobravopapa: disregard, we'll clean up post revert...

echobravopapa (approved):

alb3rtobr:
This PR also reverts code from GEODE-8565, which was an addition over GEODE-8231 solution.
So I think this PR should contain the revert of GEODE-8688, GEODE-8565 and GEODE-8231.

pdxcodemonkey:
> This PR also reverts code from GEODE-8565, which was an addition over GEODE-8231 solution.
> So I think this PR should contain the revert of GEODE-8688, GEODE-8565 and GEODE-8231.

Agree, this is the revert of all 3 of these.  This is just a title change for the PR, though, right?  I have reviewed the 3 earlier PRs and see no remaining code from them...

mmartell (approved):

albertogpz:
I wonder if it would make more sense to write a JIRA to fix the problem caused by the GEODE-8231 PR instead of reverting 3 PRs that have been for months in the develop branch.

I do not know if I understood correctly the issue but wouldn't a possible solution be to change the eventId of the "leftover" keys (the keys for those puts that were not sent because the server went down) before sending them in the new thread to another server?

If we did so, we could keep the code in the "PRs to be reverted", fix the issue and have a test case to verify that everything works as expected.

pdxcodemonkey:
> I wonder if it would make more sense to write a JIRA to fix the problem caused by the GEODE-8231 PR instead of reverting 3 PRs that have been for months in the develop branch.
> 
> I do not know if I understood correctly the issue but wouldn't a possible solution be to change the eventId of the "leftover" keys (the keys for those puts that were not sent because the server went down) before sending them in the new thread to another server?
> 
> If we did so, we could keep the code in the "PRs to be reverted", fix the issue and have a test case to verify that everything works as expected.

I'm looking into it - don't think this one is going to be merged right away, because the fix may be simpler and IMO lower-risk than reverting all this code.  Writing the test is going to be tough - we had to modify Geode server code in order to reproduce the issue in the first place.  At any rate I'll update things here.

albertogpz:
> > I wonder if it would make more sense to write a JIRA to fix the problem caused by the GEODE-8231 PR instead of reverting 3 PRs that have been for months in the develop branch.
> > I do not know if I understood correctly the issue but wouldn't a possible solution be to change the eventId of the "leftover" keys (the keys for those puts that were not sent because the server went down) before sending them in the new thread to another server?
> > If we did so, we could keep the code in the "PRs to be reverted", fix the issue and have a test case to verify that everything works as expected.
> 
> I'm looking into it - don't think this one is going to be merged right away, because the fix may be simpler and IMO lower-risk than reverting all this code. Writing the test is going to be tough - we had to modify Geode server code in order to reproduce the issue in the first place. At any rate I'll update things here.

Great. Thanks

echobravopapa:
@albertogpz this change alters consistent behavior between clients (java & native) along with data loss, in the near term the most prudent approach is to remove the code causing the issues.

pivotal-jbarrett:
> @albertogpz this change alters consistent behavior between clients (java & native) along with data loss, in the near term the most prudent approach is to remove the code causing the issues.

@echobravopapa I have to say I disagree. The change to fix the PUTALL problem is simple. The impact the targeted removal of bucket metadata is more positive since it affects all single value operations. The only operations degraded are the multikey operations, which are more rare. The timing to get either change in is the same so I would vote to get the more complete fix in.

pdxcodemonkey:
Closed because this a) doesn't really solve the problem, and b) is much more invasive than the actual fix (see GEODE-9147).

PR #784, opened 2021-04-14
GEODE-9147: Revert to multi-hop PUTALL in the face of missing metadata

pdxcodemonkey:
- This matches the behavior of the Java client
- Our current code to "tack on" values that we don't have metadata
  for will sometimes result in EventIds reaching a server out-of-order,
  causing them to be dropped and resulting in data loss.  Resorting
  to multi-hop avoids this altogether.

### DIFF
--- a/cppcache/src/ClientMetadataService.cpp
+++ b/cppcache/src/ClientMetadataService.cpp
@@ -177,23 +177,32 @@ std::shared_ptr<ClientMetadata> ClientMetadataService::SendClientPRMetadata(
           regionPath);
   std::shared_ptr<Region> region = nullptr;
   GfErrType err = m_pool->sendSyncRequest(request, reply);
+  LOGFINE("Got reply for GET_CLIENT_PR_METADATA: err=%d, type=%d", err,
+          reply.getMessageType());
   if (err == GF_NOERR &&
       reply.getMessageType() == TcrMessage::RESPONSE_CLIENT_PR_METADATA) {
     region = m_cache->getRegion(regionPath);
+    LOGDEBUG("%s region ptr=%p", __FUNCTION__, region.get());
     if (region != nullptr) {
       if (auto lregion = std::dynamic_pointer_cast<LocalRegion>(region)) {
         lregion->getRegionStats()->incMetaDataRefreshCount();
       }
     }
     auto metadata = reply.getMetadata();
-    if (metadata == nullptr) return nullptr;
+    if (metadata == nullptr) {
+      LOGDEBUG("%s metadata is nullptr", __FUNCTION__);
+      return nullptr;
+    }
     if (metadata->empty()) {
+      LOGDEBUG("%s metadata is empty", __FUNCTION__);
       delete metadata;
       return nullptr;
     }
     auto newCptr = std::make_shared<ClientMetadata>(*cptr);
+    LOGDEBUG("%s creating new metadata object", __FUNCTION__);
     for (const auto& v : *metadata) {
       if (!v.empty()) {
+        LOGDEBUG("%s Updating bucket server locations", __FUNCTION__);
         newCptr->updateBucketServerLocations(v.at(0)->getBucketId(), v);
       }
     }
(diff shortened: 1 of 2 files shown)


albertogpz:
>     * This matches the behavior of the Java client
> 
>     * Our current code to "tack on" values that we don't have metadata
>       for will sometimes result in EventIds reaching a server out-of-order,
>       causing them to be dropped and resulting in data loss.  Resorting
>       to multi-hop avoids this altogether.

I agree with this solution as it fixes a problem in the Native client and aligns the behavior of the native and Java clients.
Nevertheless, I think the comment in the Java client for ClientMetadata::groupByServerToBuckets() should be changed as it is not aligned with the code. This is extracted from ClientMetada.java
  /**
   * This function should make a map of server to buckets it is hosting. If for some bucket servers
   * are not available due to mismatch in metadata it should fill up a random server for it.
   */
private Map<ServerLocation, Set<Integer>> groupByServerToBuckets(
      ClientPartitionAdvisor prAdvisor, Set<Integer> bucketSet, boolean primaryOnly,
      Region region) {
  
...
It should say:
"If for some bucket servers are not available due to mismatch in metadata, if singleHop, it should return null. Otherwise, it should fill up a random server for it."

pivotal-jbarrett (changes requested):
Wow, this seems like really drastic action but I guess if it is consistent with the Java client...

I would suggest that we also trigger the metadata update here. Clearly we have some metadata but it is incomplete. We should attempt to get the latest.

echobravopapa:
@pivotal-jbarrett Barry verified this IS what java client does... I was surprised as well

pivotal-jbarrett:
What remove all? Does it go through this same code path?

echobravopapa:
@pivotal-jbarrett regarding java client below:

If the java client doesn't have metadata for all the buckets, it does non-single-hop behavior.
The java ClientMetadataService.groupByServerToBuckets method has this comment, and the code matches:
// If we don't have the metadata for some buckets, return
// null, indicating that we don't have any metadata. This
// will cause us to use the non-single hop path.

pdxcodemonkey:
> Wow, this seems like really drastic action but I guess if it is consistent with the Java client...
> 
> I would suggest that we also trigger the metadata update here. Clearly we have some metadata but it is incomplete. We should attempt to get the latest.

Shouldn't be necessary, AFAIK, and might be dangerous.  In the new test code, you'll notice the keys are strings.  This is because the server is supposed to return a flag telling us to update the metadata after a PUT to the wrong server (again as I understand it).  For some reason (per @boglesby) Geode doesn't set this flag for integer keys, BUT... with string keys it does.  When geode-native sees this it triggers another metadata update, and the metadata we get back from the server _has empty buckets_.  Thus, it's a perfectly normal situation to have "incomplete" metadata, and not at all an indication that we're stale and must re-request.  Probably the effect of re-requesting is innocuous - in the case of the new test, for instance, we'd just get back complete metadata after it was too late to affect the PUTALL request.  In the situation where we have a server that's unreachable from the client, however, I could see it getting us into a feedback loop, where we request metadata because we're missing buckets, then get back metadata containing buckets for the "bad" server, then remove those buckets from the metadata, then discover missing buckets and request metadata _again_, and so on.

cppcache/src/ClientMetadataService.cpp
```diff
@@ -186,7 +184,9 @@ std::shared_ptr<ClientMetadata> ClientMetadataService::SendClientPRMetadata(
       }
     }
     auto metadata = reply.getMetadata();
-    if (metadata == nullptr) return nullptr;
+    if (metadata == nullptr) {
```
echobravopapa, line 187: yes. much more civilized...

echobravopapa (approved):
This is good, essentially parity with Java client in the related features and ensured data consistency.

echobravopapa:
@pivotal-jbarrett does @pdxcodemonkey 's explanation above satisfy your concerns?  I'm quite sure we'll explore future optimizations....

pivotal-jbarrett (approved):
I don't like this change but I'm on vacation and want to remove my block. I still think this should trigger the metadata refresh. Probably where it checks if the map is null at the calling location.

PR #786, opened 2021-04-18
GEODE-9171: Fix leaked threads in .NET clients

mmartell:
For .NET applications using the native client, threads aren't being terminated when the native client Cache is closed. This causes a huge leak of memory which can cause long running apps to crash.


clicache/src/Cache.cpp
```diff
@@ -57,6 +56,20 @@ namespace Apache
...
+        //Clean-up managed resources
+        this->!Cache();
+        _disposed = true;
+        //GC.SuppressFinalize(this) is automatically added here
+        //Base destructor is automatically called too if needed
+      }
+
+      Cache::!Cache() { //Finalizer - non-deterministic when called by GC
+        //Clean-up unmanaged resources
+        m_nativeptr->get()->~Cache();
```
pivotal-jbarrett, line 70: You should never call a destructor directly. Delete the smart pointer it will deleted and destruct the object.

pivotal-jbarrett (changes requested):

clicache/src/Cache.cpp
```diff
@@ -57,6 +56,21 @@ namespace Apache
         m_typeRegistry = gcnew Apache::Geode::Client::TypeRegistry(this);
       }
 
+      Cache::~Cache() { //Destructor - deterministic
```
pivotal-jbarrett, line 59: This sure looks like we copied comments from an example somewhere. I also don't think they provide meaningful content. The method name makes it clear it is a destructor.

clicache/src/Cache.cpp
```diff
@@ -57,6 +56,21 @@ namespace Apache
...
+        if (!_disposed) {
+          //Clean-up managed resources
+          this->!Cache();
+          _disposed = true;
```
pivotal-jbarrett, line 63: Member variables should be consistently formatted. 

Should be `m_disposed` to be consistent with old convention or `disposed_` for current convention.

clicache/src/Cache.cpp
```diff
@@ -57,6 +56,21 @@ namespace Apache
...
+          //GC.SuppressFinalize(this) is automatically added here
+          //Base destructor is automatically called too if needed
+        }
+      }
+
+      Cache::!Cache() { //Finalizer - non-deterministic when called by GC
+        //Clean-up unmanaged resources
+        delete m_nativeptr;
```
pivotal-jbarrett, line 71: Technically this value is a managed resource. If you are following the example from https://manski.net/2012/01/idisposable-finalizer-and-suppressfinalize/ then it should be deleted in the destructor.

clicache/src/Cache.cpp
pivotal-jbarrett, line 63: Not thread safe. If two threads both invoke dispose, boom! Given the C++/CLI auto generates the `Dispose` method around the destructor we really aren't given a chance to make that discoverable in the documentation. 

Safest to use an atomic check and set operation in the if check.

clicache/src/native_shared_ptr.hpp
```diff
@@ -35,7 +35,8 @@ namespace Apache
         std::shared_ptr<_T>* ptr;
 
       public:
-        native_shared_ptr(const std::shared_ptr<_T>& ptr) : ptr(new std::shared_ptr<_T>(ptr)) {}
+        native_shared_ptr(const std::shared_ptr<_T>& ptr) : ptr(new std::shared_ptr<_T>(ptr)) {
```
pivotal-jbarrett, line 38: Intentional formatting change? Almost all our empty methods are `{}` on the same line.

cppcache/include/geode/Cache.hpp
```diff
@@ -167,17 +167,17 @@ class APACHE_GEODE_EXPORT Cache : public GeodeCache {
   virtual void readyForEvents();
 
   /**
-   * Creates an authenticated cache using the given user security properties.
-   * Multiple instances with different user properties can be created with a
-   * single client cache.
+   * Creates an authenticated cache using the given user security
+   * properties. Multiple instances with different user properties can be
```
pivotal-jbarrett, line 171: Why is this suddenly changing format? Will clang-format choke on this?

clicache/integration-test2/GarbageCollectCache.cs
```diff
@@ -0,0 +1,177 @@
...
+                for (int i=0; i<numIterations; i++)
+                {
+                    diff[i] = workingSetAfter[i] - workingSetBefore[i];
+                }
+
+                Console.WriteLine("diff[99] = {0}", diff[numIterations-1]);
+            }
+        }
+    }
+}
```
pivotal-jbarrett, line 177: Missing blank line at EOF.

clicache/integration-test2/GarbageCollectCache.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                }
+
+                Int64[] diff = new Int64[numIterations];
+                for (int i=0; i<numIterations; i++)
+                {
+                    diff[i] = workingSetAfter[i] - workingSetBefore[i];
+                }
+
+                Console.WriteLine("diff[99] = {0}", diff[numIterations-1]);
```
pivotal-jbarrett, line 173: Tests should not be writing to the console.

clicache/integration-test2/GarbageCollectCache.cs
```diff
@@ -0,0 +1,177 @@
...
+                        //Assert.True(.8 < ratio && ratio < 1.3);
+                        string error = "ncThreadsBefore = " + ncThreadsBefore.ToString() +
+                            ", ncThreadsAfter = " + ncThreadsAfter.ToString();
+                        Assert.False(!(.8 < ratio && ratio < 1.3), error);
+                    }
+                }
+            }
+        }
+
+        [Fact(Skip = "Need a heuristic to filter out GC effect.")]
```
pivotal-jbarrett, line 95: Added a skipped test? What does this do for us?

pivotal-jbarrett (changes requested):

clicache/src/Cache.cpp
pivotal-jbarrett: You can also just `lock` around the `if` block.

pivotal-jbarrett (changes requested):
This PR references [GEODE-9171](https://issues.apache.org/jira/browse/GEODE-9171), which states that the threads are not shutdown when the `Cache` is closed. Calling `Cache::close()` should be shutting down all the resources associated with a Cache. This PR seems to focused on providing a deterministic destructor to avoid the indeterministic GC cleanup. While this is a good thing it isn't how we should fix this root problem. All the resources should be cleaned up on close. Destruction should cleanup the rest of the state if there is anything left.

clicache/src/Cache.cpp
mmartell: Fixed.

mmartell: Good catch.

clicache/src/Cache.cpp
mmartell: Yup, fixed.

clicache/src/Cache.cpp
mmartell: The unmanaged resources are deleted in both the destructor (which calls the finalizer) and the finalizer,

clicache/src/native_shared_ptr.hpp
mmartell: Unintentional.

cppcache/include/geode/Cache.hpp
mmartell: Not sure how this got changed. Reverting

clicache/integration-test2/GarbageCollectCache.cs
mmartell: Fixed.

clicache/integration-test2/GarbageCollectCache.cs
mmartell: Good catch!

clicache/integration-test2/GarbageCollectCache.cs
mmartell: Removing until we can come up with a heuristic that works properly in the CI.

clicache/src/Cache.cpp
pivotal-jbarrett: From my reading of this guide the C++/CLI should delete managed resources in the destructor and free native resources the finalizer.
> Beside unmanaged data, a managed class can also contain managed data, i.e. instances of managed classes implementing IDisposable. Managed data is different from unmanaged data in that it should be disposed in Dispose() but not in the finalizer. This is because instances of managed classes may already have been garbage collected when the finalizer runs.
```C++
ref class DataContainer {
public:
   ...

  ~DataContainer() {
    if (m_isDisposed)
       return;

    delete m_managedData; // dispose managed data
    this->!DataContainer(); // call finalizer
    m_isDisposed = true;
  }

  // Finalizer
  !DataContainer() {
    DataProvider::DeleteUnmanagedData(m_unmanagedData); // free unmanaged data
  }

private:
  bool m_isDisposed;
  IntPtr m_unmanagedData;
  IDisposable^ m_managedData;
};
```

This class technically only has managed data which should be deleted in the dispose/destructor.

The unmanaged data is held in the managed `native_shared_ptr` class.

clicache/src/Cache.hpp
```diff
@@ -329,6 +332,10 @@ namespace Apache
 
         Apache::Geode::Client::Internal::PdxTypeRegistry^ m_pdxTypeRegistry;
         Apache::Geode::Client::TypeRegistry^ m_typeRegistry;
+
+        Object^ m_disposedLock;
```
pivotal-jbarrett, line 336: Not necessary.

clicache/src/Cache.cpp
```diff
@@ -55,6 +54,20 @@ namespace Apache
         m_nativeptr = gcnew native_shared_ptr<native::Cache>(nativeptr);
         m_pdxTypeRegistry = gcnew Apache::Geode::Client::Internal::PdxTypeRegistry(this);
         m_typeRegistry = gcnew Apache::Geode::Client::TypeRegistry(this);
+
+        m_disposedLock = gcnew Object();
+      }
+
+      Cache::~Cache() {
+        msclr::lock lockInstance(m_disposedLock);
```
pivotal-jbarrett, line 62: lock on `this`.

clicache/src/Cache.cpp
```diff
@@ -55,6 +54,20 @@ namespace Apache
...
+        if (!m_disposed) {
+          this->!Cache();
+          m_disposed = true;
```
pivotal-jbarrett, line 65: Checking and setting `m_nativeptr` to `nullptr` would negate the need for this `m_disposed` variable.

pivotal-jbarrett (changes requested):

cppcache/include/geode/Cache.hpp
```diff
@@ -176,8 +176,8 @@ class APACHE_GEODE_EXPORT Cache : public GeodeCache {
    *
    * @see RegionService
    * @see PoolFactory#setMultiuserAuthentication(boolean)
-   * @return the {@link RegionService} instance associated with a user and given
-   *         properties.
+   * @return the {@link RegionService} instance associated with a user and
+   * given properties.
```
pivotal-jbarrett, line 180: Another random formatting change.

clicache/src/Cache.cpp
pivotal-jbarrett: In fact, this class has no need for a finalizer.

clicache/src/Cache.cpp
```diff
@@ -55,6 +54,20 @@ namespace Apache
...
         m_pdxTypeRegistry = gcnew Apache::Geode::Client::Internal::PdxTypeRegistry(this);
         m_typeRegistry = gcnew Apache::Geode::Client::TypeRegistry(this);
+
+        m_disposedLock = gcnew Object();
+      }
+
+      Cache::~Cache() {
+        msclr::lock lockInstance(m_disposedLock);
+        if (!m_disposed) {
+          this->!Cache();
```
pivotal-jbarrett, line 64: We should be forcing invocation of `Close`, if it hasn't been called already, and eating any exceptions (maybe log but no exceptions should be thrown here) before deleting the `m_nativeptr`.

pivotal-jbarrett (changes requested):

clicache/src/Cache.cpp
mmartell: Good catch. Forgot that m_nativeptr is actually a gcnew object.

clicache/src/Cache.cpp
```diff
@@ -57,6 +56,22 @@ namespace Apache
...
+      Cache::~Cache() {
+        if (m_nativeptr) {
+          try {
+            if (!IsClosed) {
+              Close();
+            }
+          }
+          catch (...) {
+          }
+          finally{
```
pivotal-jbarrett, line 68: formatting is all out of wack.

cppcache/src/ThreadPool.cpp
```diff
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 #include "ThreadPool.hpp"
```
pivotal-jbarrett, line 17: Another random format change.

pivotal-jbarrett (approved):
Just som minor formatting errors in the sea of formatting mess that C++/CLI is in anyway.

PR #790, opened 2021-04-22
GEODE-9189: Upgrades dependencies. 

pivotal-jbarrett:
Upgrades Boost to 1.76.
Upgrades ACE to 7.0.1.
Upgrades SQLite to 3.35.5
Upgrades GTest to follow head.
Upgrades Benchmark to follow head.
Fixes Xerces-C to use Apache mirrors.

### DIFF
--- a/cppcache/integration/benchmark/RegionBM.cpp
+++ b/cppcache/integration/benchmark/RegionBM.cpp
@@ -59,7 +59,7 @@ class RegionBM : public benchmark::Fixture {
   void SetUp(benchmark::State&) override {
     if (!cluster) {
       cluster = std::unique_ptr<Cluster>(
-          new Cluster(Name{name_}, LocatorCount{1}, ServerCount{1}));
+          new Cluster(::Name{name_}, LocatorCount{1}, ServerCount{1}));
       cluster->getGfsh()
           .create()
           .region()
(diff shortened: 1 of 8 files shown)


moleske:
If you're like me and were curious why gtest is following head now instead of a release, gtest now follows [live-at-head philosophy](https://abseil.io/about/philosophy#upgrade-support)

pivotal-jbarrett:
> If you're like me and were curious why gtest is following head now instead of a release, gtest now follows [live-at-head philosophy](https://abseil.io/about/philosophy#upgrade-support)

Google Benchmark also started following the same philosophy but didn't really let anymore know.

mmartell:
I strongly agree with Abseil's "Live at Head" philiosophy, which appears to be largely adopted by Google's 250 million lines of C++.

mmartell (approved):
Looks good to me. Especially like that nothing breaks.

pivotal-jbarrett:
Converted this back to draft again. There is one issue I want to resolve. CMake with either always update the git based projects on every compile or never unless forced by invoking the update target directly. I want to find a middle ground like a time based invocation, like no more than every 24h, similar to how Maven snapshots are updated.

mreddington (approved):

PR #792, opened 2021-04-26
GEODE-9193: Use local object variable for PING messages.

pdxcodemonkey:
- Prevents a leak at shutdown, ensures object is properly destroyed.

### DIFF
--- a/cppcache/src/TcrEndpoint.cpp
+++ b/cppcache/src/TcrEndpoint.cpp
@@ -510,9 +510,9 @@ void TcrEndpoint::pingServer(ThinClientPoolDM* poolDM) {
     LOGFINEST("Sending ping message to endpoint %s", m_name.c_str());
     GfErrType error;
     if (poolDM != nullptr) {
-      error = poolDM->sendRequestToEP(*pingMsg, reply, this);
+      error = poolDM->sendRequestToEP(pingMsg, reply, this);
     } else {
-      error = send(*pingMsg, reply);
+      error = send(pingMsg, reply);
     }
     LOGFINEST("Sent ping message to endpoint %s with error code %d%s",
               m_name.c_str(), error, error == GF_NOERR ? " (no error)" : "");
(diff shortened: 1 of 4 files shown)


pdxcodemonkey:
Converting to draft for the moment, trying to nail down what's going on with CI failures.

cppcache/src/TcrEndpoint.cpp
```diff
@@ -504,14 +504,15 @@ void TcrEndpoint::pingServer(ThinClientPoolDM* poolDM) {
   }
 
   if (!m_msgSent && !m_pingSent) {
-    TcrMessagePing* pingMsg = TcrMessage::getPingMessage(m_cacheImpl);
+    TcrMessagePing pingMsg(new DataOutput(m_cacheImpl->createDataOutput()),
```
mmartell, line 507: Where is DataOutput deleted? Hopefully, that's not also a leak.

gaussianrecurrence: This pointer ownership is transfered to the TcrMessage by assigning it to an unique_ptr. I am with you that the way of instantiating the object might lead to think that this is a memory leak, but this is they way is done for most of the TcrMessage implementations :S

pdxcodemonkey: Yeah - it's not pretty, but it is correct.

pivotal-jbarrett, line 507: Can you copy my PR where I modified the constructor to take a `std::unique_ptr<DataOutput>` to avoid any leaking potentials in allocating here and failing to take ownership in the constructor. Also makes it more clean who owns it.

cppcache/src/TcrEndpoint.cpp
```diff
@@ -504,14 +504,15 @@ void TcrEndpoint::pingServer(ThinClientPoolDM* poolDM) {
...
+                           true);
```
pivotal-jbarrett, line 508: If this value is always `true` then let's just roll it into the constructor initializer and remove from parameters.

pivotal-jbarrett (changes requested):

pivotal-jbarrett (approved):
YES!

mmartell (approved):

PR #797, opened 2021-04-29
GEODE-9214: Remove ACE time library references

gaussianrecurrence:
- Replaced all references to ACE time library by boost/STL functions.
 - Also replaced old ITs logging function implementation so it doesn't
   use ACE.


pivotal-jbarrett:
@gaussianrecurrence you are on a roll at getting ACE out of here! Thanks!

tests/cpp/fwklib/FwkLog.cpp
```diff
@@ -54,32 +55,24 @@ const char* dirAndFile(const char* str) {
...
+      duration_cast<microseconds>(now.time_since_epoch()).count() % 1000;
 
-  fprintf(stdout, "[%s %s %s:P%d:T%" PRIu64 "]::%s::%d  %s  %s\n", buf,
-          u.sysname, u.nodename, ACE_OS::getpid(),
-          hacks::aceThreadId(ACE_OS::thr_self()), fil, lineno, l, s);
-  fflush(stdout);
+  const char* fil = dirAndFile(filename);
+  std::cout << '[' << std::put_time(localtime, "%Y/%m/%d %H:%M:%S") << '.'
+            << std::setfill('0') << std::setw(6) << usec << std::setw(0)
+            << std::put_time(localtime, " %Z ") << boost::asio::ip::host_name()
```
pivotal-jbarrett, line 71: If it simplifies anything to drop the hostname, change the formatting, and/or switch to boost::log like the new test framework I don't think it will matter any.

The old framework used to support running the child processes remotely, so logging the hostnames was useful in debugging. It can't run remotely anymore so hostname Is not so useful.

I don't think anyone is tied to the log format either.

tests/cpp/fwklib/FwkLog.cpp
```diff
@@ -54,32 +55,24 @@ const char* dirAndFile(const char* str) {
 }
 
 void plog(const char* l, const char* s, const char* filename, int32_t lineno) {
```
pivotal-jbarrett, line 57: If it simlifies things, the only callers to this method are macros that are converting a `std::ostringstream` to a `char *` so you could just pass a `std::ostringstream &` or `const std::string &` here.

```c++
void plog(const std::string& level,  const std::string& message, const std::string& filename, size_t lineNumber);
```

tests/cpp/fwklib/FwkLog.cpp
gaussianrecurrence: I think I'll go for boost log. Thanks for pointing it out.

cppcache/integration-test/fw_dunit.cpp
```diff
@@ -688,36 +680,63 @@ void sleep(int millis) {
...
 // log a message and print the worker id as well.. used by fw_helper with no
 // worker id.
 void log(std::string s, int lineno, const char * /*filename*/, int /*id*/) {
-  char buf[128] = {0};
-  dunit::getTimeStr(buf, sizeof(buf));
-
-  fprintf(stdout, "[TEST 0:pid(%d)] %s at line: %d\n",
-          boost::this_process::get_id(), s.c_str(), lineno);
-  fflush(stdout);
+  using std::chrono::duration_cast;
```
pivotal-jbarrett, line 703: With the transition to boost::log for the test framework, what is still using this logger?

gaussianrecurrence: Sadly I only changed tests/cpp/security code to log with Boost.Log
Thing is in theory this PR is supposed to be focused remove references for ACE time library.
Also, I tried to replace logging for Boost.Log on the old ITs, and the changes were huge.
So what do you think if a open a separate PR to adress this?

pivotal-jbarrett: No. If it's big change it isn't worth it. It was only worth it if it made refactoring this crusty log code easier.

pivotal-jbarrett (approved):

PR #802, opened 2021-05-05
GEODE-9236: Fix testThinClientFailover2 and reenable

mmartell:
This test is important because it validates CACHING_PROXY with client notifications enabled for multiple client caches.


pdxcodemonkey (approved):
We'll keep an eye on this one in CI for consistency.  It's all green now, so we have a clean slate.  More tests are always better than fewer!

PR #804, opened 2021-05-10
GEODE-9249: remove(key, value) bug

mmartell:
Fixes TcrMessageDestroy when value is a nullptr.


pivotal-jbarrett (changes requested):

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,114 @@
...
+  bool resultRemove = region->remove(key, value);
+  ASSERT_FALSE(resultRemove);
+
+  // Can remove by nullptr value
+  resultRemove =
+      region->remove(key, static_cast<std::shared_ptr<Cacheable>>(nullptr));
+  ASSERT_TRUE(resultRemove);
+
+  //**********************************************************************
+  // Test remove after localInvalidate
```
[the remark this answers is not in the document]
mmartell: Good catch. Broke into three separate tests.

mreddington (approved):

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+  auto region = cache.createRegionFactory(RegionShortcut::CACHING_PROXY)
+                    .setPoolName("pool")
+                    .create("region");
+
+  region->put(key, value);
+
+  region->localInvalidate(key);
+
+  bool resultRemove =
```
pivotal-jbarrett, line 132: `auto`?

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+      region->remove(key, static_cast<std::shared_ptr<Cacheable>>(nullptr));
```
pivotal-jbarrett, line 133: We don't have a specialization for `nullptr`? We should probably add that.

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+      .addLocator("localhost", cluster.getLocatorPort())
+      .create("pool");
+
+  auto region = cache.createRegionFactory(RegionShortcut::CACHING_PROXY)
+                    .setPoolName("pool")
+                    .create("region");
+
+  region->put(key, value);
+
+  bool resultRemove = region->remove(key);
```
pivotal-jbarrett, line 156: `auto`?

cppcache/src/ThinClientRegion.cpp
```diff
@@ -1066,8 +1066,8 @@ GfErrType ThinClientRegion::removeNoThrow_remote(
 
   // do TCR remove
   TcrMessageDestroy request(new DataOutput(m_cacheImpl->createDataOutput()),
-                            this, keyPtr, cvalue, aCallbackArgument,
-                            m_tcrdm.get());
+                            this, keyPtr, cvalue, cvalue == nullptr,
```
pivotal-jbarrett, line 1069: Is it necessary to put this parameter here since we are sending the value as the parameter just before it? Internally we can check for `nullptr` rather than putting it on the caller.

pivotal-jbarrett (changes requested):

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+  cache.getPoolManager()
+      .createFactory()
+      .addLocator("localhost", cluster.getLocatorPort())
+      .create("pool");
+
+  auto region = cache.createRegionFactory(RegionShortcut::CACHING_PROXY)
+                    .setPoolName("pool")
+                    .create("region");
+
```
pivotal-jbarrett, line 74: All of this looks duplicated in each test. You could put this in the `SetUp` method of the test class.

Same for the `cache.close()` at the end.

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+  region->put(key, value);
+
+  auto user = region->get(key);
+
+  region->localInvalidate(key);
+
+  bool resultLocalRemove = region->localRemove(key, value);
```
pivotal-jbarrett, line 81: `auto`

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+  auto region = cache.createRegionFactory(RegionShortcut::CACHING_PROXY)
+                    .setPoolName("pool")
+                    .create("region");
+
+  region->put(key, value);
+
+  region->invalidate(key);
+
+  bool resultRemove = region->remove(key, value);
```
pivotal-jbarrett, line 106: `auto`

pivotal-jbarrett (changes requested):

cppcache/integration/test/CachingProxyTest.cpp
mmartell: Done.

cppcache/integration/test/CachingProxyTest.cpp
mmartell: Nice idea. Added a function for this in CacheableBuiltins.hpp so don't need to cast:
     region->remove(key, nullptr);

cppcache/integration/test/CachingProxyTest.cpp
mmartell: Done

cppcache/src/ThinClientRegion.cpp
mmartell: Since the region->remove() and region->destroy() APIs all use TcrMessageDestroy, there's actually two types of Destroy messages that can be sent to the server. This parameter is needed to send the proper one based on whether the user is removing a key or a key and a value.

cppcache/integration/test/CachingProxyTest.cpp
mmartell: Good catch.

cppcache/integration/test/CachingProxyTest.cpp
mmartell: Done

cppcache/integration/test/CachingProxyTest.cpp
mmartell: Done

cppcache/src/ThinClientRegion.cpp
pivotal-jbarrett: Gotcha! Confusing...

`shared_ptr` converts to boolean, where `nullptr` converts to `false` and all other values to `true`. so `!cvalue` == `cvlaue == nullptr`

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -51,86 +53,67 @@ class CachingProxyTest : public ::testing::Test {
...
 
   region->localInvalidate(key);
 
-  bool resultLocalRemove = region->localRemove(key, value);
+  auto resultLocalRemove = region->localRemove(key, value);
   ASSERT_FALSE(resultLocalRemove);
 
   resultLocalRemove = region->localRemove(
-      key, static_cast<std::shared_ptr<Cacheable>>(nullptr));
+      // key, static_cast<std::shared_ptr<Cacheable>>(nullptr));
```
pivotal-jbarrett, line 87: Delete commented out code and format. Repeated in a few other places.

pivotal-jbarrett (changes requested):

pivotal-jbarrett (changes requested):

cppcache/src/ThinClientRegion.cpp
mmartell: Yup. C++11 allows using cvalue == nullptr, or !cvalue.

cppcache/integration/test/CachingProxyTest.cpp
mmartell: Good catch. Thorough, that jbarrett is.

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -51,86 +53,67 @@ class CachingProxyTest : public ::testing::Test {
...
+
+    region = cache.createRegionFactory(RegionShortcut::CACHING_PROXY)
+                 .setPoolName("pool")
+                 .create("region");
+  }
 
   void TearDown() override {}
 
   Cluster cluster = Cluster{LocatorCount{1}, ServerCount{1}};
+  Cache cache = CacheFactory().create();
```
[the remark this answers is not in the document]
mmartell: Correct. The default CTOR is a deleted function.

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -51,86 +53,67 @@ class CachingProxyTest : public ::testing::Test {
...
+        .createFactory()
+        .addLocator("localhost", cluster.getLocatorPort())
+        .create("pool");
+
+    region = cache.createRegionFactory(RegionShortcut::CACHING_PROXY)
+                 .setPoolName("pool")
+                 .create("region");
+  }
 
   void TearDown() override {}
```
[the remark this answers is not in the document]
mmartell: Good catch.

cppcache/integration/test/CachingProxyTest.cpp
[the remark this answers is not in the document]
pivotal-jbarrett: Not sure what feels dirtier. Calling a builder pattern in the initializer for a member variable or converting that variable to `std::shared_ptr` and initializing it in the `SetUp`.

We should probably improve the `Cache` to default construct to an effectively "closed" cache so you can have it declared as a member but initialized to something useful later via move.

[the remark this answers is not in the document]
mmartell: That feels a bit contrived. Maybe we should just follow the pattern in the other new framework tests, which setup server and cache in each test.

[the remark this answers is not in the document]
mreddington: Seconded on wanting an explanation; nothing seems dirty here, this functionality was introduced in C++11 (though I personally don't care for it AT ALL, I can't find fault with it so I won't otherwise contest it).

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,121 @@
...
+using apache::geode::client::Cacheable;
+using apache::geode::client::CacheableString;
+using apache::geode::client::CacheFactory;
+using apache::geode::client::Region;
+using apache::geode::client::RegionShortcut;
+using apache::geode::client::Serializable;
+
+namespace CachingProxy {
+
+class CachingProxyTest : public ::testing::Test {
```
mreddington, line 41: You're going to want something like this:

```
protected:
static std::unique_ptr<Cluster> cluster_;
static std::unique_ptr<Cache> cache_;
static std::shared_ptr<Region> region_;

public:
static void SetUpTestSuite() {
  cluster_ = make_unique<Cluster>(Name{"CachingProxyTest"}, LocatorCount{1}, ServerCount{1});

  cluster_->start();

  cluster.getGfsh()
         .create()
         .region()
         .withName("region")
         .withType("PARTITION")
         .execute();

  cache_ = make_unique<Cache>(cluster_->createCache());

  cache_.getPoolManager()
         .createFactory()
         .addLocator("localhost", cluster_.getLocatorPort())
         .create("pool");

  region_ = cache_->createRegionFactory(CACHING_PROXY)
                     .setPoolName("pool")
                     .create("region");
  }

  static void TearDownTestSuite() {
    region_.reset();
    cache_.reset();
    cluster_.reset();
  }
```

Now your cluster is running and configured, your cache is configured, and your region is configured. All will be shared between every test scenario. It will be setup before any scenario is run, it is shutdown after all the scenarios are completed.

And don't forget with statics, you need to define their initial state outside the class declaration. Since they're all smart pointers, this is easy:

```
}; // class CachingProxyTest

std::unique_ptr<Cluster> CachingProxyTest::cluster_{};
std::unique_ptr<Cache> CachingProxyTest::cache_{};
std::shared_ptr<Region> CachingProxyTest::region_{};
```

You can even do this with your test key and test value, and I would recommend you include the test region and test pool names. Don't rely on duplicating literals.

You then only need to be sure to insert the initial value in the constructor, and clear the region in the destructor in preparation for the next test. You'll also be able to get rid of `SetUp` and `TearDown`, which (isn't always the case, but) would be redundant to the ctor and dtor.

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,121 @@
...
+  void TearDown() override { cache.close(); }
+
+  Cluster cluster = Cluster{LocatorCount{1}, ServerCount{1}};
+  Cache cache = CacheFactory().create();
+  std::string key = std::string("scharles");
+  std::string value = std::string("Sylvia Charles");
+  std::shared_ptr<Region> region;
+};
+
+TEST_F(CachingProxyTest, LocalRemoveAfterLocalInvalidate) {
```
mreddington, line 76: With your test harness configuring once, that makes each test fast and cheap. That means you can write more tests. I would rather see:

* zero setup in the test
* all your expectations
* ONE ACTION
* all your assertions

Your tests contain multiple actions and multiple assertions. This is a habit that is borne out of tests being expensive, so you're trying to avoid costs by cramming as much into one as possible. Now you don't have to do that.

cppcache/integration/test/CachingProxyTest.cpp
```diff
@@ -0,0 +1,121 @@
...
+  region->put(key, value);
+
+  auto user = region->get(key);
+
+  region->localInvalidate(key);
+
+  auto resultLocalRemove = region->localRemove(key, value);
+  ASSERT_FALSE(resultLocalRemove);
```
mreddington, line 84: See? This is the confusing bit. Why are you asserting this? The line prior is test setup code. THAT IT WORKS is the foregone conclusion. The real test is the two lines below, that's it. So everything above goes into the test harness constructor.

I recommend, after my learning experience in PR#805, that you keep the test harness hierarchy flat. I'd rather you duplicate the code and put all your configuration in the ctor so it's all in one place, easier to find, easier to read, and easier to understand. You can make a `CachingProxyWithLocallyInvalidatedKey` harness, and the test case can then be named `LocalRemove`.

mreddington (changes requested):

cppcache/integration/test/CachingProxyTest.cpp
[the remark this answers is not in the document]
pivotal-jbarrett: @mreddington Two issues:

1. We are initializing test specific objects in the fixture's, these will not be per-test method instances of these objects.
2. We are performing complex and exception throwing initialization of a member variable in the initialization of a type before the constructor is invoked. This particular factory can take variable parameters, like connection time outs, which could not be forwarded from some constructor or member method on this type. Initialization of a variable should never be done like this in production code so it also shouldn't in test code.

What this attempt to use the test fixture setup method has shown us it that we have a deficiency in how we defined the construction of `Cache` in that you can't declare it without also initializing it with some complex builder. Like many other state holding types in C++11 it should default construct to some safe but unusable state for its declaration, like `std::shared_ptr` default initializes to `nullptr`. It should then allow rvalue assignment from the value returned from the factory so that it can be fully initialized in the `SetUp` method. Similar to how someone with a `Application` class would declare it as a member and then initialize it someone in the application lifecycle.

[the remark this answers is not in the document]
pivotal-jbarrett: Based on the complexity of this change though @mmartell and I discussed holding off on using a test fixture to initialize these common values and copy the pattern of the other tests. We are then going to explore changes to `Cache` default construction and move assignment under another ticket.

cppcache/integration/test/CachingProxyTest.cpp
pivotal-jbarrett: This is not a pattern you should ever use for testing.

 `Cache`, which is the root of all the components you are integration testing here, should NEVER have shared state between tests. It is too easy to end up writing tests that have indeterminate outcomes based on previous test. Did the test pass because the previous test put it a state it will always pass? Will the order of my tests change the behavior of each test?

Sharing the external system, `Cluster` in this case, is arguably ok to share within a test fixture. I lean very heavily in not for the same reasons above. The previous test framework did this and many tests were inadvertently left in strange states because of it.

cppcache/integration/test/CachingProxyTest.cpp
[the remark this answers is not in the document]
mmartell: Switched to use TEST instead of TEST_F.

cppcache/integration/test/CachingProxyTest.cpp
mmartell: I have no problem avoiding this pattern for the Cache. However, I think it's worth considering this pattern for the server, which is the expensive part. Also, we're not testing the server here.

pivotal-jbarrett: For moving the `Cluster` initialization to the fixture level to have any advantage you would need also modify the test discovery and execution model in CMake such that those that had a shared `Cluster` were executed serially in the same processes space. Currently all tests are executed in their own process space so any shared effort in the fixture setup is effectively test level setup.

pivotal-jbarrett (approved):

mreddington (approved):

PR #805, opened 2021-05-12
GEODE-9110: test thin client after region live

mreddington:
We had a test porting sprint to get as much done as possible.


cppcache/integration/test/ThinClientHARegionCacheListenerTest.cpp
```diff
@@ -0,0 +1,426 @@
...
+::std::unique_ptr<T> make_unique(Args &&... args) {
+  return ::std::unique_ptr<T>(new T(::std::forward<Args>(args)...));
+}
+
+// A simple comparator for cachables wrapped in shared pointers.
+MATCHER_P(CashableEq, value, "") {
+  return arg->toString() == value->toString();
+}
+
+class ThinClientHARegionCacheListenerTest : public ::testing::Test {
```
pivotal-jbarrett, line 55: I think we can drop the "ThinClient" prefix from all these tests in the future. All tests are for the "thin client" The original C++ was intended to be a peer "heavy client".

cppcache/integration/framework/Cluster.h
```diff
@@ -127,6 +127,8 @@ using ConserveSockets = NamedType<bool, struct useConserveSocketsParameter>;
 
 class Cluster {
  public:
+  enum class Subscription_State { Enabled, Disabled };
```
pivotal-jbarrett, line 130: This doesn't match the naming conventions used in the framework.
I love the introduction of a named type to replace a general purpose type!!!

pivotal-jbarrett (changes requested):

cppcache/integration/framework/Cluster.h
mreddington: Thy will be done.

cppcache/integration/test/ThinClientHARegionCacheListenerTest.cpp
mreddington: Also will do.

cppcache/integration/test/HARegionCacheListenerARLEPDTest.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <geode/RegionEvent.hpp>
+#include <geode/RegionFactory.hpp>
+#include <geode/RegionShortcut.hpp>
+
+#include "mock/CacheListenerMock.hpp"
+
+namespace {
+// Until we get C++14 support...
+template <typename T, typename... Args>
+::std::unique_ptr<T> make_unique(Args &&... args) {
```
pivotal-jbarrett, line 46: Why not just make this a util and start using it all over? ;)

pivotal-jbarrett (approved):
I would just say move that `make_unique` into a util header in the main sources and use it everywhere.

cppcache/integration/test/HARegionCacheListenerARLEPDTest.cpp
```diff
@@ -0,0 +1,231 @@
...
+  return ::std::unique_ptr<T>(new T(::std::forward<Args>(args)...));
+}
+
+// A simple comparator for cachables wrapped in shared pointers.
+// MATCHER_P(CashableEq, value, "") {
+//   return arg->toString() == value->toString();
+// }
+
+class HARegionCacheListenerARLEPDTest : public ::testing::Test {
```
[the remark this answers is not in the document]
mreddington: AfterRegionLiveEndPointDisconnected. As you can see, I haven't figured out how to get a grip on naming tests. The test cases are beautifully simple and easy to understand, but to retain context of just what the preconditions are being tested against, the names are getting huge. I'm very open to suggestions because I don't like what's going on, either.

cppcache/integration/test/HARegionCacheListenerARLEPDTest.cpp
mreddington: Because at the time, I couldn't be bothered, I was more focused on figuring out just how not-trivial cache listener tests were. And then we ran out of time for the sprint and I was trying to get a PR posted. I'll move it into a utility.

cppcache/integration/test/HARegionCacheListenerARLEPDTest.cpp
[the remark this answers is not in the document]
pivotal-jbarrett: Yikes yeah that's a screenful. Perhaps just a class level comment doc with the full intent.

pivotal-jbarrett (approved):

PR #807, opened 2021-05-14
GEODE-9259: cleanup and reenable testThinClientRegionOps

mmartell:
This is an important test because it is a quite extensive test of the Region API. It has not been run in a very long time, because it failed. With the recent fix to TcrMessageDestroy (PR #804), this test now passes.


pdxcodemonkey (approved):
Looks good to me, and gates are passing.  Ship it!

PR #808, opened 2021-05-21
Bump puma to 5.3.1 in docs

moleske:
I have GitHub security alerts turned on for my fork, and was alerted to these bumps.  It is security fix.  Haven't got docker running at the moment so haven't tested that nothing breaks (though would be surprised to find out that's the case since  it is not a major version bump)


moleske:
closing this as https://github.com/apache/geode/pull/6508#issuecomment-847998578 points out it is being worked on.  Then someone can just copy pasta the changes to native

PR #810, opened 2021-05-27
GEODE-9322: Fix possible race condition on new ITs

gaussianrecurrence:
- Solved a potential race condition in TransactionCleaningTest TS
   due to the fact that no checks were performed after starting/stopping
   the server.

### DIFF
--- a/cppcache/integration/test/TransactionCleaningTest.cpp
+++ b/cppcache/integration/test/TransactionCleaningTest.cpp
@@ -17,25 +17,55 @@
 
 #include <gmock/gmock.h>
 
-#include <future>
-#include <thread>
-
 #include <gtest/gtest.h>
 
 #include <geode/CacheTransactionManager.hpp>
+#include <geode/EntryEvent.hpp>
+#include <geode/RegionEvent.hpp>
 #include <geode/RegionFactory.hpp>
 #include <geode/RegionShortcut.hpp>
 
 #include "framework/Cluster.h"
+#include "gmock_actions.hpp"
+#include "mock/CacheListenerMock.hpp"
+#include "util/concurrent/binary_semaphore.hpp"
 
 namespace {
 
+using apache::geode::client::binary_semaphore;
+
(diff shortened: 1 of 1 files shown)


gaussianrecurrence:
If @pdxcodemonkey @pivotal-jbarrett @mreddington @alb3rtobr @mkevo @mivanac @jvarenina could take a look at this, I'd appreciate it quite much!

Also note that PR #806 depends on this PR, given the fixed IT fails pretty frequently.

pdxcodemonkey (approved):
Looks good - though I don't fully understand the nature of the race condition you mention, I trust you've been over the details and fixed.

Quick question - this part of your description: "Modified RegisterKeysTest to use gmock_actions helper."  I see no changes to RegisterKeysTest.  Are they missing, or is your PR description just out of date?

gaussianrecurrence:
> Looks good - though I don't fully understand the nature of the race condition you mention, I trust you've been over the details and fixed.

Regarding this, basically it solves race condition in which the test proceeds before the server is actually stopped. Note that gfsh command "stop server" is not synchronous, so after its execution the server could be still up.

> Quick question - this part of your description: "Modified RegisterKeysTest to use gmock_actions helper." I see no changes to RegisterKeysTest. Are they missing, or is your PR description just out of date?

Thanks for noticing this. As it seems to me, I made this fix along with PR #809, so in order to make the changes work there was some duplicity on both PRs, as #809 was merged I guess the delta doesn't show the RegisterKeysTest anymore. I'll update the description.

Just a comment: I would like to improve the readability for the code in general and ITs in particular. So do you think it is possible to add a comment on ITs with a short description of what it does, as well as what's the test for?

pdxcodemonkey:
I agree it'd probably be a great idea to document the tests intent in Doxygen comments or similar.  Perhaps bring this up on the dev list or just file a JIRA ticket.  I don't think we wish to bother with the "old" tests, but for sure everything in the new suite should be properly understood.

PR #817, opened 2021-06-02
Update clang tools version in CI

pdxcodemonkey:
Homebrew formulae for MacOS  installs v12 of clang tools by default now, and once again it is not compatible with the previous version.  This installs v12 on our Docker image, so we don't have to jump through more hoops for Mac development.

### DIFF
--- a/ci/docker/clang-tools/Dockerfile
+++ b/ci/docker/clang-tools/Dockerfile
@@ -18,7 +18,7 @@ FROM ubuntu:groovy
 LABEL maintainer="Apache Geode <dev@geode.apache.org>"
 LABEL description="Minimal image for building with clang toolset."
 
-ARG CLANG_VERSION=11
+ARG CLANG_VERSION=12
 RUN apt-get update \
     && apt-get install -y \
         bash \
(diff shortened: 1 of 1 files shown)


pivotal-jbarrett:
> Homebrew formulae for MacOS installs v12 of clang tools by default now, and once again it is not compatible with the previous version. This installs v12 on our Docker image, so we don't have to jump through more hoops for Mac development.

What about the Windows peeps?

pdxcodemonkey:
clang-format 11 on Windows isn't even compatible with *the same version* on other platforms, so I'm not sure how valuable it is to cater to Windows here.  I've had to switch platforms and re-run the tool to fix bad Windows formatting multiple times.

pivotal-jbarrett:
> clang-format 11 on Windows isn't even compatible with _the same version_ on other platforms, so I'm not sure how valuable it is to cater to Windows here. I've had to switch platforms and re-run the tool to fix bad Windows formatting multiple times.

I am good with that answer!

pivotal-jbarrett (approved):

pivotal-jbarrett:
Word of warning, the CI won't use this version to validate the current sources in this PR's checks. It is possible merging will break CI if the current sources are not correct to this `clang-format` or `clang-tidy`. Be prepared to post a PR to correct the sources too.

PR #821, opened 2021-06-15
GEODE-9358: Initial revision of C bindings

pdxcodemonkey:
This is the initial work to create a C API for `geode-native`.  The goal of this work is _not_ to enable access to `geode-native` from the C language, though that will happen as a matter of course, but rather to enable access from all the other languages (.net core, Python, Go, etc.) from which access to a C API is simple and straightforward.

[RFC](https://cwiki.apache.org/confluence/display/GEODE/Add+C+Bindings+to+Native+Client+Library)

### DIFF
--- a/c_bindings/src/region.cpp
+++ b/c_bindings/src/region.cpp
@@ -36,12 +36,12 @@
 #include "region.hpp"
 #include "region/factory.hpp"
 
-RegionWrapper::RegionWrapper(RegionFactoryWrapper &region_factory, std::shared_ptr<apache::geode::client::Region> region)
-    : region_(region), region_factory{region_factory} {
-      region_factory.AddRecord(this, "RegionWrapper");
+RegionWrapper::RegionWrapper(std::shared_ptr<apache::geode::client::Region> region)
+    : region_(region) {
+      AddRecord(this, "RegionWrapper");
     }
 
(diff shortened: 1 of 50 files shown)


mreddington (approved):

mmartell (approved):
Awesome addition to geode-native. With C's stable ABI, seems like the right way to open the door for other languages.

c_bindings/src/region.cpp
```diff
@@ -0,0 +1,134 @@
...
+      std::dynamic_pointer_cast<apache::geode::client::CacheableBytes>(val);
+  int valSize = val->objectSize();
+#if defined(_WIN32)
+  int8_t* byteArray = static_cast<int8_t*>(CoTaskMemAlloc(valSize));
+#else
+  int8_t* byteArray = static_cast<int8_t*>(malloc(valSize));
+#endif
+  if (bytes) {
+    memcpy(byteArray, bytes->value().data(), valSize);
+    *value = (char*)byteArray;
```
mreddington, line 82: There's your clang-tidy problem, use a reinterpret_cast.

pdxcodemonkey: Got it, thx.

echobravopapa (approved):
👍 👍

PR #822, opened 2021-06-21
GEODE-9387: Ignore version trace in gnmsg

pdxcodemonkey:
- makes the tool robust against weird versions of NC
- no longer any need to combine rolled log files to parse messages

Also fixed several issues I noticed in timestamps:
- chunked server responses didn't have timestamps
- some timestamps were strings, others datetime objects, so they printed out differently ('/' vs '-')
- gnmsg wasn't parsing out the date, only the timestamp, so output had today's date rather than the date in the log file

@todmorrison

### DIFF
--- a/tools/gnmsg/chunked_message_decoder.py
+++ b/tools/gnmsg/chunked_message_decoder.py
@@ -21,24 +21,33 @@ from read_values import read_int_value, read_byte_value, call_reader_function
 
 CHUNKED_MESSAGE_HEADER_LENGTH = 17
 
+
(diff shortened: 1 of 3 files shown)


mreddington (commented):
You know, I wanted to comment about some of these regular expressions, trying to make them more legible and clearer with intent, but there's just no way to improve it. \d\d\d\d\ is as bad as \d{4} which is as bad as a four_digit_year + '/' + two_digit_month + '/' +...

mreddington (approved):

PR #825, opened 2021-06-25
GEODE-9401: fix duplicate defines mac

mmartell:
Recent changes to system headers causes compile errors on OSX due to variable names that collide with these predefined macros:

BYTE_SIZE
PAGE_SIZE

This PR fixes the issue by renaming the variables, and others in the same file for consistency.

### DIFF
--- a/clicache/src/impl/PdxFieldType.cpp
+++ b/clicache/src/impl/PdxFieldType.cpp
@@ -166,23 +166,23 @@ namespace Apache
           switch (m_typeId)
           {
           case PdxFieldTypes::BYTE:
-            return PdxTypes::PDX_BYTE_SIZE;
+            return PdxTypes::kPdxByteSize;
           case PdxFieldTypes::BOOLEAN:
-            return PdxTypes::PDX_BOOLEAN_SIZE;
+            return PdxTypes::kPdxBooleanSize;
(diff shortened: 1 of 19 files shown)


clicache/src/impl/PdxFieldType.cpp
```diff
@@ -166,23 +166,23 @@ namespace Apache
           switch (m_typeId)
           {
           case PdxFieldTypes::BYTE:
-            return PdxTypes::BYTE_SIZE;
+            return PdxTypes::PDX_BYTE_SIZE;
```
pivotal-jbarrett, line 169: If you are changing these then we should be changing them to match the style guide we currently have approved. In our last discussion about the style guide we did not agree on a deviation for enum names. By using prescribed naming convention you will avoid the macro name collision.

https://google.github.io/styleguide/cppguide.html#Enumerator_Names

pivotal-jbarrett (changes requested):

clicache/src/impl/PdxFieldType.cpp
mmartell: Done. Nice catch!

pivotal-jbarrett (approved):

PR #826, opened 2021-06-29
GEODE-9405: Remove broken security tests

mmartell:
Some of the legacy C++ tests use strncpy. This causes warnings in gcc when the length parameter depends on the length of the source parameter. To avoid such warnings, which we treat as errors, we can use memcpy in the test code where we know there is no buffer overrun issue.


pdxcodemonkey (changes requested):
This code really bothers me.  Ordinarily I'm a big fan of the most minimal change, but in this case can we rewrite this just using `std::string` throughout?  And while we're at it, maybe stuff one copy of this function in a `utility.cpp` somewhere, and delete the other (gulp) 7 copies?

cppcache/integration-test/CMakeLists.txt
```diff
@@ -205,7 +205,7 @@ set_tests_properties(
     testThinClientPoolAttrTest
     testThinClientPoolLocator
     testThinClientPoolRedundancy
-    testThinClientSecurityAuthentication
+    #testThinClientSecurityAuthentication
```
pivotal-jbarrett, line 208: Delete?

cppcache/integration-test/ThinClientSecurityHelper.hpp
```diff
@@ -59,13 +59,15 @@ const char* regionNamesAuth[] = {"DistRegionAck"};
 std::shared_ptr<CredentialGenerator> credentialGeneratorHandler;
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char* path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
+
+  int indexOfCppcache = path.find("cppcache");
```
pivotal-jbarrett, line 64: We use boost::filesystem elsewhere so using it here to make file path operations safe would be nice.

cppcache/integration-test/testThinClientSecurityAuthentication.cpp
```diff
@@ -138,7 +140,7 @@ DUNIT_TASK_DEFINITION(LOCATORSERVER, CreateServer1)
         printf("Input to server cmd is -->  %s",
                cmdServerAuthenticator.c_str());
         CacheHelper::initServer(
-            1, nullptr, locHostPort,
+            1, "cacheserver_notify_subscription.xml", locHostPort,
```
pivotal-jbarrett, line 143: Wouldn't this change the intent of the test now that it starts with a cache xml file?

cppcache/integration-test/testThinClientSecurityAuthentication.cpp
```diff
@@ -138,7 +140,7 @@ DUNIT_TASK_DEFINITION(LOCATORSERVER, CreateServer1)
...
             const_cast<char *>(cmdServerAuthenticator.c_str()));
```
pivotal-jbarrett, line 144: This should be fixed too. Really bad to be taking stripping the `const` off here. The funny thing is the method takes a `const`. The method should really be change to take a `const std::string&`. I know this might feel out of scope but it's right there!

cppcache/integration-test/testThinClientSecurityAuthorization.cpp
```diff
@@ -54,13 +54,15 @@ const std::string locHostPort =
 std::shared_ptr<CredentialGenerator> credentialGeneratorHandler;
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char *path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
pivotal-jbarrett, line 57: Use `auto` on the lefthand side whenever you can. Almost all the new code should be corrected to use auto for all these local variables.

pivotal-jbarrett (changes requested):

cppcache/integration-test/CMakeLists.txt
mmartell: Actually forgot to uncomment this. I was trying to see if fixing getXmlPath() would fix this test. It doesn't, so I'll disable this test again.

mmartell:
Converting to Draft for now. All but one of the files that fail to compile on rhel-8 due to strncpy warnings are disabled tests that use a bogus getXmlPath() function. This function not only uses an unsafe strncpy(), but uses an incorrect path to the Xml folder. Looks like this is an artifact of moving the Xml folder in native client 10, but not fixing up all the tests to use the new path.

Rather than just fix the compile problem for tests that we never run, I think it is worth investigating whether we can reenable all these tests by using the correct path. Of course we will also fix the compile problem that's causing the build failure on rhel-8.

The list of tests that experience this issue includes:

testThinClientSecurityAuthentication.cpp (broken)
testThinClientSecurityAuthenticationSetAuthInitialize.cpp
testThinClientSecurityAuthorization.cpp (broken)
testThinClientSecurityAuthorizationMU.cpp (broken)
testThinClientSecurityCQAuthorizationMU.cpp (flaky)
testThinClientSecurityDurableCQAuthorizationMU.cpp (broken)
testThinClientWriterException.cpp (broken)
ThinClientSecurityHelper.hpp (not a test)

cppcache/integration-test/ThinClientSecurityHelper.hpp
```diff
@@ -59,13 +59,15 @@ const char* regionNamesAuth[] = {"DistRegionAck"};
 std::shared_ptr<CredentialGenerator> credentialGeneratorHandler;
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char* path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
gaussianrecurrence, line 62: Be careful when assigning the output of std::getenv to an std::string. If the env var is not defined it will return a nullptr, and passing a nullptr to a std::string constructor result in a coredump. There is an utility function Utils::getEnv which acts safely by returning an empty string when the variable is not defined.

cppcache/integration-test/testThinClientSecurityAuthentication.cpp
```diff
@@ -34,13 +34,15 @@ const char *regionNamesAuth[] = {"DistRegionAck", "DistRegionNoAck"};
 std::shared_ptr<CredentialGenerator> credentialGeneratorHandler;
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char *path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
gaussianrecurrence, line 37: Same comment as above regarding code safety for std::getenv

cppcache/integration-test/testThinClientSecurityAuthenticationSetAuthInitialize.cpp
```diff
@@ -35,13 +35,15 @@ const char *regionNamesAuth[] = {"DistRegionAck", "DistRegionNoAck"};
 std::shared_ptr<CredentialGenerator> credentialGeneratorHandler;
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char *path = std::getenv("TESTSRC");
-  ASSERT(path != NULL,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
gaussianrecurrence, line 38: Same comment as above regarding code safety for std::getenv

cppcache/integration-test/testThinClientSecurityAuthorization.cpp
gaussianrecurrence, line 57: Same comment as above regarding code safety for std::getenv

cppcache/integration-test/testThinClientSecurityAuthorizationMU.cpp
```diff
@@ -65,13 +65,15 @@ std::shared_ptr<CredentialGenerator> credentialGeneratorHandler;
 const char *exFuncNameSendException = "executeFunction_SendException";
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char *path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
gaussianrecurrence, line 68: Same comment as above regarding code safety for std::getenv

cppcache/integration-test/testThinClientSecurityCQAuthorizationMU.cpp
```diff
@@ -138,13 +138,15 @@ class MyCqListener : public CqListener {
 };
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char *path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
gaussianrecurrence, line 141: Same comment as above regarding code safety for std::getenv

cppcache/integration-test/testThinClientSecurityDurableCQAuthorizationMU.cpp
```diff
@@ -139,13 +139,15 @@ class MyCqListener : public CqListener {
 };
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char *path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
gaussianrecurrence, line 142: Same comment as above regarding code safety for std::getenv

cppcache/integration-test/testThinClientWriterException.cpp
```diff
@@ -52,13 +52,15 @@ const char *regionNamesAuth[] = {"DistRegionAck"};
 std::shared_ptr<CredentialGenerator> credentialGeneratorHandler;
 
 std::string getXmlPath() {
-  char xmlPath[1000] = {'\0'};
-  const char *path = std::getenv("TESTSRC");
-  ASSERT(path != nullptr,
+  std::string path = std::string(std::getenv("TESTSRC"));
```
gaussianrecurrence, line 55: Same comment as above regarding code safety for std::getenv

cppcache/integration-test/ThinClientSecurityHelper.hpp
pivotal-jbarrett: Oh yes, good catch!!

cppcache/integration-test/testThinClientSecurityAuthentication.cpp
mmartell: N/A. Test deleted.

cppcache/integration-test/testThinClientSecurityAuthentication.cpp
mmartell: N/A. Test deleted.

cppcache/integration-test/testThinClientSecurityAuthorization.cpp
mmartell: N/A. Test deleted.

cppcache/integration-test/ThinClientSecurityHelper.hpp
mmartell: N/A. This function has been removed entirely, along with all the tests that use it, which were already disabled.

cppcache/integration-test/testThinClientSecurityAuthenticationSetAuthInitialize.cpp
mmartell: N/A. getXmlPath() is unnecessary in this test and has been removed. Note: The initServer function's 4th arg is unused.

cppcache/integration-test/testThinClientSecurityAuthorization.cpp
mmartell: N/A. Test removed.

cppcache/integration-test/testThinClientSecurityCQAuthorizationMU.cpp
mmartell: N/A. Test removed.

cppcache/integration-test/testThinClientWriterException.cpp
mmartell: N/A. Test removed.

mmartell:
This ticket originated from compile warnings on rhel-8 related to unsafe use of strncpy. In the course of fixing this, we discovered that all the tests that were using getXmlPath have been disabled for a very long time. These tests have now all been deleted.

cppcache/integration-test/testThinClientSecurityAuthentication.cpp
```diff
@@ -1,501 +0,0 @@
-/*
```
pivotal-jbarrett, line 1: Do we have replacements for this test?

pivotal-jbarrett (changes requested):
Are there replacements for these deleted tests? We need tests that test auth, multi-user auth, and cache writers. It is likely these tests never resolved all the library loading issues.

cppcache/integration-test/testThinClientSecurityAuthentication.cpp
mmartell: I feel our security test coverage is pretty weak. Suggest we create a single new security test that covers what's in these tests that isn't covered by new framework tests.

mmartell: We have only a very basic test in the new framework (SimpleAuthInitialize). Agree that we need a new test for auth (including exceptions handling), multi-user auth, and CacheWriter.

pivotal-jbarrett: Can you open a ticket to address these deleted test scenarios. While they weren't running they did stand as a place holder that we needed to fix. Deleted now we won't see them and we will forget about them.

mmartell: Good idea! Just created: https://issues.apache.org/jira/browse/GEODE-9414

pivotal-jbarrett (approved):

PR #829, opened 2021-07-02
GEODE-9412: Fixes translation of String between .NET and C++

pivotal-jbarrett:
Adds new String.hpp similar to C++ client's string.hpp.
Converts String to std::string via UTF-16 to UTF-8 transform.
Converts std::string to String via UTF-8 to UTF-16 transform.
Adds basic unit tests for String.hpp methods.

Adds basic unit tests for Serializable.

Splits the .NET library into objects and assembly so that unit tests
can link directly against the objects to avoid symbol export issues
crossing C++/CLI layers.

Fixes ACE PDB location so it can be correctly included in dependent projects.

### DIFF
--- a/clicache/src/CMakeLists.txt
+++ b/clicache/src/CMakeLists.txt
@@ -193,6 +193,7 @@ add_library(Apache.Geode-object OBJECT
   Struct.hpp
   StructSet.cpp
   StructSet.hpp
+  String.hpp
   SystemProperties.cpp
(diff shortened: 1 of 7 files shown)


pdxcodemonkey (approved):
Looks good - ship it.

pivotal-jbarrett:
> Looks good - ship it.

I would if the CI would go green.

PR #830, opened 2021-07-07
GEODE-8891: Move static regex in ThinClientRegion

pdxcodemonkey:
- Latest compiler/runtime on RHEL-8 causes an order-of-initialization
  bug for this static regex, and will core dump at app startup for
  optimized builds.
- Changing to local static ensures regex variable isn't initialized
  until the first time we use it, long after all static initializers are
  called.

### DIFF
--- a/cppcache/src/ThinClientRegion.cpp
+++ b/cppcache/src/ThinClientRegion.cpp
@@ -49,8 +49,14 @@ namespace apache {
 namespace geode {
 namespace client {
 
-static const std::regex PREDICATE_IS_FULL_QUERY_REGEX(
-    "^\\s*(?:select|import)\\b", std::regex::icase);
+class IsFullQueryRegex {
+ public:
+  static const std::regex& instance() {
+    static const std::regex predicateIsFullQueryRegex(
+        "^\\s*(?:select|import)\\b", std::regex::icase);
+    return predicateIsFullQueryRegex;
+  }
+};
 
 void setThreadLocalExceptionMessage(std::string exMsg);
 
@@ -589,7 +595,7 @@ std::shared_ptr<SelectResults> ThinClientRegion::query(
   }
 
   std::string squery;
-  if (std::regex_search(predicate, PREDICATE_IS_FULL_QUERY_REGEX)) {
+  if (std::regex_search(predicate, IsFullQueryRegex::instance())) {
     squery = predicate;
   } else {
     squery = "select distinct * from ";
--- a/cppcache/src/ThinClientRegion.cpp
+++ b/cppcache/src/ThinClientRegion.cpp
@@ -49,15 +49,6 @@ namespace apache {
 namespace geode {
 namespace client {
 
-class IsFullQueryRegex {
- public:
-  static const std::regex& instance() {
-    static const std::regex predicateIsFullQueryRegex(
-        "^\\s*(?:select|import)\\b", std::regex::icase);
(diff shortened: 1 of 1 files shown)


cppcache/src/ThinClientRegion.cpp
```diff
@@ -49,8 +49,14 @@ namespace apache {
 namespace geode {
 namespace client {
 
-static const std::regex PREDICATE_IS_FULL_QUERY_REGEX(
-    "^\\s*(?:select|import)\\b", std::regex::icase);
+class IsFullQueryRegex {
```
pivotal-jbarrett, line 52: I guess I would have just don it within the `ThinClientRegion` class and made the method `private` since there is no need for it to be public (well maybe for tests but there are none).

pivotal-jbarrett (approved):

cppcache/src/ThinClientRegion.cpp
```diff
@@ -589,7 +595,7 @@ std::shared_ptr<SelectResults> ThinClientRegion::query(
   }
 
   std::string squery;
-  if (std::regex_search(predicate, PREDICATE_IS_FULL_QUERY_REGEX)) {
+  if (std::regex_search(predicate, IsFullQueryRegex::instance())) {
```
mreddington, line 598: Why wrap the static const in a singleton? That sounds like an even more complicated solution. You could have just made this regex a local const variable. Who cares about runtime performance when this isn't where we're slow? The class still needs to be initialized at runtime anyway, as you've found out the hard way, and instead of burdening the whole test harness with the initialization of a global static, you could isolate that initialization to the instantiation of this test when it's ran.

mreddington (changes requested):

cppcache/src/ThinClientRegion.cpp
pivotal-jbarrett: I am not quite following this argument since it seems to argue for both options. The Meyers singleton won't initialize the static until first use so there is no burden to the test harness as a whole.

mmartell: If initializing a constant something is relatively expensive (i.e., a significant cost relative to its use), I'd prefer to initialize it only once, so I can amortize the cost of initialization.

In this case, I believe mreddington is advocating just declaring it locally inside of functions that use it, and not worry about the cost. Although calling the regex CTOR is likely a small percentage of the total test time, having to recreate a regex that's const seems silly.

If you declare the regex as a const static inside the function that uses it, then it will be initialized only once when the function is first used, thereby solving any cost concern associated with a regex CTOR. Also, C++11 guarantees that dynamic initialization of function-local statics is thread-safe, in case two threads call the function using the local regex at the same time.

So my recommendation is to simply move the original 2 lines of code inside the function that uses it.
```

mmartell (changes requested):
Although this works, I'd like to just use local variables whenever possible.

cppcache/src/ThinClientRegion.cpp
pivotal-jbarrett: Given this value doesn't have any use outside this method I could agree with making it a local const. There is technically no different between that and the proposed implementation other than encapsulation of the const logic in the additional method.

pivotal-jbarrett (approved):

pivotal-jbarrett (approved):

mmartell (approved):

mreddington (approved):

PR #832, opened 2021-07-20
GEODE-9438: NC user guides - Delete obsolete Auth properties

davebarnes97:
Deleted three obsolete system properties: 

- security-client-auth-factory
- security-client-auth-library
- security-client-dhalgo

Removed a redundant copy of the same table and updated the only link that pointed to it.


pdxcodemonkey (approved):
Looks good to me - thanks for getting rid of these!

PR #835, opened 2021-07-28
GEODE-9472: Relax the ratio of before and after threads.

mmartell:
The heuristic was a bit too strict. With 25 iterations, there was no reason to have the range be so small.

### DIFF
--- a/clicache/integration-test2/GarbageCollectCache.cs
+++ b/clicache/integration-test2/GarbageCollectCache.cs
@@ -85,7 +85,7 @@ namespace Apache.Geode.Client.IntegrationTests
                     {
                         string error = "ncThreadsBefore = " + ncThreadsBefore.ToString() +
                             ", ncThreadsAfter = " + ncThreadsAfter.ToString();
-                        Assert.False(!(.5 < ratio && ratio < 2.0), error);
+                        Assert.True((.5 < ratio && ratio < 2.0), error);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


clicache/integration-test2/GarbageCollectCache.cs
```diff
@@ -83,10 +83,9 @@ public void VerifyNoLeakedThreads()
                     // environment startup.
                     if (i > 5)
                     {
-                        //Assert.True(.8 < ratio && ratio < 1.3);
                         string error = "ncThreadsBefore = " + ncThreadsBefore.ToString() +
                             ", ncThreadsAfter = " + ncThreadsAfter.ToString();
-                        Assert.False(!(.8 < ratio && ratio < 1.3), error);
+                        Assert.False(!(.5 < ratio && ratio < 2.0), error);
```
pdxcodemonkey, line 88: Can we please make this a positive assert, rather than a negative?  It's very hard to read.  The commented-out line that you removed is MUCH clearer, while stating the same thing.  
```
Assert.True(.5 < ratio < 2.0);
```

Even changing it to two different assertions is easier to read, and potentially more helpful for debugging/diagnosis:

```
Assert.False(ratio < .5, error);
Assert.False(ratio > 2.0, error);
```

pdxcodemonkey (changes requested):

clicache/integration-test2/GarbageCollectCache.cs
mmartell: I put that in for debugging purposes. Assert.True wasn't giving me the output. Agree with your comment. Will change it back.

pdxcodemonkey (approved):

PR #841, opened 2021-08-14
GEODE-9501: Works with toolset Microsoft v142

mmartell:


mmartell:
This PR switches to latest Microsoft tools (i.e., compiler and linker). It uses cmake to decide which tools to use based on what's available on the build machine.

Notes:

I have tested locally on my Windows machine. Since I have Visual Studio 2017 and 2019 installed, I have to be careful when setting up the environment. Normally, I launch the developer command prompt that comes with the IDE to set the tools path.  Alternatively, I can select the desired toolchain from a normal command prompt by using the desired generator at cmake config time.

Since the CI uses a separate image for Visual Studio 2017 and 2019, using a developer command prompt isn't required. cmake will find the tools that are installed.

ci/base/base.yml
```diff
@@ -61,7 +61,7 @@ builds:
     with_dot_net: #@ True
     #@yaml/map-key-override
     params:
-      CMAKE_CONFIGURE_FLAGS: "-A x64 -Tv141,version=14.16,host=x64 -DCMAKE_SYSTEM_VERSION=10.0.16299.0"
```
pdxcodemonkey, line 64: I thought host=x64 was still necessary for VS2017 builds?

pdxcodemonkey (changes requested):

ci/base/base.yml
mmartell: I don't think we need this anymore. I'm able to build fine using VS 2017 without it. I've verified that my local build was using the 32 bit linker, and everything still builds and all tests pass. When I do a dumpbin /headers on the 32 bit linker, I see:
     "Application can handle large (>2GB) addresses"
Maybe this has been fixed.

mmartell: Good catch though. If this really is no longer needed we need to remove the check for it, which is a fatal error.

pdxcodemonkey: Okay, so some Googling tells me that for cmake 3.13 or newer, we need to ditch "Win64" in the -G parameter and use "-A x64" to configure 64-bit builds:  https://stackoverflow.com/questions/28350214/how-to-build-x86-and-or-x64-on-windows-from-command-line-with-cmake
And "-T host=x64" is used only to specifically force 64-bit toolset, so I don't suppose this is anything other than maybe a small performance hit.

pdxcodemonkey (approved):

PR #842, opened 2021-08-16
Revert "GEODE-9501: Works with toolset Microsoft v142 (#841)"

pdxcodemonkey:
This reverts commit 95c6b5ee53f9c8b7525fd34885cd9f3284d579d4.  Windows builds are breaking, and it looks like we should really be fixing the SDK version at a certain point and only updating intentionally, rather than by virtue of picking up whatever's the latest from Microsoft.

### DIFF
--- a/ci/base/base.yml
+++ b/ci/base/base.yml
@@ -61,7 +61,7 @@ builds:
     with_dot_net: #@ True
     #@yaml/map-key-override
     params:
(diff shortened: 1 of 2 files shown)


mmartell:
Reverting looks like the right thing to do. Not sure what is causing the problem yet, but at least we now have a build failure to look at.

mmartell (approved):

PR #862, opened 2021-08-28
GEODE-9559: demacroize clicache

mmartell:
Macros in C++ complicate debug efforts and code maintenance and are [generally considered old school](https://stroustrup.com/icsm-2012-demacro.pdf). This PR is to remove all the complicated macros in the .NET Framework client, e.g. the clicache module.

In addition to improving the maintainability of the clicache module, removing the macros will greatly assist the creation of the .NET Core client. [dotPeek ](https://www.jetbrains.com/decompiler/) is proving to be a valuable tool in the .NET Core project, but is currently limited by the extensive use of macros in the clicache code.


mmartell:
The procedure for this PR consisted of cutting/pasting the original macro invocations with the corresponding preprocessor output. Although there are quite a few files involved, the changes are basically all the same. For this reason it seemed prudent to submit this as a single PR.

clicache/src/Cache.cpp
```diff
@@ -159,13 +165,19 @@ namespace Apache
...
+        catch (System::AccessViolationException^ ex) {
+          throw ex;
+        }
       }
 
       generic<class TKey, class TValue>
       Client::IRegion<TKey,TValue>^ Cache::GetRegion( String^ path )
       {
-        _GF_MG_EXCEPTION_TRY2/* due to auto replace */
+        try {/* due to auto replace */
```
pivotal-jbarrett, line 180: I think this comment is common and not particularly helpful. Can you globally delete?

clicache/src/Cache.cpp
```diff
@@ -215,12 +233,18 @@ namespace Apache
             GC::KeepAlive(m_nativeptr);
           }
 
-        _GF_MG_EXCEPTION_CATCH_ALL2
+        }
+        catch (const apache::geode::client::Exception& ex) {
+          throw Apache::Geode::Client::GeodeException::Get(ex);
+        }
+        catch (System::AccessViolationException^ ex) {
```
pivotal-jbarrett, line 240: Any idea why this managed exception is specifically caught only to be rethrown?

clicache/src/CacheTransactionManager.cpp
```diff
@@ -33,111 +33,165 @@ namespace Apache
...
+        catch (System::AccessViolationException^ ex) {
+          throw ex;
+        }
       }
 
       void CacheTransactionManager::Prepare( )
       {
-        _GF_MG_EXCEPTION_TRY2
-
+        try {
```
pivotal-jbarrett, line 49: Did you consider using lambdas to wrap the function body between these two macros to keep the common code non-repetitive?

Something like (untested):
```C++
void CacheTransactionManager::Prepare() {
  UnmanagedUtils::run([&](){
    m_nativeptr->prepare();
  });
}

...
template <typename F>
void NativeUtils::run(F function) {
  try {
    f();
  } catch (const apache::geode::client::Exception& ex) {
    throw Apache::Geode::Client::GeodeException::Get(ex);
  } catch (System::AccessViolationException^ ex) {
    throw ex;
  }
}
```

clicache/src/CacheableBuiltins.hpp
```diff
@@ -356,129 +356,56 @@ namespace Apache
...
-
       // Built-in CacheableKeys
 
       /// <summary>
       /// An immutable wrapper for booleans that can serve
       /// as a distributable key object for caching.
       /// </summary>
-      _GFCLI_CACHEABLE_KEY_DEF_NEW(native::CacheableBoolean,
-                                   CacheableBoolean, bool);
+      ref class CacheableBoolean : public CacheableBuiltinKey<native::CacheableBoolean, bool, static_cast<int8_t>(DSCode::CacheableBoolean)> { public: inline CacheableBoolean() : CacheableBuiltinKey() { } inline CacheableBoolean(bool value) : CacheableBuiltinKey(value) { } inline static CacheableBoolean^ Create(bool value) { return gcnew CacheableBoolean(value); } inline static explicit operator bool (CacheableBoolean^ value) { return value->Value; } static ISerializable^ CreateDeserializable() { return gcnew CacheableBoolean(); } internal: static ISerializable^ Create(std::shared_ptr<native::Serializable> obj) { return (obj != nullptr ? gcnew CacheableBoolean(obj) : nullptr); } private: inline CacheableBoolean(std::shared_ptr<native::Serializable> nativeptr) : CacheableBuiltinKey(nativeptr) { } };;
```
pivotal-jbarrett, line 365: Oh no, we really need to format this.

clicache/src/CacheableHashSet.hpp
```diff
@@ -598,69 +604,11 @@ namespace Apache
...
...lic ref class CacheableHashSet : public Internal::CacheableHashSetType<static_cast<int8_t>(native::internal::DSCode::CacheableHashSet), apache::geode::client::CacheableHashSet> { public: inline CacheableHashSet() : Internal::CacheableHashSetType<static_cast<int8_t>(native::internal::DSCode::CacheableHashSet), apache::geode::client::CacheableHashSet>() {} inline CacheableHashSet(System::Int32 size) : Internal::CacheableHashSetType<static_cast<int8_t>(native::internal::DSCode::CacheableHashSet), apache::geode::client::CacheableHashSet>(size) {} inline static CacheableHashSet^ Create() { return gcnew CacheableHashSet(); } inline static CacheableHashSet^ Create(System::Int32 size) { return gcnew CacheableHashSet(size); } static ISerializable^ CreateDeserializable() { return gcnew CacheableHashSet(); } internal: static ISerializable^ Create(std::shared_ptr<apache::geode::client::Serializable> obj) { return gcnew CacheableHashSet(obj); } private: inline CacheableHashSet(std::shared_ptr<apache::geode::client::Serializable> nativeptr) : Internal::CacheableHashSetType<static_cast<int8_t>(native::internal::DSCode::CacheableHashSet), apache::geode::client::CacheableHashSet>(nativeptr) { } };;
```
pivotal-jbarrett, line 611: These need some basic formatting too.

clicache/src/ExceptionTypes.hpp
```diff
@@ -266,417 +266,3411 @@ namespace Apache
...
+          return std::make_shared<
+              apache::geode::client::IllegalArgumentException>(message);
+        }
+      };
 
       /// <summary>
       /// Thrown when the state of cache is manipulated to be illegal.
       /// </summary>
-      _GF_MG_EXCEPTION_DEF3( IllegalStateException );
+      [Serializable] public ref class IllegalStateException
```
pivotal-jbarrett, line 385: Is there a way to templatize this instead?

pivotal-jbarrett (dismissed):

pivotal-jbarrett:
> The procedure for this PR consisted of cutting/pasting the original macro invocations with the corresponding preprocessor output. Although there are quite a few files involved, the changes are basically all the same. For this reason it seemed prudent to submit this as a single PR.

Is there a way to get the preprocessor to do this for you for specific macros only and then preserve that output so we can be certain there are not copy/paste errors?

clicache/src/Cache.cpp
mmartell: Interesting question that deserves some thought. AccessViolations are baaaad. So you don't want to just eat them. OK, good, we don't. We rethrow them. But why catch them in the first place?

The only thing I can come up with is to help track down bugs in our code. We catch/rethrow AccessViolations all over the place (212 places), so it allows setting breakpoints which can be caught in the debugger so we can look around.

This may have been useful early on, but I doubt it provides any value now. I'm good with removing: catch (System::AccessVilationException) everywhere.

clicache/src/Cache.cpp
mmartell: Agreed!

clicache/src/CacheTransactionManager.cpp
mmartell: Seems like a nice refactor.

clicache/src/CacheableBuiltins.hpp
mmartell: Yup. Missed this one.

clicache/src/ExceptionTypes.hpp
mmartell: I guess macros are kind of like templates. Generally old school, but they do help minimize lines of source. This one in particular would be a good one to try and templatize. Will investigate.

mmartell:
Although Microsoft has a new preprocessor that follows the standard (I didn't even know there was a standard), it doesn't appear to allow enabling/disabling specific macro expansion. However, I'm not too worried about copy/paste errors here. The compiler and subsequent clicache tests should catch any errors.

clicache/src/ExceptionTypes.hpp
mmartell: I don't see a way to templatize the actual name of a class. Compiler gives this error:
     "class template and template parameter cannot have the same name"

This is probably why our native exceptions are also not templated.

clicache/src/Cache.cpp
mmartell: Done

clicache/src/CacheTransactionManager.cpp
mmartell: We don't currently use any lambdas in C++/CLI code. It's not clear to me if it is even supported. Possibly for unmanaged code I believe you are indicating.  However, for this PR I'm not sure it's with introducing new features in a deprecated compiler.

pivotal-jbarrett (commented):
I don't have the time to review. I will just remove my block.

echobravopapa (changes requested):
Let's re-group on this effort to make large changes to soon to be deprecated library

mmartell:
Closing this PR since it doesn't add new functionality or fix an existing bug. For anyone interested in decompiling the C++/CLI code they can checkout my branch: mmartell:GEODE-9559-demacroize-clicache

PR #864, opened 2021-09-02
GEODE-9576: Fix for single-hop function execution

jvarenina:
Fault:
"InternalFunctionInvocationTargetException: Multiple target nodes
found for single hop operation" occurs on server when executing
function in a single hop manner for all buckets during the period when
client bucket metadata doesn't contains all buckets locations.

Fix:
The client will execute function in a non single-hop manner until it
recevies all buckets locations. This solution is aligned with java
client.

### DIFF
--- a/cppcache/integration/test/FunctionExecutionTest.cpp
+++ b/cppcache/integration/test/FunctionExecutionTest.cpp
@@ -32,6 +32,7 @@
 using apache::geode::client::Cache;
 using apache::geode::client::Cacheable;
 using apache::geode::client::CacheableArrayList;
+using apache::geode::client::CacheableKey;
 using apache::geode::client::CacheableString;
 using apache::geode::client::CacheableVector;
 using apache::geode::client::CacheFactory;
@@ -52,6 +53,9 @@ std::shared_ptr<Region> setupRegion(Cache &cache) {
 }
 
 class TestResultCollector : public ResultCollector {
+ public:
+  bool isCleared = false;
+
   virtual std::shared_ptr<CacheableVector> getResult(
       std::chrono::milliseconds) override {
     return std::shared_ptr<CacheableVector>();
@@ -61,7 +65,9 @@ class TestResultCollector : public ResultCollector {
 
   virtual void endResults() override {}
 
-  virtual void clearResults() override {}
+  virtual void clearResults() override { isCleared = true; }
+
+  bool isClearTriggered() { return isCleared; }
 };
 
 TEST(FunctionExecutionTest, UnknownFunctionOnServer) {
@@ -175,6 +181,58 @@ TEST(FunctionExecutionTest,
   cache.close();
 }
 
+void populateRegion(const std::shared_ptr<Region> &region) {
+  for (int i = 0; i < 113; i++) {
+    region->put("KEY--" + std::to_string(i), "VALUE--" + std::to_string(i));
+  }
+}
+
+TEST(FunctionExecutionTest, FunctionExecutionSingleHopNonHA) {
+  Cluster cluster{
+      LocatorCount{1}, ServerCount{3},
+      CacheXMLFiles(
+          {std::string(getFrameworkString(FrameworkVariable::TestCacheXmlDir)) +
+               "/func_cacheserver1_pool_nonHA.xml",
+           std::string(getFrameworkString(FrameworkVariable::TestCacheXmlDir)) +
+               "/func_cacheserver2_pool_nonHA.xml",
+           std::string(getFrameworkString(FrameworkVariable::TestCacheXmlDir)) +
+               "/func_cacheserver3_pool_nonHA.xml"})};
+
+  cluster.start([&]() {
+    cluster.getGfsh()
+        .deploy()
+        .jar(getFrameworkString(FrameworkVariable::JavaObjectJarPath))
+        .execute();
+  });
+
+  auto cache = CacheFactory().create();
+  auto poolFactory = cache.getPoolManager().createFactory();
+
+  cluster.applyLocators(poolFactory);
+
+  auto pool =
+      poolFactory.setPRSingleHopEnabled(true).setRetryAttempts(0).create(
+          "pool");
+
+  auto region = cache.createRegionFactory(RegionShortcut::PROXY)
+                    .setPoolName("pool")
+                    .create("partition_region");
+
+  populateRegion(region);
+
+  for (int i = 0; i < 30; i++) {
+    auto functionService = FunctionService::onRegion(region);
+    auto rc =
+        functionService.withCollector(std::make_shared<TestResultCollector>())
+            .execute("MultiGetAllFunctionNonHA");
+
+    std::shared_ptr<TestResultCollector> resultCollector =
+        std::dynamic_pointer_cast<TestResultCollector>(rc);
+
+    ASSERT_FALSE(resultCollector->isClearTriggered());
+  }
+}
+
 const std::vector<std::string> serverResultsToStrings(
     std::shared_ptr<CacheableVector> serverResults) {
   std::vector<std::string> resultList;
@@ -191,6 +249,7 @@ const std::vector<std::string> serverResultsToStrings(
 
   return resultList;
 }
+
 TEST(FunctionExecutionTest, OnServersWithReplicatedRegionsInPool) {
   Cluster cluster{
       LocatorCount{1}, ServerCount{2},
--- a/cppcache/src/ClientMetadataService.cpp
+++ b/cppcache/src/ClientMetadataService.cpp
@@ -569,7 +569,6 @@ std::shared_ptr<BucketServerLocation> ClientMetadataService::findNextServer(
 std::shared_ptr<ClientMetadataService::ServerToBucketsMap>
 ClientMetadataService::pruneNodes(
     const std::shared_ptr<ClientMetadata>& metadata, const BucketSet& buckets) {
-  BucketSet bucketSetWithoutServer;
   ServerToBucketsMap serverToBucketsMap;
 
   auto prunedServerToBucketsMap = std::make_shared<ServerToBucketsMap>();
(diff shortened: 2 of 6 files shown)


cppcache/integration/test/FunctionExecutionTest.cpp
```diff
@@ -175,6 +181,58 @@ TEST(FunctionExecutionTest,
...
+    region->put("KEY--" + std::to_string(i), "VALUE--" + std::to_string(i));
+  }
+}
+
+TEST(FunctionExecutionTest, FunctionExecutionSingleHopNonHA) {
+  Cluster cluster{
+      LocatorCount{1}, ServerCount{3},
+      CacheXMLFiles(
+          {std::string(getFrameworkString(FrameworkVariable::TestCacheXmlDir)) +
+               "/func_cacheserver1_pool_nonHA.xml",
```
pivotal-jbarrett, line 195: We should be leveraging modern configuration APIs, not the legacy cache.xml, whenever possible. This may mean providing `gfsh` commands to configure the servers via cluster config. At the very least the cache.xml files should be located in the new integration tests as resources and not the old tests.

cppcache/integration/test/FunctionExecutionTest.cpp
```diff
@@ -175,6 +181,58 @@ TEST(FunctionExecutionTest,
...
+  auto region = cache.createRegionFactory(RegionShortcut::PROXY)
+                    .setPoolName("pool")
+                    .create("partition_region");
+
+  populateRegion(region);
+
+  for (int i = 0; i < 30; i++) {
+    auto functionService = FunctionService::onRegion(region);
+    auto rc =
+        functionService.withCollector(std::make_shared<TestResultCollector>())
```
pivotal-jbarrett, line 226: How is this test asserting that the PR metadata is not complete for region prior to and during the execution of this function?

cppcache/src/ClientMetadataService.cpp
```diff
@@ -578,12 +577,10 @@ ClientMetadataService::pruneNodes(
     const auto locations = metadata->adviseServerLocations(bucketId);
     if (locations.size() == 0) {
       LOGDEBUG(
-          "ClientMetadataService::pruneNodes Since no server location "
-          "available for bucketId = %d  putting it into "
-          "bucketSetWithoutServer ",
+          "ClientMetadataService::pruneNodes Use non single-hop path "
```
pivotal-jbarrett, line 580: These changes strike me as unit testable. Please produce a unit test for these changes.

tests/javaobject/MultiGetAllFunctionNonHA.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.geode.cache.execute.ResultSender;
+import org.apache.geode.cache.Region;
+import org.apache.geode.cache.execute.Function;
+import org.apache.geode.cache.execute.FunctionContext;
+import org.apache.geode.cache.execute.RegionFunctionContext;
+import org.apache.geode.cache.partition.PartitionRegionHelper;
+import org.apache.geode.distributed.internal.InternalDistributedSystem;
+import org.apache.geode.internal.cache.execute.InternalFunctionInvocationTargetException;
+
+public class MultiGetAllFunctionNonHA extends FunctionAdapter implements Declarable{
```
pivotal-jbarrett, line 43: `FunctionAdapter` is deprecated and should not be used. `Function` has generic type parameters that should be declared.

tests/javaobject/MultiGetAllFunctionNonHA.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public void execute(FunctionContext context) {
```
pivotal-jbarrett, line 45: This and other methods should be declared as `@Override`.

tests/javaobject/MultiGetAllFunctionNonHA.java
```diff
@@ -0,0 +1,70 @@
...
+    RegionFunctionContext regionContext = (RegionFunctionContext) context;
+    final Region<String, String> region =
+        PartitionRegionHelper.getLocalDataForContext(regionContext);
+
+    Set<String> keys = region.keySet();
+    List<String> listKey = new ArrayList<>();
```
pivotal-jbarrett, line 51: Why and intermediate list since `ResultSender` can send multiple results, just iterate until you are at the last key and send that using `lastResult`.

cppcache/src/ClientMetadataService.cpp
```diff
@@ -578,12 +579,10 @@ ClientMetadataService::pruneNodes(
...
+          "since no server location is available for bucketId = %d",
           bucketId);
-      bucketSetWithoutServer.insert(bucketId);
-      continue;
+      return nullptr;
```
echobravopapa, line 585: I'd like to see an integration test that exposes the issue being fixed...

echobravopapa (dismissed):

cppcache/integration/test/FunctionExecutionTest.cpp
pivotal-jbarrett: Unfortunately this is notoriously hard to test. PR metadata updates are asynchronous and there isn't any hooks to trigger when an update is complete. You may want to start with adding that function. The downside to that is to use them in an integration tests we would have to export those functions and we try not to export internal/test symbols if we can. Perhaps it might be easier to create a unit test that asserts this part of the function execution logic does the correct thing when the metadata is incomplete. It should be easier to tease out a function that is given the metadata and makes this decisions and then write a test to verify the expected behaviors.

pivotal-jbarrett (approved):
I like the unit test. I am less concerned about the integration test given the nature of testing this asynchronous event.

cppcache/integration/test/FunctionExecutionTest.cpp
echobravopapa: A similarly difficult issue was tested and fixed in recent months, regarding incorrect EventIds etc...  which fixed a "fix" that was making native operate at parity with java client, but the initial fix was incomplete... 
I'm sure @pdxcodemonkey has more to say about this...

pdxcodemonkey: The `putAllAndVerifyKeysExist` test, added in this PR: https://github.com/apache/geode-native/pull/784/files

(very nearly?) guarantees a state in which not all metadata buckets are present on the client, so I kind of expect executing a single-hop function instead of the `putAll` call in this test code would reproduce.  This test has been _very_ reliable, so I wouldn't expect to have a lot of issues with flaky behavior with respect to metadata updates.

jvarenina: Thanks for the review! Really appreciate it!

Actually I already did the same thing in my integration test `FunctionExecutionWithIncompleteBucketLocations`. If you inspect it closely you will see following similarities:

1. I do same here as you when execute 50 puts. Additionally, I use hook to check that metadata is updated on client. For this case it is important to have incomplete metadata (number of locations is not important in retrieved metadata), because in case metadata is never refreshed on client the faulty case would not be triggered. I think that you use sleep instead the hook for this reason.
```
  // Populate region in a way that not all buckets are created.
  // Servers in this case will create 88 of possible 113 buckets.
  populateRegion(region);

  // Check that PR metadata is updated. This is done to be sure
  // that client will execute function in a non single hop manner
  // because metadata doesn't contain all bucket locations.
  // After metadata is refreshed, it will contain at least one
  // bucket location.
  CacheImpl *cacheImpl = CacheRegionHelper::getCacheImpl(&cache);
  waitUntilPRMetadataIsRefreshed(cacheImpl);
```
2. Then I execute function, which then trigger faulty case. You can remove fix and execute this case, and you will get "InternalFunctionInvocationTargetException" every time.
```
  auto functionService = FunctionService::onRegion(region);
  auto rc =
      functionService.withCollector(std::make_shared<TestResultCollector>())
          .execute("MultiGetAllFunctionNonHA");
```

cppcache/src/ClientMetadataService.cpp
jvarenina: Thank you for the review!

There is already integration "FunctionExecutionTest.FunctionExecutionWithIncompleteBucketLocations" test in this PR that exposes the issue. Please check comment https://github.com/apache/geode-native/pull/864#discussion_r705952887 .

cppcache/integration/test/FunctionExecutionTest.cpp
pdxcodemonkey: Sorry, totally missed this test - dunno how that happened.  To confirm your conjecture above, yes my test sleeps because it has to have a metadata update, with partial info, to repro the failure, just as yours does.

cppcache/test/mock/ClientMetadataMock.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define GEODE_CLIENTMETADATAMOCK_H_
+
+#include <gmock/gmock.h>
+
+#include "ClientMetadata.hpp"
+
+namespace apache {
+namespace geode {
+namespace client {
+class ClientMetadataMock : public ClientMetadata {
```
mreddington, line 30: I would rather you overload the ClientMetadata constructor to accept a bucket set than burden the system with yet more polymorphism so you can inject a bucket set.

pivotal-jbarrett: I am not sure I understand @mreddington, this is mock and this is how you mock classes using gmock.

cppcache/src/ClientMetadataService.cpp
jvarenina: Hi @echobravopapa ,

Is this integration test OK to you? Maybe to remove request change if there is nothing else to add?

Thanks/Jakov

jvarenina:
Hi all, sorry for bothering. I just want to double check with you all if there is something that I missed in this PR and should be done by me. I still see request change by @echobravopapa , but I think that one is resolved. Am I right?

jvarenina:
Should we merge this PR or still need to wait on some more reviews?

cppcache/integration/test/FunctionExecutionTest.cpp
```diff
@@ -175,6 +196,88 @@ TEST(FunctionExecutionTest,
...
+  std::shared_ptr<TestResultCollector> resultCollector =
+      std::dynamic_pointer_cast<TestResultCollector>(rc);
+
+  // check that function is not re-executed
+  ASSERT_FALSE(resultCollector->isClearTriggered());
+
+  // check that PR metadata is updated after function is executed
+  waitUntilPRMetadataIsRefreshed(cacheImpl);
+
+  // check that correct nubmer of events is received in function result
```
alb3rtobr, line 274: There is a typo here (`nubmer`)

alb3rtobr (approved):

PR #866, opened 2021-09-13
GEODE-9591: Don't retry function when Function.isHA=false

jvarenina:
Fault:
Client clears result by calling ResultCollector::clearResult and
re-executes the function even thought Function.isHa() is set to false.

Fix:
Client will not clear result and retry the function, but it will fail with the
received exception from server.

Added integration test:
1.FunctionExecutionTest.testThatFunctionExecutionThrowsExceptionNonHAWithFilter
2.FunctionExecutionTest.testThatFunctionExecutionThrowsExceptionNonHA

The integration test reproduces the problem in a following way:

1. Populate region so that all buckets are created
2. Shutdown one of the three available servers
3. Peforme rebalance operation, so that buckets are recovered on running
servers
4. Execute the function. Function execution will fail with the
InternalFunctionInvocationTargetException exception, because client will
try to execute single hop function with old PR meatada information (PR
metadata received before the rebalance operation).


jvarenina:
I put as review ready even thought two test jobs failed, because I think that failed test cases could be flaky:

- HARegionCacheListenerARLTest.afterRegionLive 
- TransactionCleaningTest.txWithStoppedServer

pdxcodemonkey:
Can confirm, if any test cases are gonna fail it'll be these two.

Thanks,

Blake

alb3rtobr (approved):

pdxcodemonkey:
@jvarenina This one needs a little attention, I think.  If you rebase to fix the merge conflicts, we can look at the resultant CI run and see if there are any lingering test issues.

jvarenina:
@pdxcodemonkey Ok, I will keep eye on it. I have checked `OnServerHAExecuteFunctionTest` test case, and I couldn't find any connection to this PR solution. This PR only impacts function executions that have `Function.isHA` set to `false` (this flag is checked before all impact introduced in this PR), and `OnServerHAExecuteFunctionTest` has this parameter set to `true`. I will execute this failing case and check this in more details. Thank you pointing this out to me.

PR #868, opened 2021-09-20
GEODE-9608: Use Warnings as Errors and add RelWithDebInfo in NetCore

mmartell:


pivotal-jbarrett (changes requested):
I think you can just eliminate `Release`.

I also think there are other properties you need in RelWithDebInfo to actually produce debug information and release optimized byte code.

PR #871, opened 2021-09-24
GEODE-9553: remove sprintf and snprintf from geode native

pdxcodemonkey:


cppcache/src/TcrMessage.cpp
```diff
@@ -3212,12 +3206,10 @@ TcrMessageHelper::ChunkObjectType TcrMessageHelper::readChunkPartHeader(
   // TODO enum - wtf?
```
echobravopapa, line 3206: while your in there... maybe delete this line, thanks...

echobravopapa (approved):

PR #874, opened 2021-10-01
GEODE-9639: Removed the use of std::unary_function.

mreddington:


pivotal-jbarrett:
@mreddington You will find them also in use in SerializationRegistry.hpp.

mmartell (commented):
Shouldn't we also remove a couple others that are deprecated?
* std::binary_function (functional.hpp)
* std::not1 (TcrMessage.cpp)

pdxcodemonkey (changes requested):
As Jake mentioned, there are two instances of unary_function in SerializationRegistry.hpp.  Please fix these the same way, and you should be good to go.

pivotal-jbarrett:
> * std::not1 (TcrMessage.cpp)

Replace that whole `std::string::erase` statement with [`boost::algorithm::trim_left`](https://www.boost.org/doc/libs/1_77_0/doc/html/boost/algorithm/trim_left.html).

PR #881, opened 2021-10-12
GEODE-9719: Start/Stop cluster in a test fixture

mmartell:
This switches the NetCore test framework to Xunit fixtures. A separate fixture is used for each of the test suites (i.e., netcore-integration-test and netcore-session-integration-test. It also just creates an authRegion not separate cluster for the netcore auth tests.


mmartell:
Rebased against develop to pickup fix for .../integration/test/FunctionExecutionTest.cpp

PR #884, opened 2021-10-19
GEODE-9741: Use v142 Toolchain in Visual Studio 2019 CI Pipeline

mmartell:
Currently the VS 2019 builds use the old toolchain (v141). All tests pass with v142, so just requires a small change to CI code, as well as adding the v142 tools to the Windows 2019 image.


mmartell:
Actually, looks like there is still a problem related to my last commit. Investigating.

PR #888, opened 2021-11-01
GEODE-9741: Fix Release Build on Windows 2019

mmartell:
Building with the new tools (v142) on Windows caused the Release build to fail due to ACE not picking up the specified tools from the cmake configuration. This PR solves this problem by setting environment variables by calling vcvarsall.bat with the correct tools version  before configuring with cmake.

Note: Once geode-native is no longer dependent on ACE, the call to vcvarsall.bat can be removed.

### DIFF
--- a/ci/lib/templates.lib.txt
+++ b/ci/lib/templates.lib.txt
@@ -203,6 +203,9 @@ git_rev=$(git rev-parse HEAD)
 popd
 version=$(cat version/number)
 builddate=$(date "+%Y-%m-%d")
+#@ if build.image_family == "build-windows-2019-vs-2019":
+  remote_shell 'c:/"Program Files (x86)\Microsoft Visual Studio"/2019/Community/VC/Auxiliary/Build/vcvarsall.bat x64 10.0.16299.0 -vcvars_ver=14.20'
+#@end
 remote_shell cmake -E make_directory build
 remote_shell cmake -E time cmake -E chdir build cmake ../source ${CMAKE_CONFIGURE_FLAGS} \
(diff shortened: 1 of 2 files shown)


ci/lib/templates.lib.txt
```diff
@@ -203,6 +203,9 @@ git_rev=$(git rev-parse HEAD)
 popd
 version=$(cat version/number)
 builddate=$(date "+%Y-%m-%d")
+(@- if use_vcvarsall: @)
+remote_shell "c:/Program Files (x86)\Microsoft Visual Studio/2019/Community/VC/Auxiliary/Build/vcvarsall.bat" x64 10.0.16299.0 -vcvars_ver=14.20
```
pivotal-jbarrett, line 207: How are these env changes surviving past this single invocation of the command processor?

Needed my vcvars is a heavy hammer. A small fix to ACE CMake file should pass all you need to ACE.

pivotal-jbarrett (changes requested):
I don't see how this changes the build env between calls to the shell.

ci/lib/templates.lib.txt
mmartell: **Great catch. I**t worked on my local system because it's a single cmd shell. I took a look at how to do this directly in ACE's CMakeLists.txt, but it doesn't look like MPC has a value_template for the tools version. Any ideas on this would be greatly appreciated. Otherwise, can I just pass the vcvarsall.bat as the first argument to the remote_shell invocation that does the cmake config?

pivotal-jbarrett:
Replaced with #889

PR #891, opened 2021-11-05
GEODE-9753: Solve PDX serialization coredump

gaussianrecurrence:
- While serializing a PdxSerializable object, there is a possible race
   condition which might cause the client to crash.
   This race-condition happens whenever the cluster is restarted during
   the serialization process and if
   on-client-disconnect-clear-pdxType-Ids is set to true, meaning the
   PdxTypeRegistry will be cleaned up if the connection towards the
   cluster is lost.
 - This issue has been solved by using the previously fetched local PDX
   type.
 - In order to properly test this solution, PdxRemoteWriterFactory has been added,
   so the race-condition can be force at test-time.
 - A new IT has been added to test that the solution is working fine.
 - make_unique was needed inside cppcache/src, so it was moved to
   utils/cxx_extensions.hpp
 - Also in order to ease the use of newer C++ standard a preprocessor
   check was added to make_unique, so if the used standard >= C++14, the
   standard implementation of make_unique will be used instead.


gaussianrecurrence:
I am reviewing all of my pending PRs, and this has been awating review for quite some time. Maybe @pdxcodemonkey, @pivotal-jbarrett @albertogpz @mkevo @jvarenina or anyone else could take a look at it and provide some feedback?
Thanks!

cppcache/src/PdxHelper.cpp
```diff
@@ -137,28 +136,18 @@ void PdxHelper::serializePdx(
...
 
-    auto createPdxRemoteWriter = [&]() -> PdxRemoteWriter {
-      if (auto pd = pdxTypeRegistry->getPreserveData(pdxObject)) {
-        auto mergedPdxType = pdxTypeRegistry->getPdxType(pd->getMergedTypeId());
-        return PdxRemoteWriter(output, mergedPdxType, pd, pdxTypeRegistry);
-      } else {
-        return PdxRemoteWriter(output, className, pdxTypeRegistry);
-      }
-    };
+    auto prw = cacheImpl->getPdxRemoteWriterFactory().create(
```
pdxcodemonkey, line 139: Can we use a slightly longer, meaningful name?  like `remoteWriter` for instance?

cppcache/src/PdxRemoteWriter.hpp
```diff
@@ -22,11 +22,13 @@
...
 #include "PdxLocalWriter.hpp"
 
+#include <geode/internal/geode_globals.hpp>
+
 namespace apache {
 namespace geode {
 namespace client {
 
-class PdxRemoteWriter : public PdxLocalWriter {
+class APACHE_GEODE_EXPORT PdxRemoteWriter : public PdxLocalWriter {
```
pdxcodemonkey, line 31: Please don't export this, we can't change the ABI and anyway we don't want any more publicly-accessible things than we already have(!).

cppcache/src/PdxRemoteWriterFactoryImpl.hpp
```diff
@@ -0,0 +1,45 @@
...
+#define GEODE_PDXREMOTEWRITERFACTORYIMPL_H_
+
+#include "PdxRemoteWriterFactory.hpp"
+
+
+namespace apache {
+namespace geode {
+namespace client {
+
+class APACHE_GEODE_EXPORT PdxRemoteWriterFactoryImpl : public virtual PdxRemoteWriterFactory {
```
pdxcodemonkey, line 30: Again, please don't export.

cppcache/src/PdxHelper.cpp
pivotal-jbarrett, line 139: What this extracted into an pure virtual interface for the sole purpose of integration testing?

pivotal-jbarrett:
> * This issue has been solved by using the previously fetched local PDX
>   type.

Do you mean this issue is solved in this PR or in a previous and this PR just adds more testing?

cppcache/src/PdxHelper.cpp
```diff
@@ -121,13 +122,11 @@ void PdxHelper::serializePdx(
...
     pdxTypeRegistry->addPdxType(nTypeId, nType);
 
-    if (cacheImpl != nullptr) {
-      uint8_t* stPos = const_cast<uint8_t*>(output.getBuffer()) +
-                       ptc.getStartPositionOffset();
-      int pdxLen = PdxHelper::readInt32(stPos);
-      cachePerfStats.incPdxSerialization(
-          pdxLen + 1 + 2 * 4);  // pdxLen + 93 DSID + len + typeID
-    }
+    uint8_t* stPos =
```
pivotal-jbarrett, line 125: `auto`

cppcache/src/PdxHelper.cpp
```diff
@@ -137,28 +136,18 @@ void PdxHelper::serializePdx(
...
 
     //[ToDo] need to write bytes for stats
-    if (cacheImpl != nullptr) {
-      uint8_t* stPos = const_cast<uint8_t*>(output.getBuffer()) +
-                       prw.getStartPositionOffset();
-      int pdxLen = PdxHelper::readInt32(stPos);
-      cachePerfStats.incPdxSerialization(
-          pdxLen + 1 + 2 * 4);  // pdxLen + 93 DSID + len + typeID
-    }
+    uint8_t* stPos = const_cast<uint8_t*>(output.getBuffer()) +
```
pivotal-jbarrett, line 146: `auto`

pivotal-jbarrett, line 146: Any idea why there was a `nullptr` check here?

cppcache/src/PdxRemoteWriter.cpp
```diff
@@ -49,18 +49,18 @@ PdxRemoteWriter::PdxRemoteWriter(
...
       m_remoteTolocalMapLength(0) {
-  m_preserveData = nullptr;
   if (m_pdxType != nullptr) {
-    m_remoteTolocalMapLength = m_pdxType->getTotalFields();
     m_remoteTolocalMap = m_pdxType->getRemoteToLocalMap();
+    m_remoteTolocalMapLength = m_pdxType->getTotalFields();
   }
-  initialize();
+
+  m_pdxClassName = pdxType->getPdxClassName();
```
pivotal-jbarrett, line 63: If `pdxType` is nullable then this line will throw.

cppcache/src/PdxRemoteWriter.cpp
```diff
@@ -49,18 +49,18 @@ PdxRemoteWriter::PdxRemoteWriter(
...
-    DataOutput& output, std::string pdxClassName,
+    DataOutput& output, std::shared_ptr<PdxType> pdxType,
     std::shared_ptr<PdxTypeRegistry> pdxTypeRegistry)
-    : PdxLocalWriter(output, nullptr, pdxClassName, pdxTypeRegistry),
+    : PdxLocalWriter(output, pdxType, pdxTypeRegistry),
       m_preserveDataIdx(0),
       m_currentDataIdx(-1),
       m_remoteTolocalMapLength(0) {
-  m_preserveData = nullptr;
   if (m_pdxType != nullptr) {
```
pivotal-jbarrett, line 58: If `pdxType` is not nullable then this check is unnecessary.

cppcache/src/PdxRemoteWriter.hpp
```diff
@@ -45,10 +47,10 @@ class PdxRemoteWriter : public PdxLocalWriter {
   PdxRemoteWriter() = default;
 
   PdxRemoteWriter(DataOutput& output, std::shared_ptr<PdxType> pdxType,
-                  std::shared_ptr<PdxRemotePreservedData> preservedData,
                   std::shared_ptr<PdxTypeRegistry> pdxTypeRegistry);
 
-  PdxRemoteWriter(DataOutput& output, std::string pdxClassName,
+  PdxRemoteWriter(DataOutput& output, std::shared_ptr<PdxType> pdxType,
+                  std::shared_ptr<PdxRemotePreservedData> preservedData,
```
pivotal-jbarrett, line 53: It this really feels like an extension of the other constructor with the optional `preservedData` parameters to maybe re-order the parameters to look like that.

cppcache/src/PdxRemoteWriterFactoryImpl.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "PdxTypeRegistry.hpp"
+#include "util/cxx_extensions.hpp"
+
+namespace apache {
+namespace geode {
+namespace client {
+
+PdxRemoteWriterFactoryImpl::~PdxRemoteWriterFactoryImpl() = default;
+
+std::unique_ptr<PdxRemoteWriter> PdxRemoteWriterFactoryImpl::create(
```
pivotal-jbarrett, line 31: It really seems like you could test this at unit level by mocking `PdxTypeRegistery` and `PdxType`. Make this a member function back from where it came and test it a the unit level. This avoids the creation of this factory, the exporting of internal symbols, and the interdiction of artificial behavior into the production implementation, via overriding this implementation.

pivotal-jbarrett (changes requested):

pivotal-jbarrett:
> > * This issue has been solved by using the previously fetched local PDX
> >   type.
> 
> Do you mean this issue is solved in this PR or in a previous and this PR just adds more testing?

Ok, I think I found the fix but I a not sure I understand it. How does preserving the `PdxType` in all cases of the `RemotePdxWriter` construction fix the race condition?

cppcache/src/PdxHelper.cpp
gaussianrecurrence: Exactly, I needed to force the race-condition and this was the only way I came up with. I am not sure there is any other way to force the race-condition.

gaussianrecurrence:
> > * This issue has been solved by using the previously fetched local PDX
> >   type.
> 
> Do you mean this issue is solved in this PR or in a previous and this PR just adds more testing?

The issue I am referring to here is the one I described in the bullet above.

cppcache/src/PdxHelper.cpp
gaussianrecurrence: Well, not really, and it doesn't make any sense, since CacheImpl should exist as long as there is a cache and even if that block of code should not be executed whenever the cache is closed, given cacheImpl is a pointer, the code would still execute. This would only would make sense in the case that cacheImpl were a std::weak_ptr<CacheImpl> instead, but that's not the case.

cppcache/src/PdxRemoteWriter.cpp
gaussianrecurrence: Thing is I did leave that because I thought what if pdxType passed were nullptr, but that's not the way it's intended to be used, so I guess it would be best to force a coredump if pdxType is nullptr, right?

cppcache/src/PdxRemoteWriter.hpp
gaussianrecurrence: This class is exported to allow testing. Thing is I thought there are no issues from the ABI-side if you added new symbols, rather than whenever you changed symbols. Am I wrong?

cppcache/src/PdxRemoteWriterFactoryImpl.cpp
gaussianrecurrence: Your approach seems promising. I'll look into it and see if it's feasible.

cppcache/src/PdxRemoteWriter.cpp
pivotal-jbarrett: Yeah, if it is `nullptr` then line 63 is going to fault anyway. Also, if passed here as `nulltpr` but later use excepted `m_pdxType` to not be `nullptr`... boom.

gaussianrecurrence:
Closing this PR as this issue is already addressed in GEODE-10276

PR #894, opened 2021-11-11
GEODE-9804: new register interest tests

mmartell:


cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -575,4 +581,257 @@ TEST(RegisterKeysTest, RegisterAnyWithProxyRegion) {
...
+      .withName("region")
+      .withType("PARTITION")
+      .execute();
+
+  auto cache1 = createCache();
+  auto pool1 = createPool(cluster, cache1);
+  auto region1 = setupRegion(cache1, pool1);
+  auto attrMutator = region1->getAttributesMutator();
+
+  // put key/value pairs in the region using cache2
```
pivotal-jbarrett, line 629: If a block of code needs a comment it should be refactored to be readable or extracted into a method with a descriptive name. This test is getting pretty long.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -575,4 +581,257 @@ TEST(RegisterKeysTest, RegisterAnyWithProxyRegion) {
...
+  // put new values in the region using cache2
+
+  for (int i = 0; i < NUMKEYS; i++) {
+    region2->put(apache::geode::client::CacheableInt32::create(i),
+                 apache::geode::client::CacheableInt32::create(i + 2000));
+  }
+
+  // wait for the new values in cache1 due to receiveValues = true, and verify
+
+  std::this_thread::sleep_for(5000ms);
```
pivotal-jbarrett, line 757: A test should not have sleeps. Updates arriving at the other cache should trigger a cache listener. You can use that trigger to release this thread to expect the values. Alternatively this thread can await/loop until the assertions are true or a timeout expires.

pivotal-jbarrett (changes requested):
Sleeps in a test are a smell that leads to flaky tests. Please await for assertions to be true or block on a condition.

cppcache/integration/test/RegisterKeysTest.cpp
pivotal-jbarrett: Also, this code will not compile in C++11 since `ms` was added in C++ 14. Also, 5000ms == 5s.
```C++
std::this_thread::sleep_for(std::chromo::seconds(5));
```

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: I agree comments in general are a liability, just like code. They help while debugging the test, but I'm OK removing now.

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: No more sleeps.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -39,19 +42,56 @@ namespace {
...
+using apache::geode::client::CacheListener;
 
 using ::testing::_;
 using ::testing::DoAll;
 using ::testing::InvokeWithoutArgs;
 using ::testing::Return;
 
+const int NUMKEYS = 100;
+
+class MyCacheListener : public CacheListener {
```
pivotal-jbarrett, line 64: For future reference you should learn and use gmock for this sort of thing. You can mock up the `CacheListener` interface and register the mock with the cache. Then you can assert that specific methods on the mock were invoked.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -575,4 +615,233 @@ TEST(RegisterKeysTest, RegisterAnyWithProxyRegion) {
...
+  auto attrMutator = region1->getAttributesMutator();
+  auto listener = std::make_shared<MyCacheListener>();
+  attrMutator->setCacheListener(listener);
+
+  auto cache2 = createCache();
+  auto pool2 = createPool(cluster, cache2);
+  auto region2 = setupRegion(cache2, pool2);
+
+  for (int i = 0; i < NUMKEYS; i++) {
+    region2->put(apache::geode::client::CacheableInt32::create(i),
```
pivotal-jbarrett, line 670: The using statements above mean you can just invoke `CacheableInt32::create`, otherwise delete the using statements.

pivotal-jbarrett (changes requested):

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: You agreed a real CacheListener with count_down latches was appropriate.

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: Good catch. Done.

mmartell:
**NOTE:** The integration test failure on ubuntu 16.04 debug is not related to this PR:

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -39,19 +42,46 @@ namespace {
...
 
 using ::testing::_;
 using ::testing::DoAll;
 using ::testing::InvokeWithoutArgs;
 using ::testing::Return;
 
+constexpr auto kNumKeys = 100;
+
+class MyCacheListener : public CacheListener {
+  std::shared_ptr<boost::latch> m_allKeysUpdatedLatch;
```
pivotal-jbarrett, line 65: Why are these `std::shared_ptr`? Just make them public and await on them directly?
Also, style guide says `camelCaseFollowedByUnderscore_` for members names.

pdxcodemonkey (approved):

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -575,4 +605,223 @@ TEST(RegisterKeysTest, RegisterAnyWithProxyRegion) {
...
+  }
+
+  allKeysInvalidLatch->reset(kNumKeys);
+  listener->reset();
+
+  for (int i = 0; i < kNumKeys; i++) {
+    region2->put(CacheableInt32::create(i), CacheableInt32::create(i + 1000));
+  }
+
+  allKeysInvalidLatch->wait();
```
pivotal-jbarrett, line 686: Should use `wait_for` and assert that it didn't timeout otherwise this test will hang FOREVER if the latch is never satisfied.

pivotal-jbarrett: Something like this snippet from another test.
```C++
ASSERT_THAT(task1.wait_for(minutes(1)), Eq(std::future_status::ready));
```

pivotal-jbarrett (approved):

pivotal-jbarrett (changes requested):
Accidentally approved

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: Done

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: Yah, good catch. Will switch it to wait_for.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -575,4 +605,227 @@ TEST(RegisterKeysTest, RegisterAnyWithProxyRegion) {
...
+      .execute();
+
+  auto cache1 = createCache();
+  auto pool1 = createPool(cluster, cache1);
+  auto region1 = setupRegion(cache1, pool1);
+  auto attrMutator = region1->getAttributesMutator();
+
+  boost::latch allKeysUpdatedLatch{kNumKeys};
+  boost::latch allKeysInvalidLatch{kNumKeys};
+  auto listener = std::make_shared<MyCacheListener>(allKeysUpdatedLatch,
```
pivotal-jbarrett, line 656: There is an issue with the lifespan of these latches not matching that of the listener. Really the listener should construct and destroy these. Add accessor methods to check them or just `wait_for` methods on the listener that calls through to the latch's `wait_for` methods.

cppcache/integration/test/RegisterKeysTest.cpp
```diff
@@ -575,4 +605,227 @@ TEST(RegisterKeysTest, RegisterAnyWithProxyRegion) {
...
```
pivotal-jbarrett, line 718: Same lifespan issue here.

pivotal-jbarrett (changes requested):

cppcache/integration/benchmark/RegisterInterestBM.cpp
```diff
@@ -0,0 +1,156 @@
...
+using apache::geode::client::CacheableString;
+using apache::geode::client::HashMapOfCacheable;
+using apache::geode::client::Region;
+using apache::geode::client::RegionShortcut;
+
+class RegisterInterestBM : public benchmark::Fixture {
+ public:
+  RegisterInterestBM() {
+    boost::log::core::get()->set_filter(boost::log::trivial::severity >=
+                                        boost::log::trivial::debug);
```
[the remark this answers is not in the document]
mmartell: Done

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: Nice suggestion. Moved latches to CountdownListener.

cppcache/integration/test/RegisterKeysTest.cpp
mmartell: Done

pivotal-jbarrett (approved):

PR #897, opened 2021-11-24
Ignore generated netcore/shared/Config.cs

moleske:

I finally got annoyed at my git status being dirty after running cmake ..

I don't like ignoring it, but I couldn't figure out where this file should go.  In the other cases where Config.cs.in is templated out, it is put into the build directory.  But I couldn't see any cs files in the build directory for the new dotnet work.  Let me know the cmakelists.txt should actually be updated instead of ignoring this file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@
 
 /tools/gnmsg/.idea/
 /tools/gnmsg/__pycache__/
+
+netcore/shared/Config.cs


pdxcodemonkey:
Sadly, this is probably not the only such file we need to add.  We're between a rock and a hard place, because `cmake` doesn't have acceptable support for .net core builds (yet).  So, we have to use `dotnet build`, which _refuses_ to build code outside of the source tree.  This means all intermediate files - generated code, object files, libraries, apps - are in the source tree at the end of a full build :/.

moleske:
grumble, grumble, something grr cmake

feel free to close this then.  No strong feelings from me on how to solve the problem.  I will continue to shake fist at clouds like an old man
old man

PR #898, opened 2021-12-02
Fix to disable auto updates on rhel8

mmartell:
Looks like a typo to me. This should disable background updates on rhel8.


gaussianrecurrence:
Thanks for contributing this! I am wondering now if that's the reason why RHEL-8 CI was randomly failing due to some SO errors?

mmartell:
Very likely. If your build is trying to link in one of the shared libraries as it is being updated, you can a file not found (or similar).

pivotal-jbarrett (approved):

PR #901, opened 2021-12-15
GEODE-9502: Eliminate templates used to work around now-obsolete MSVC compiler warning

mreddington:


cppcache/integration/framework/Cluster.cpp
```diff
@@ -146,7 +146,10 @@ void Locator::start() {
 }
 
 void Locator::stop() {
-  cluster_.getGfsh().stop().locator().withDir(name_).execute();
+  try {
+    cluster_.getGfsh().stop().locator().withDir(name_).execute();
+  } catch (...) {
```
pivotal-jbarrett, line 151: Why are we eating exceptions in the teardown of the stopping of these components? If there is an exception that should affect the test. If the specific test wants to ignore the exceptions it would but it shouldn't be the default.

pivotal-jbarrett (changes requested):

cppcache/integration/framework/Cluster.cpp
mreddington: Destructors aren't supposed to throw. We throw an exception when BOOST_PROCESS tries to execute a command on a locator and the locator is already down. This has absolutely nothing to do with the test itself or the native client. We aren't testing the cluster, the testing framework itself, or the environment.

PR #905, opened 2022-01-14
GEODE-9960: Remove strong name signing support for .net assemblies

pdxcodemonkey:
- This has caused problems with build pipelines, and is no longer used by .net runtimes


mmartell (changes requested):
Minor cleanup is the only issue I see: 

- Remove @STRONG_NAME_PUBLIC_KEY_ATTRIBUTE@ from clicache/src/impl/AssemblyInfo.cpp.in

pivotal-jbarrett (approved):

mmartell (approved):

PR #906, opened 2022-01-14
GEODE-9848: Eliminate duplicate REGISTER_INTEREST message

pdxcodemonkey:
- first endpoint in m_redundantEndpoints was originally supposed to
be left out of the for loop.  This broke during a refactor roughly three years ago, moving from old-school iterator to a ranged for loop.


mmartell (approved):
This is a righteous fix!

PR #909, opened 2022-01-15
GEODE-9712: genericize netcore

mmartell:
This brings in support for our existing IRegion<TKey, TValue> interface with the following restrictions:

TKey currently limited to string, short, and int.
TValue currently limited to string only.


mmartell:
Please disregard my previous PR for GEODE-9712. It had merge conflicts due to erroneous branch being created on apache.

pdxcodemonkey (changes requested):
There are a whole ton of cmake output files in this PR.   Please remove all the extraneous stuff and let me know when I can review just the actual changes.  Thanks!

c-bindings/include/geode/region.h
```diff
@@ -40,18 +40,41 @@ APACHE_GEODE_C_EXPORT void apache_geode_Region_PutString(
     apache_geode_region_t* region, const char* key, const char* value);
 
 APACHE_GEODE_C_EXPORT void apache_geode_Region_PutByteArray(
-    apache_geode_region_t* region, const char* key, const char* value, size_t size);
+    apache_geode_region_t* region, const char* key, size_t keyLength,
```
pivotal-jbarrett, line 43: We should really require length parameters for all array pointers, it's not safe to rely on `NULL` termination.

pivotal-jbarrett (changes requested):

c-bindings/src/data_serializable_raw.hpp
```diff
@@ -0,0 +1,53 @@
...
+class DataSerializableRaw : public DataSerializablePrimitive , public CacheableKey {
+ public:
+  DataSerializableRaw(const int8_t* data, size_t size);
+  ~DataSerializableRaw() noexcept override = default;
+
+  static std::shared_ptr<DataSerializableRaw> create(
+      const int8_t* data, size_t size);
+
+
+  virtual void toData(DataOutput& dataOutput) const override;
```
pivotal-jbarrett, line 36: `virtual` is implied by `override`. `clang-tidy` should be complaining about this.

c-bindings/src/data_serializable_raw.cpp
```diff
@@ -0,0 +1,80 @@
...
+  if (auto otherKey = dynamic_cast<const DataSerializableRaw*>(&other)) {
+    return bytes_ == otherKey->bytes_;
+  }
+
+  return false;
+}
+
+int32_t DataSerializableRaw::hashcode() const {
+  if (hashCode_ == 0) {
+    hashCode_ = internal::geode_hash<std::vector<int8_t>>{}(bytes_);
```
pivotal-jbarrett, line 72: This is going to result in a hash on the client that is not consistent with the hash on the server. The hash value needs to come from the pre-serialized form of the object being serialized here. This will result in sub-optimal put operations.

c-bindings/src/data_serializable_raw.cpp
```diff
@@ -0,0 +1,80 @@
...
+bool DataSerializableRaw::operator==(const CacheableKey& other) const {
+  if (auto otherKey = dynamic_cast<const DataSerializableRaw*>(&other)) {
+    return bytes_ == otherKey->bytes_;
+  }
+
+  return false;
+}
+
+int32_t DataSerializableRaw::hashcode() const {
+  if (hashCode_ == 0) {
```
pivotal-jbarrett, line 71: `0` is a valid hash so this could result in multiple re-hashes.

c-bindings/src/data_serializable_raw.hpp
```diff
@@ -0,0 +1,53 @@
...
+  virtual void fromData(DataInput& dataInput) override;
+  virtual DSCode getDsCode() const override;
+
+  bool operator==(const CacheableKey& other) const override;
+
+  int32_t hashcode() const override;
+
+  private:
+   std::vector<int8_t> bytes_;
+   static constexpr int dsCodeSize_ = 1;
```
pivotal-jbarrett, line 46: Constants have a different naming convention.

c-bindings/src/data_serializable_raw.hpp
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
```
pivotal-jbarrett, line 19: Missing header guard. `#pragma once` is not a standard.

c-bindings/src/data_serializable_raw.cpp
```diff
@@ -0,0 +1,80 @@
...
+    hashCode_ = internal::geode_hash<std::vector<int8_t>>{}(bytes_);
+  }
+  return 0;
+}
+
+}  // namespace internal
+}  // namespace client
+}  // namespace geode
+}  // namespace apache
```
pivotal-jbarrett, line 80: Missing new line.

c-bindings/src/data_serializable_raw.cpp
```diff
@@ -0,0 +1,80 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <iterator>
+
+// C++ client public headers
+#include "geode/DataInput.hpp"
+#include "geode/DataOutput.hpp"
+#include "geode/internal/DataSerializablePrimitive.hpp"
```
pivotal-jbarrett, line 23: The corresponding header for a class should always be listed first.

pivotal-jbarrett (changes requested):

c-bindings/include/geode/region.h
mmartell: I don't disagree. However, I'm thinking we should consider the single more powerful apache_geode_Region_Put (line 50), which includes lengths for both key and value, and do away with all other versions of the Put API.

c-bindings/src/data_serializable_raw.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+using apache::geode::client::DataInput;
+using apache::geode::client::DataOutput;
+
+namespace apache {
+namespace geode {
+namespace client {
+namespace internal {
+
+DataSerializableRaw::DataSerializableRaw(const int8_t* data, size_t size) {
```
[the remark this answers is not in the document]
mmartell: Ideally, the C# layer would have access to the DataInput and DataOutput buffers via the Region. However, this would likely require moving too much buffer management code to C#. An alternative that might be worth consideration is to have an interop buffer in the C++ heap (an std::vector), and give C# its address. This could be created during the cache create and deleted when the cache is closed. Always passing the length(s) of interop parameters would allow growing the buffer as needed.

c-bindings/src/data_serializable_raw.hpp
mmartell: Good catch! I guess you're better than clang-tidy.

c-bindings/src/data_serializable_raw.cpp
mmartell: Good catch! I think we're going to have to create the CacheableKey objects at the C_Bindings layer, similar to what clicache does.

pivotal-jbarrett: The probably with `CacheableKey` is that it expects an equality rule that isn't necessary at the serialization layer but is at the local cache layer. The C-binding is going to need to provide a function pointer, or something I can't think of right now, to solve that local caching key issue. Value objects don't need hash or equality.

pivotal-jbarrett: When I thought about this problem years ago I felt that the C-binding layer would either need to below the local caching and serialization layers and effectively be a networking layer, or the C-binding will need to keep track of pointers back to the original C types and some function pointers. Neither of them are going to be easy.

```
class VoidCachableKey {
  public void VoidCacheableKey(const void* key, int32_t (*hash)(const void *), bool (*equals)(const void *, const void *, const char* serialized, size_t serializedLength) : ... {}
  int32_t hashcode() {
    return hash_(key);
  }
  bool equals(const VoidCachableKey& other) {
    return equals_(key, other.key_);
  }
}
```

c-bindings/src/data_serializable_raw.cpp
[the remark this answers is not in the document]
pdxcodemonkey: To an extent this is a "six of one and half a dozen of another" issue, but I suppose it is a little bit cleaner to just go with one parameter of type `std::vector` in the ctor, then use `std:move` move to copy it into bytes_.

c-bindings/src/data_serializable_raw.cpp
pdxcodemonkey: @pivotal-jbarrett I think I need the short course in what we're using `hashcode` for that causes issues here.  I wasn't aware that we were comparing or otherwise using server-side hashes against client-side, which... sounds like a bad idea to begin with, but I have to admit I'm not surprised by.

pivotal-jbarrett: The hash is used to route the keys to the correct server. If the client and server don't agree on the hashing algorithm and result then they will silently work but it results in suboptimal operations with multiple hops.
Equals is only going to be used client side for local caching, if caching is enabled. Hash maps require both a hash and equality function. The equality on the client and server don't have to match but given the hash needs to match it silly not to have the equality match. One problem that could exist if you don't use constant equality is that the server and client could disagree on what keys are in the caches. Imagine an object that has a field that is serialized, like timestamp, that isn't part of the equality or hash because its ephemeral but we hash the serialized form then the server and client won't agree on the hash value. Then if caching was enabled and equality is also based on the serialized form then the sever would only have one Key('foo') with some timestamp value X and the client could have Key('foo') with timestamp Y and Key('foo') with timestamp X. Local cache lookups would fail as timestamp advances, which could cause massive heap exhaustion.

pdxcodemonkey: Okay, so the `DataSerializableRaw` hack is essentially useless, is what I hear you saying, because without a way to implement  `hashcode` and `==` we can only use it for values.  Probably time for a new plan for keys.

pivotal-jbarrett: Yes, and for values you don't need the hash.

mmartell:
With the decision to go with a pure C# client for .NET, we no loner need this interop based approach.

PR #910, opened 2022-01-19
GEODE-9968: Fix new attrs in PdxSerializable

gaussianrecurrence:
- Solved addition of attributes for previously defined PdxSerializable
   implementations, which was not working before.
 - Added a new IT to verify that it's possible to add a new attribute
   under all scenarios.

### DIFF
--- a/cppcache/integration/test/PdxSerializableTest.cpp
+++ b/cppcache/integration/test/PdxSerializableTest.cpp
@@ -1,0 +1,178 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


cppcache/integration/test/PdxSerializableTest.cpp
```diff
@@ -0,0 +1,178 @@
...
+  entryFetched = region->get("2");
+  ASSERT_TRUE(entryFetched);
+
+  address = std::dynamic_pointer_cast<DynamicFieldsType>(entryFetched);
+  ASSERT_TRUE(address);
+
+  ASSERT_EQ(*entry, *address);
+}
+
+}  // namespace
```
pivotal-jbarrett, line 178: Trailing new line needed.

cppcache/src/PdxRemoteReader.cpp
```diff
@@ -34,571 +34,391 @@ PdxRemoteReader::~PdxRemoteReader() {
...
-      PdxLocalReader::resettoPdxHead();
-      m_dataInput->advanceCursor(position);
-      auto retVal = PdxLocalReader::readDate(fieldName);
-      PdxLocalReader::resettoPdxHead();
-      return retVal;
-    }
+  int position = m_pdxType->getFieldPosition(fieldName, m_offsetsBuffer,
+                                             m_offsetSize, m_serializedLength);
+
+  if (position == -1) {
```
pivotal-jbarrett, line 411: I wonder, if with all this refactoring, we should define a `constexpr` for this `-1` sentinel value. If you make further changes I would suggest you do this too, otherwise something to keep in your mind for later.

pivotal-jbarrett (commented):
Not going to approve yet since it is still in draft but wanted to say this is an awesome refactor and thanks for the new test.

gaussianrecurrence:
Closing this PR as the fix is already covered in GEODE-10276

PR #899, opened 2021-12-08
GEODE-9896: Upgrades dependencies.

pivotal-jbarrett:
* ACE 7.0.5
* Boost 1.78.0
* SQLite 3.37

PR #914, opened 2022-01-28
GEODE-9896: Bump sqlite to 3.37.2

moleske:

Seemed rude not to bump the other dependencies after doing ACE, so here is another one
May also conflict with #899

maybe some day I'll look into dependabot so this type of bumping can be automated


moleske:
failing lgtm is due to geode 1.14.2 no longer being available. this is fixed #912

pdxcodemonkey (approved):

PR #915, opened 2022-01-28
GEODE-9896: Bump benchmark to 1.6.1

moleske:

Just another old fashioned dependency bump after doing ACE


moleske:
closing for now cause I'm confused what happened and may not look for a bit.  don't let that stop you from looking into it though!

PR #917, opened 2022-02-02
GEODE-10004: support older geode versions

mmartell:
This PR is to update the cmake config files for tests/javaobject to allow building javaobject.jar against geode versions that pre-date the addition of throwing AuthenticationExpiredExceptions.


pdxcodemonkey (changes requested):
Tanzu GemFire Native is a proprietary product, and can't be referenced in the Apache Geode OSS repo.  This conditional build should happen based on whether or not Geode_VERSION is < 1.15.0-build.502 (I believe it was build 502 - you're probably safest to use like 500, which we know didn't yet have the new exception).

CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ set(PRODUCT_BUILDDATE "1970-01-01" CACHE STRING "Product build date")
 set(PRODUCT_SOURCE_REVISION "0000000000000000000000000000000000000000" CACHE STRING "Product source SHA")
 set(PRODUCT_SOURCE_REPOSITORY "<unspecified>" CACHE STRING "Product source branch")
 set(PRODUCT_BITS "${BUILD_BITS}bit")
+set(GEODE_AUTHEXPIREDEXCEPTION_VERSION "1.15.0-build.546" CACHE STRING "First build containing AuthenticationExpiredException")
```
pivotal-jbarrett, line 64: Why not just test for the existence of the exception class or not and set a "has feature" type flag?

pivotal-jbarrett (changes requested):

CMakeLists.txt
pdxcodemonkey: Just because that's a more complex fix, and neither of us really have a clue how to implement it using cmake.  Like how would we do this, go dump the class list from a jar file after we find Geode?  Which of the umpteen Geode jar files should we look in?  How do we run the jar tool at this stage via cmake?  After we do or do not find the exception class, how do we put that back into a cmake variable we can use to conditionally compile our Java?

pivotal-jbarrett: Ok, fair enough, but why the specific build number. When building locally on your machine, since Apache doesn't do binary releases, that build number is almost always 0. Just detect "1.15.0". We don't need to handle building sources between GAs.

pivotal-jbarrett: Also the GA release ends up with build 0 too.

mmartell: Good suggestion. Makes sense to use 1.15.0.

cmake/FindGeode.cmake
```diff
@@ -73,7 +73,7 @@ if(Geode_gfsh_EXECUTABLE)
     # TODO error checking
   else()
     if(var MATCHES "([0-9]+\\.[0-9]+\\.[0-9]+)")
-      set(Geode_VERSION "${CMAKE_MATCH_1}")
+      set(Geode_VERSION ${var})
```
pdxcodemonkey, line 76: What is this all about?  I don't understand why we need a change in the Geode find module...

pdxcodemonkey (changes requested):

cmake/FindGeode.cmake
pivotal-jbarrett: I agree, the group 1 has all the parts we care about in the version.

CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ set(PRODUCT_BUILDDATE "1970-01-01" CACHE STRING "Product build date")
 set(PRODUCT_SOURCE_REVISION "0000000000000000000000000000000000000000" CACHE STRING "Product source SHA")
 set(PRODUCT_SOURCE_REPOSITORY "<unspecified>" CACHE STRING "Product source branch")
 set(PRODUCT_BITS "${BUILD_BITS}bit")
+set(GEODE_AUTHEXPIREDEXCEPTION_VERSION "1.15.0" CACHE STRING "First build containing AuthenticationExpiredException")
```
pivotal-jbarrett, line 64: Why is this a cache variable and why in the root project. This should just be a string literal in the javaobject project.

tests/javaobject/CMakeLists.txt
```diff
@@ -22,6 +22,11 @@ include(UseJava)
 
 file(GLOB_RECURSE SOURCES "*.java")
 
+# Check for versions of GEODE that don't support AuthenticationExpiredException
+if (${Geode_VERSION} VERSION_LESS ${GEODE_AUTHEXPIREDEXCEPTION_VERSION})
```
pivotal-jbarrett, line 26: Use version literal, `"1.15.0"`, here.

pivotal-jbarrett (changes requested):

cmake/FindGeode.cmake
mmartell: Good catch! No longer need the build number since we're just comparing with first 3 parts.

tests/javaobject/CMakeLists.txt
mmartell: It needs to be a variable that we can override on the cmake config command line, for closed source builds:
   cmake ... -DGEODE_AUTHEXPIREDEXCEPTION_VERSION=9.15.0

CMakeLists.txt
pdxcodemonkey: See Martell's comment - this is a string variable because there may be different version(s) to test against, depending on which product is being built.

mmartell: Seems reasonable to put it in the cmake config file for the project it applies to.

mmartell: Removed.

tests/javaobject/CMakeLists.txt
mmartell: No longer need to define the variable. Switched to searching for the Java class.

cmake/FindGeode.cmake
```diff
@@ -73,7 +73,8 @@ if(Geode_gfsh_EXECUTABLE)
     # TODO error checking
   else()
     if(var MATCHES "([0-9]+\\.[0-9]+\\.[0-9]+)")
-      set(Geode_VERSION "${CMAKE_MATCH_1}")
+	  set(Geode_VERSION "${CMAKE_MATCH_1}")
+	  set(Geode_VERSION_FULL ${var})
```
pdxcodemonkey, line 77: Unless I'm mistaken, nothing in this change depends on Geode_VERSION any more, so there should be no changes necessary in this file.  Please remove, we don't want to accidentally introduce any new ways to break the build.

pdxcodemonkey (changes requested):

cmake/FindGeode.cmake
mmartell: Good catch!

pivotal-jbarrett (approved):
Hey this is pretty cool! The CMake way!

pdxcodemonkey (approved):

PR #918, opened 2022-02-03
GEODE-10016: Add map of threadId to threadName

mmartell:
Since all internally created native client threads are named, we should print the threadName instead of threadId. This will be extremely helpful to understanding the flow of messages since there are many background threads in the native client.

### DIFF
--- a/cppcache/src/DistributedSystemImpl.cpp
+++ b/cppcache/src/DistributedSystemImpl.cpp
@@ -38,6 +38,10 @@
 #include "util/Log.hpp"
 #include "version.h"
 
+namespace {
+std::map<std::thread::id, std::string> m_threadNames;
+}
+
 namespace apache {
 namespace geode {
 namespace client {
@@ -143,10 +147,23 @@ void DistributedSystemImpl::unregisterCliCallback(int appdomainId) {
   }
 }
 
-void DistributedSystemImpl::setThreadName(const std::string& threadName) {
+std::string DistributedSystemImpl::getThreadName(std::thread::id id) {
+  std::string threadName = m_threadNames[id];
+  if (threadName == "") {
+    std::stringstream ss;
+    ss << id;
+    threadName = ss.str();
+  }
+  return threadName;
+}
+
+void DistributedSystemImpl::setThreadName(const std::string& threadName,
+                                          std::thread::id tid) {
   if (threadName.empty()) {
     throw IllegalArgumentException("Thread name is empty.");
   }
+
+  m_threadNames[tid] = threadName;
 
 #if defined(HAVE_pthread_setname_np)
 
--- a/cppcache/src/DistributedSystemImpl.hpp
+++ b/cppcache/src/DistributedSystemImpl.hpp
@@ -47,7 +47,8 @@ using CliCallbackMethod = std::function<void(Cache&)>;
  */
 class DistributedSystemImpl {
  public:
-  static void setThreadName(const std::string& threadName);
+  static void setThreadName(const std::string& threadName, std::thread::id tid);
+  static std::string getThreadName(std::thread::id id);
 
   /**
    * @brief destructor
--- a/cppcache/src/ExpiryTaskManager.cpp
+++ b/cppcache/src/ExpiryTaskManager.cpp
@@ -51,7 +51,7 @@ void ExpiryTaskManager::start() {
   auto start_future = start_promise.get_future();
   runner_ = std::thread{[this, &start_promise] {
     start_promise.set_value(true);
-    DistributedSystemImpl::setThreadName(NC_ETM_Thread);
+    DistributedSystemImpl::setThreadName(NC_ETM_Thread, runner_.get_id());
 
     LOGFINE("ExpiryTaskManager thread is running.");
     io_context_.run();
--- a/cppcache/src/Log.cpp
+++ b/cppcache/src/Log.cpp
@@ -38,6 +38,7 @@
 #include <geode/ExceptionTypes.hpp>
 #include <geode/util/LogLevel.hpp>
 
+#include "DistributedSystemImpl.hpp"
 #include "geodeBanner.hpp"
 #include "util/chrono/time_point.hpp"
 
@@ -381,8 +382,10 @@ std::string Log::formatLogLine(LogLevel level) {
   msg << "[" << Log::levelToChars(level) << " "
       << std::put_time(&tm_val, "%Y/%m/%d %H:%M:%S") << '.' << std::setfill('0')
       << std::setw(6) << microseconds.count() << ' '
-      << std::put_time(&tm_val, "%z  ") << g_hostName << ":"
-      << boost::this_process::get_id() << " " << std::this_thread::get_id()
+      << std::put_time(&tm_val, "%z  ") << g_hostName
+      << ":"
+      << boost::this_process::get_id() << " "
+      << DistributedSystemImpl::getThreadName(std::this_thread::get_id())
       << "] ";
 
   return msg.str();
--- a/cppcache/src/ThreadPool.cpp
+++ b/cppcache/src/ThreadPool.cpp
@@ -16,6 +16,8 @@
  */
 #include "ThreadPool.hpp"
 
+#include <sstream>
+
 #include "DistributedSystemImpl.hpp"
 
 namespace apache {
@@ -29,7 +31,8 @@ ThreadPool::ThreadPool(size_t threadPoolSize)
   workers_.reserve(threadPoolSize);
 
   std::function<void()> executeWork = [this] {
-    DistributedSystemImpl::setThreadName(NC_Pool_Thread);
+    DistributedSystemImpl::setThreadName(NC_Pool_Thread,
+                                         std::this_thread::get_id());
     while (true) {
       std::unique_lock<decltype(queueMutex_)> lock(queueMutex_);
       queueCondition_.wait(lock,
--- a/cppcache/src/ClientMetadataService.cpp
+++ b/cppcache/src/ClientMetadataService.cpp
@@ -66,7 +66,7 @@ void ClientMetadataService::stop() {
 }
 
 void ClientMetadataService::svc() {
-  DistributedSystemImpl::setThreadName(NC_CMDSvcThread);
+  DistributedSystemImpl::setThreadName(NC_CMDSvcThread, m_thread.get_id());
 
   LOGINFO("ClientMetadataService started for pool " + m_pool->getName());
 
--- a/cppcache/src/EvictionController.cpp
+++ b/cppcache/src/EvictionController.cpp
@@ -65,7 +65,7 @@ void EvictionController::stop() {
 
 void EvictionController::svc() {
   std::mutex mutex;
-  DistributedSystemImpl::setThreadName(NC_EC_Thread);
+  DistributedSystemImpl::setThreadName(NC_EC_Thread, thread_.get_id());
 
   while (running_) {
     {
--- a/cppcache/src/Task.hpp
+++ b/cppcache/src/Task.hpp
@@ -68,7 +68,7 @@ class Task {
   }
 
   inline void svc(void) {
-    DistributedSystemImpl::setThreadName(threadName_);
+    DistributedSystemImpl::setThreadName(threadName_, thread_.get_id());
 
     if (appDomainContext_) {
       appDomainContext_->run(
--- a/cppcache/src/statistics/HostStatSampler.cpp
+++ b/cppcache/src/statistics/HostStatSampler.cpp
@@ -427,7 +427,8 @@ void HostStatSampler::checkDiskLimit() {
 }
 
 void HostStatSampler::svc(void) {
-  client::DistributedSystemImpl::setThreadName("NC HSS Thread");
+  client::DistributedSystemImpl::setThreadName("NC HSS Thread",
+                                               m_thread.get_id());
   try {
     // createArchiveFileName instead of getArchiveFileName here because
     // for the first time the sampler needs to add the pid to the filename
--- a/cppcache/src/statistics/PoolStatsSampler.cpp
+++ b/cppcache/src/statistics/PoolStatsSampler.cpp
@@ -47,7 +47,8 @@ PoolStatsSampler::PoolStatsSampler(milliseconds sampleRate, CacheImpl* cache,
           cache->getStatisticsManager().getStatisticsFactory()) {}
 
 void PoolStatsSampler::svc() {
-  client::DistributedSystemImpl::setThreadName(NC_PSS_Thread);
+  client::DistributedSystemImpl::setThreadName(NC_PSS_Thread,
+                                               m_thread.get_id());
   while (!m_stopRequested) {
     auto sampleStart = high_resolution_clock::now();
     putStatsInAdminRegion();


cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -38,6 +38,10 @@
 #include "util/Log.hpp"
 #include "version.h"
 
+namespace {
+std::map<std::thread::id, std::string> m_threadNames;
```
pivotal-jbarrett, line 42: Does not conform to naming convention.

pivotal-jbarrett (changes requested):

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -143,11 +147,24 @@ void DistributedSystemImpl::unregisterCliCallback(int appdomainId) {
   }
 }
 
-void DistributedSystemImpl::setThreadName(const std::string& threadName) {
+std::string DistributedSystemImpl::getThreadName(std::thread::id id) {
```
gaussianrecurrence, line 150: Is there any particular reason why you would want to obtain the thread name from another thread which is not the caller thread? Because if there is no reason, I'd say it'd be best to get the thread ID of the current thread within the function, and that would make syntax a little bit simpler.

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -143,11 +147,24 @@ void DistributedSystemImpl::unregisterCliCallback(int appdomainId) {
...
+  std::string threadName = m_threadNames[id];
+  if (threadName == "") {
+    std::stringstream ss;
+    ss << id;
+    threadName = ss.str();
+  }
+  return threadName;
+}
+
+void DistributedSystemImpl::setThreadName(const std::string& threadName,
```
gaussianrecurrence, line 160: Same comment regarding thread ID as for getThreadName, IMHO I don't see a reason why you would need to pass a thread ID here

cppcache/src/Log.cpp
```diff
@@ -382,7 +383,8 @@ std::string Log::formatLogLine(LogLevel level) {
       << std::put_time(&tm_val, "%Y/%m/%d %H:%M:%S") << '.' << std::setfill('0')
       << std::setw(6) << microseconds.count() << ' '
       << std::put_time(&tm_val, "%z  ") << g_hostName << ":"
-      << boost::this_process::get_id() << " " << std::this_thread::get_id()
+      << boost::this_process::get_id() << " "
+      << DistributedSystemImpl::getThreadName(std::this_thread::get_id())
```
gaussianrecurrence, line 387: I am not totally sure about this, but changing the logging format might break the GNMSG tool that @pdxcodemonkey implemented some time ago. I am not sure what level of support you are intending to provide for this tool, but might be  good to make sure that the tool could handle both log formats

cppcache/src/ThreadPool.cpp
```diff
@@ -16,6 +16,8 @@
  */
 #include "ThreadPool.hpp"
 
+#include <sstream>
```
gaussianrecurrence, line 19: I think this might be a remaining of some dev debugging. Is this needed anymore?

gaussianrecurrence (commented):
I've been wanting to introduce this change for a really long time. Thanks for putting some effort into this!

cppcache/src/DistributedSystemImpl.cpp
mmartell: I'm using this anonymous namespace to solve an apparent linking issue. Without it, this map is an unresolved reference in the clicache. Do you know of another way to solve that?

pivotal-jbarrett (changes requested):

cppcache/src/DistributedSystemImpl.cpp
pivotal-jbarrett: I am referring to the variable name. This is a global variable, now a member variable and even member variables aren't supported to have the old `m_` prefix.

mmartell: Good catch.

cppcache/src/DistributedSystemImpl.cpp
mmartell: Good catch! You're right. Don't need to pass in threadId.

cppcache/src/DistributedSystemImpl.cpp
mmartell: Good catch.

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -143,11 +148,26 @@ void DistributedSystemImpl::unregisterCliCallback(int appdomainId) {
   }
 }
 
+std::string DistributedSystemImpl::getThreadName() {
+  std::thread::id id = std::this_thread::get_id();
+  std::stringstream ss;
+  ss << id;
+  std::string threadName;
+  if (g_threadNames[id] != "") {
```
pivotal-jbarrett, line 156: This is not a thread safe access to a non-thread safe data structure. There is actually no reason to use a map with thread ID as key since that is pretty much what `thread_local` is.

cppcache/src/Log.cpp
mmartell: Good catch. gnmsg has been updated to handle either kind of log lines as below:
```[debug 2022/02/06 07:44:02.450497 -0800  FirstPro:20336 18736] EventId::EventId(000002894033BF60) - doInit=false, reserveSize=0, fullValueAfterDeltaFail=false
[fine 2022/02/06 07:44:02.453857 -0800  FirstPro:20336 18812 (NC ETM Thread)] ExpiryTaskManager thread is running.```

File: cppcache/src/DistributedSystemImpl.cpp [hunk 7]
```diff
@@ -143,11 +148,26 @@ void DistributedSystemImpl::unregisterCliCallback(int appdomainId) {
   }
 }
 
+std::string DistributedSystemImpl::getThreadName() {
```
pivotal-jbarrett, line 151: You can use `const std::string&` once you use the `thread_local` implementation.

pivotal-jbarrett (changes requested):

cppcache/src/ThreadPool.cpp
mmartell: You're right. Not needed.

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -143,11 +147,24 @@ void DistributedSystemImpl::unregisterCliCallback(int appdomainId) {
   }
 }
 
-void DistributedSystemImpl::setThreadName(const std::string& threadName) {
+std::string DistributedSystemImpl::getThreadName(std::thread::id id) {
+  std::string threadName = m_threadNames[id];
```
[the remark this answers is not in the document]
mmartell: Good catch!

cppcache/src/Log.cpp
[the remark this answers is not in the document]
mmartell: Yah this is a perfect case threadlocal storage. Just required moving the threadName and its accessors to the Log class. Which makes more sense anyway since logging is the only use of threadName.

cppcache/src/DistributedSystemImpl.hpp
```diff
@@ -24,6 +24,7 @@
 #include <memory>
 #include <mutex>
 #include <string>
+#include <thread>
```
pivotal-jbarrett, line 27: Drop this.

cppcache/src/ExpiryTaskManager.cpp
```diff
@@ -51,7 +51,7 @@ void ExpiryTaskManager::start() {
   auto start_future = start_promise.get_future();
   runner_ = std::thread{[this, &start_promise] {
     start_promise.set_value(true);
-    DistributedSystemImpl::setThreadName(NC_ETM_Thread);
+    Log::setThreadName(NC_ETM_Thread);
```
pivotal-jbarrett, line 54: Can we just inline the string constants, they aren't used anywhere else.

cppcache/src/Log.cpp
```diff
@@ -38,6 +38,7 @@
 #include <geode/ExceptionTypes.hpp>
 #include <geode/util/LogLevel.hpp>
 
+#include "DistributedSystemImpl.hpp"
```
pivotal-jbarrett, line 41: You can drop this too I think.

cppcache/src/DistributedSystemImpl.cpp
mmartell: Done.

cppcache/src/DistributedSystemImpl.cpp
mmartell: Using thread_local now.

cppcache/src/Log.cpp
```diff
@@ -60,6 +61,8 @@ static FILE* g_log = nullptr;
 
 static std::string g_hostName;
 
+static thread_local std::string g_threadName;
```
pivotal-jbarrett, line 64: See https://google.github.io/styleguide/cppguide.html#thread_local and https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables for naming guidelines. I think you want `kThreadName` but I am not too keen on that given it isn't really a constant.

cppcache/src/Log.cpp
```diff
@@ -369,6 +372,46 @@ LogLevel Log::charsToLevel(const std::string& chars) {
   }
 }
 
+void Log::setThreadName(const std::string& threadName) {
```
pivotal-jbarrett, line 375: Change signature to `std::string threadName` then `std::move` into `g_threadName`.

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -38,6 +38,10 @@
 #include "util/Log.hpp"
 #include "version.h"
 
+namespace {
+std::map<std::thread::id, std::string> g_threadNames;
```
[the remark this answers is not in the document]
mmartell: Can't acually use thread_local in DistributedSystemImpl anyway since it's used my clicache. C++/CLI used to support [System::ThreadLocalAttribute] like C#. But no longer supported.

cppcache/src/Log.cpp
```diff
@@ -378,12 +421,21 @@ std::string Log::formatLogLine(LogLevel level) {
       now - std::chrono::system_clock::from_time_t(secs));
   auto tm_val = apache::geode::util::chrono::localtime(secs);
 
+  std::thread::id id = std::this_thread::get_id();
```
pivotal-jbarrett, line 424: Why the variable?

cppcache/src/DistributedSystemImpl.hpp
```diff
@@ -48,6 +49,7 @@ using CliCallbackMethod = std::function<void(Cache&)>;
 class DistributedSystemImpl {
  public:
   static void setThreadName(const std::string& threadName);
```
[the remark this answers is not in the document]
mmartell: Indeed. I moved it to the Log class.

cppcache/src/Log.cpp
```diff
@@ -378,12 +421,21 @@ std::string Log::formatLogLine(LogLevel level) {
...
+    threadIdWithName = ss.str();
+  }
+
   msg << "[" << Log::levelToChars(level) << " "
       << std::put_time(&tm_val, "%Y/%m/%d %H:%M:%S") << '.' << std::setfill('0')
       << std::setw(6) << microseconds.count() << ' '
       << std::put_time(&tm_val, "%z  ") << g_hostName << ":"
-      << boost::this_process::get_id() << " " << std::this_thread::get_id()
-      << "] ";
+      << boost::this_process::get_id() << " " << threadIdWithName << "] ";
```
pivotal-jbarrett, line 438: just 
```c++
.. << std::this_thread::id << g_threadName << ...
```
Avoid the extra stream creation.

cppcache/src/SystemProperties.cpp
```diff
@@ -169,7 +169,8 @@ SystemProperties::SystemProperties(
       m_sslTrustStore(DefaultSslTrustStore),
       m_sslKeystorePassword(DefaultSslKeystorePassword),
       m_conflateEvents(DefaultConflateEvents),
-      m_threadPoolSize(DefaultThreadPoolSize),
+      // m_threadPoolSize(DefaultThreadPoolSize),
+      m_threadPoolSize(16),
```
pivotal-jbarrett, line 173: Looks like an unintentional change!

pivotal-jbarrett (changes requested):

cppcache/src/DistributedSystemImpl.cpp
```diff
@@ -143,11 +148,26 @@ void DistributedSystemImpl::unregisterCliCallback(int appdomainId) {
...
+  std::thread::id id = std::this_thread::get_id();
+  std::stringstream ss;
+  ss << id;
+  std::string threadName;
+  if (g_threadNames[id] != "") {
+    threadName = ss.str() + " (" + g_threadNames[id] + ")";
+  } else {
+    threadName = ss.str();
+  }
+  return threadName;
```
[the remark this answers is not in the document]
mmartell: I moved this logic to the formatLogLine function. Should be fine since most work is on the main thread has no name.

cppcache/src/DistributedSystemImpl.hpp
mmartell: Done

cppcache/src/ExpiryTaskManager.cpp
mmartell: Done

cppcache/src/Log.cpp
mmartell: Leaving this one to be consistent with name of several other variables in this file.

cppcache/src/Log.cpp
mmartell: Good catch. Inlined.

cppcache/src/SystemProperties.cpp
mmartell: Good catch. Yup that was for investigating impact of threadLocals on TSSDataOutput buffers. Reverted.

cppcache/src/Log.cpp
mmartell: Done

cppcache/src/Log.cpp
mmartell: Need to leave this alone because of the try/except block. If string is not const we get this error:

Error	C2712	Cannot use __try in functions that require object unwinding	apache-geode-static	C:\temp\geode-native\cppcache\src\Log.cpp	413

cppcache/src/Log.cpp
mmartell: Good catch. Done.

cppcache/src/ThreadPool.cpp
```diff
@@ -16,20 +16,19 @@
  */
 #include "ThreadPool.hpp"
 
+#include "./util/Log.hpp"
```
pivotal-jbarrett, line 19: ./ is redundant here.

pivotal-jbarrett (changes requested):

cppcache/src/ThreadPool.cpp
mmartell: Done

pivotal-jbarrett (approved):

PR #919, opened 2022-02-04
GEODE-10017: Fix servers restart within new ITs

gaussianrecurrence:
- Due to a rare race-condition described on the Jira case, it could
   happen that some test cases that needed to restart a/some server(s)
   could end up getting stuck.
 - This change introduces a mechanism to verify the server process has
   been actually stopped.
 - Also, removed sleeps on some TCs which were put in place as a WA.
 - Modified all of the new IT TCs that were restarting servers in order
   to use the above mentioned mechanism, so the race-condition is
   avoided.

### DIFF
--- a/cppcache/integration/test/FunctionExecutionTest.cpp
+++ b/cppcache/integration/test/FunctionExecutionTest.cpp
@@ -324,7 +324,9 @@ TEST(FunctionExecutionTest, testThatFunctionExecutionThrowsExceptionNonHA) {
       .withType("PARTITION")
       .execute();
 
-  cluster.getServers()[2].stop();
+  auto &targetServer = cluster.getServers()[2];
+  targetServer.stop();
+  targetServer.wait();
 
   auto cache = CacheFactory().create();
   auto poolFactory = cache.getPoolManager().createFactory();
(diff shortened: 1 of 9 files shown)


cppcache/integration/test/FunctionExecutionTest.cpp
```diff
@@ -324,7 +324,9 @@ TEST(FunctionExecutionTest, testThatFunctionExecutionThrowsExceptionNonHA) {
       .withType("PARTITION")
       .execute();
 
-  cluster.getServers()[2].stop();
+  auto &targetServer = cluster.getServers()[2];
+  targetServer.stop();
+  targetServer.wait();
```
pdxcodemonkey, line 329: Why this separation of concerns between stop() and wait()?  Is there ever a circumstance where we call stop() and don't intend for the process to actually be terminated yet when the function returns?

cppcache/integration/test/PdxInstanceTest.cpp
```diff
@@ -395,9 +395,8 @@ TEST(PdxInstanceTest, testInstancePutAfterRestart) {
     cv_status.wait(lock, [&status] { return !status; });
   }
 
-  std::this_thread::sleep_for(std::chrono::seconds{30});
-
   for (auto& server : cluster.getServers()) {
+    server.wait();
```
pdxcodemonkey, line 399: Okay I see this used separately here in several places, but... would it be necessary to call wait() in the other places if stop() just actually made sure the process was terminated?

pdxcodemonkey (changes requested):

gaussianrecurrence:

PR #920, opened 2022-02-05
 GEODE-10012: Avoid non-HA function retries

gaussianrecurrence:
- Currently, in the native client, if you configure retryAttemps to be >0 for your pool, non-HA function executions are retried. 
    This happens as retries are coded at the network layer.
 - Geode documentation states that retries should happen for functions
   which isHA=true and in the function execution logic. And this is how
   it works in the Java client API.
 - If a function is retried at the network layer, it could happen that due to bucket
   partition it could end up in a InternalFunctionInvocationTargetException
 - So, this commit changes the behavior, so non-HA functions are not
   retried.
 - Also, FunctionAttributes were introduced in order to improve code
   readability.
 - Moved function execution classes to a file of its own, in order to
   improve readability and modularity.
 - Solved a bug in which there could not exist 2 functions with the same
   name, defined on different clusters and with different function
   attributes.
 - Implemented several new ITs to test that non-HA function executions are not
   retried by the API.
 - These are the cases being tested:
    * Function execution with PR enabled:
      - A timeout is forced from the server-side, so the server closes
        the connection towards the client.
      - A timeout is forced in the client.
    * Function execution with PR disabled.

Co-authored-by: Jakov Varenina <jakov.varenina@est.tech>

### DIFF
--- a/cppcache/integration/test/FunctionExecutionTest.cpp
+++ b/cppcache/integration/test/FunctionExecutionTest.cpp
@@ -50,6 +50,7 @@ using apache::geode::client::NotConnectedException;
 using apache::geode::client::Region;
 using apache::geode::client::RegionShortcut;
 using apache::geode::client::ResultCollector;
+using apache::geode::client::TimeoutException;
 const int ON_SERVERS_TEST_REGION_ENTRIES_SIZE = 34;
 const int PARTITION_REGION_ENTRIES_SIZE = 113;
 
@@ -61,10 +62,34 @@ std::shared_ptr<Region> setupRegion(Cache &cache) {
   return region;
 }
 
-class TestResultCollector : public ResultCollector {
- private:
-  std::shared_ptr<CacheableVector> resultList;
+std::shared_ptr<CacheableVector> populateRegionReturnFilter(
+    const std::shared_ptr<Region> &region, const int numberOfPuts) {
+  auto routingObj = CacheableVector::create();
+  for (int i = 0; i < numberOfPuts; i++) {
+    region->put("KEY--" + std::to_string(i), "VALUE--" + std::to_string(i));
+    routingObj->push_back(CacheableKey::create("KEY--" + std::to_string(i)));
+  }
+  return routingObj;
+}
 
+const std::vector<std::string> serverResultsToStrings(
+    std::shared_ptr<CacheableVector> serverResults) {
+  std::vector<std::string> resultList;
+  for (auto result : *serverResults) {
+    auto resultArray = std::dynamic_pointer_cast<CacheableArrayList>(result);
+    for (decltype(resultArray->size()) i = 0; i < resultArray->size(); i++) {
+      auto value =
+          std::dynamic_pointer_cast<CacheableString>(resultArray->at(i));
+      if (value != nullptr) {
+        resultList.push_back(value->toString());
+      }
+    }
+  }
+
+  return resultList;
+}
+
+class TestResultCollector : public ResultCollector {
  public:
   TestResultCollector() : resultList(CacheableVector::create()) {}
 
@@ -74,6 +99,9 @@ class TestResultCollector : public ResultCollector {
   }
 
   virtual void addResult(const std::shared_ptr<Cacheable> &result) override {
+    LOGINFO("Before mutex lock!");
+    std::lock_guard<decltype(mutex_)> lock{mutex_};
+    LOGINFO("Adding a new result!");
     resultList->push_back(result);
   }
 
@@ -84,6 +112,14 @@ class TestResultCollector : public ResultCollector {
     throw Exception(
         "Clear should not be triggered when Function.isHa is set to false");
   }
+
+  void lock() { mutex_.lock(); }
+
+  void unlock() { mutex_.unlock(); }
+
+ protected:
+  std::mutex mutex_;
+  std::shared_ptr<CacheableVector> resultList;
 };
 
 TEST(FunctionExecutionTest, UnknownFunctionOnServer) {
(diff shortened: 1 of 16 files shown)


cppcache/src/ThinClientRegion.cpp
```diff
@@ -2852,22 +2853,25 @@ void ThinClientRegion::executeFunction(
...
-        new ChunkedFunctionExecutionResponse(reply, (getResult & 2) == 2, rc));
+        new ChunkedFunctionExecutionResponse(reply, funcAttrs.hasResult(), rc));
     reply.setChunkedResultHandler(resultCollector);
     reply.setTimeout(timeout);
     GfErrType err = GF_NOERR;
-    err = m_tcrdm->sendSyncRequest(*msg, reply, !(getResult & 1));
+
+    // Function failover logic is not handled in the network layer. That's why
+    // attemptFailover should be always be false when calling sendSyncRequest
+    err = m_tcrdm->sendSyncRequest(*msg, reply, false);
```
pdxcodemonkey, line 2874: If the comment here is true, why do we still need the attemptFailover parameter?

pdxcodemonkey (changes requested):

cppcache/integration/test/FunctionExecutionTest.cpp
```diff
@@ -74,6 +99,9 @@ class TestResultCollector : public ResultCollector {
   }
 
   virtual void addResult(const std::shared_ptr<Cacheable> &result) override {
+    LOGINFO("Before mutex lock!");
+    std::lock_guard<decltype(mutex_)> lock{mutex_};
+    LOGINFO("Adding a new result!");
```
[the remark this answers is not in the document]
pivotal-jbarrett: Tests shouldn't be logging period. If the behavior is worth logging then it should just be asserted.

cppcache/src/ExecutionImpl.cpp
```diff
@@ -406,24 +355,25 @@ void ExecutionImpl::addResults(
 }
 
 void ExecutionImpl::executeOnAllServers(const std::string& func,
-                                        uint8_t getResult,
+                                        FunctionAttributes funcAttrs,
```
pivotal-jbarrett, line 358: Is the copy construction cheaper or necessary here rather than a `const &`?

cppcache/src/ExecutionImpl.cpp
```diff
@@ -329,26 +281,23 @@ std::shared_ptr<ResultCollector> ExecutionImpl::execute(
...
   } else {
     throw IllegalStateException("Execution::execute: should not be here");
   }
   return m_rc;
 }
 
-GfErrType ExecutionImpl::getFuncAttributes(
-    const std::string& func, std::shared_ptr<std::vector<int8_t>>* attr) {
+GfErrType ExecutionImpl::getFuncAttributes(const std::string& func,
+                                           FunctionAttributes& attr) {
```
pivotal-jbarrett, line 300: Does it make more sense to have the function return `FunctionAttributes` and `throw` exceptions?

cppcache/src/ExecutionImpl.hpp
```diff
@@ -78,6 +78,20 @@ class ExecutionImpl {
...
+  std::shared_ptr<CacheableVector> executeOnPool(
+      const std::string& func, FunctionAttributes funcAttrs, int32_t retryAttempts,
+      std::chrono::milliseconds timeout = DEFAULT_QUERY_RESPONSE_TIMEOUT);
+
+  void executeOnAllServers(
+      const std::string& func, FunctionAttributes funcAttrs,
+      std::chrono::milliseconds timeout = DEFAULT_QUERY_RESPONSE_TIMEOUT);
+
+  FunctionAttributes getFunctionAttributes(
+      const std::string& func);
```
pivotal-jbarrett, line 91: Please use full names, like `functionId` here.

cppcache/src/ExecutionImpl.hpp
```diff
@@ -78,6 +78,20 @@ class ExecutionImpl {
...
+  FunctionAttributes updateFunctionAttributes(const std::string &funcName);
+  GfErrType getFuncAttributes(const std::string& func, FunctionAttributes& attr);
```
pivotal-jbarrett, line 93: `functionId` and `functionAttributes` please.

cppcache/src/FunctionExecution.cpp
```diff
@@ -0,0 +1,162 @@
...
+    std::chrono::milliseconds timeout, std::shared_ptr<Cacheable> args,
+    TcrEndpoint* ep, ThinClientPoolDM* poolDM,
+    std::shared_ptr<std::recursive_mutex> resultCollectorMutex,
+    std::shared_ptr<ResultCollector> resultCollector,
+    std::shared_ptr<UserAttributes> userAttr) {
+  exceptionMsg_.clear();
+  resultCollectorMutex_ = std::move(resultCollectorMutex);
+  resultCollector_ = resultCollector;
+  error_ = GF_NOTCON;
+  funcName_ = funcName;
```
pivotal-jbarrett, line 43: Again, full names here.

cppcache/src/ThinClientRegion.cpp
```diff
@@ -2907,18 +2911,20 @@ void ThinClientRegion::executeFunction(
...
       } else if (err == GF_TIMEOUT) {
-        LOGINFO("function timeout. Name: %s, timeout: %s, params: %" PRIu8
-                ", "
-                "retryAttempts: %d ",
-                func.c_str(), to_string(timeout).c_str(), getResult,
-                retryAttempts);
+        LOGINFO(
+            "function timeout. Name: %s, timeout: %s, FunctionState: %" PRIu8
+            ", "
+            "retryAttempts: %d ",
```
pivotal-jbarrett, line 2917: Cleanup the formatting here.

pivotal-jbarrett (changes requested):
Overall I love the changes here, especially removing all the magic bit comparison stuff and the global.

I am concerned with some inconsistencies in style.
- Use fully qualified names and not abbreviations for symbols. Like `functionId` rather than `funcId`.
- Pass values by `const &` unless the value is going to be intentionally copied (or the copy is more expensive than a ref). 
- Return values rather than copy out by parameter. Even if that means using `std::tuple`.
- When pass by value is initializing a copied value use `std:move` explicitly.

I also want to call out terminology differences we should correct, specifically the `functionName` usage. The server and Java client refers to the function's identification as `functionId`. The `Function` interface has a `getId()` method, not `getName()`. I think it is important to use terms consistent with the rest of the product. So anywhere we reference the function by its symbolic identification lets call it `functionId`.

gaussianrecurrence:

PR #924, opened 2022-02-15
GEODE-10040: Increase wait timeout for gfsh commands in the test framework

mmartell:
Occassionaly we see timeouts in the CI when running the legacy clicache tests. These appear to mostly be timeouts waiting for gfsh commands to complete.

The error message varies depending on what branch you're on:

- develop: "ExecuteGfsh: Timeout, killing 4604"
- 10.2.7: "Timed out waiting for Java cacheserver to stop"


pdxcodemonkey (approved):
Any idea why on Earth it might take longer than *60 seconds* for a server or locator to stop?  I mean, let's make the change, because what could it hurt (?), but wow that seems excessive.

PR #927, opened 2022-02-21
GEODE-10073: Print bytes of ClientConnectionRequest/Response in logs

pdxcodemonkey:
- Dump the bytes at debug log level, so we can parse
- Also added parsing of these to gnmsg tool


mmartell (approved):
I like capturing the class::function in the log.

PR #928, opened 2022-02-22
GEODE-10074: Remove ACE remainings

gaussianrecurrence:
- Removed all the ACE includes along the repository
 - Removed ACE from CMake project.
 - Removed ACE dependency.
 - There were a couple of files in which ACE's strcmp was used.
   So, its usage was replaced to to std::string::operator==


cppcache/CMakeLists.txt
```diff
@@ -89,7 +89,6 @@ elseif (WIN32)
 endif()
 
 target_link_libraries(_apache-geode INTERFACE
-  ACE::ACE
```
pivotal-jbarrett, line 92: Tears of joy!

cppcache/integration-test/CMakeLists.txt
```diff
@@ -69,6 +67,10 @@ add_clangformat(unit_test_callbacks)
 
 enable_testing()
 
+if(UNIX AND NOT APPLE)
```
pivotal-jbarrett, line 70: This seems a little wonky... Is this perhaps a Linux or GCC specific dependency? It is also likely that our dependency or `rt` comes from another library and if so that is where this dependency should be added. Also, see root CMakeLists.txt for some of these more platform specific dependencies.

cppcache/integration-test/fw_helper.hpp
```diff
@@ -107,7 +105,7 @@ BEGIN_TEST.
 #define ASSERT_EQ(x, y) _ASSERT_EQ(x, y, __LINE__, __FILE__)
 
 #define ASSERT_STREQ(x, y)                                            \
-  ASSERT((strcmp(x, y) == 0),                                         \
+  ASSERT(x == y,                                         \
```
pivotal-jbarrett, line 108: I would be a little hesitant to change the expected type of x and y from c-string to `std::string`, or really anything with `==` operator overload.

What about `std::strcmp`?

cppcache/src/CppCacheLibrary.cpp
```diff
@@ -36,18 +33,15 @@ namespace apache {
...
-}
+void CppCacheLibrary::closeLib(void) {}
 
 std::string CppCacheLibrary::initProductLibDir() {
   // otherwise... get the DLL path, and work backwards from it.
-  char buffer[PATH_MAX + 1];
+  char buffer[4097];
   buffer[0] = '\0';
-  DllMainGetPath(buffer, PATH_MAX);
+  DllMainGetPath(buffer, 4096);
```
pivotal-jbarrett, line 44: use `sizeof(buffer)` to keep them in sync.

cppcache/src/CppCacheLibrary.cpp
```diff
@@ -36,18 +33,15 @@ namespace apache {
...
-void CppCacheLibrary::closeLib(void) {
-  // DO NOT CALL ACE::fini() HERE!
-  // Things might be using ace beyond the life of Geode.
-}
+void CppCacheLibrary::closeLib(void) {}
 
 std::string CppCacheLibrary::initProductLibDir() {
   // otherwise... get the DLL path, and work backwards from it.
-  char buffer[PATH_MAX + 1];
+  char buffer[4097];
```
pivotal-jbarrett, line 42: How did you arrive at 4097?
Perhaps refactoring these implementations of `DllMainGetPath` to allocate and return appropriately sized `std::string` values?
For Windows the old `MAX_PATH`, 260, limit is gone and [`GetModulFileName`](https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamea#return-value) will set an error condition if the buffer it was given was too small. We could increase he buffer as needed to get the full path and the initialize a `std::string` from that buffer before deallocating it.
For POSIX the [`realpath`](https://pubs.opengroup.org/onlinepubs/009696799/functions/realpath.html) method can allocate the buffer for us by using `NULL` for the `resolved_path` parameter, we just need to free it after initializing an `std:string` with it.

pivotal-jbarrett (changes requested):

cppcache/integration-test/CMakeLists.txt
pivotal-jbarrett: Perhaps something more like this?
```
find_library(LIB_RT  rt)
  if (LIBRT)
    target_link_libraries(<target> ${LIB_RT})
  endif()
```

gaussianrecurrence: It's true that my solution for this is not the best one.
Regarding root CMakeLists.txt I don't think it makes sense, as rt is only needed for the shared memory API, which is used only by integration-tests. I'll try the find_library approach for this

cppcache/src/CppCacheLibrary.cpp
gaussianrecurrence: Ups, my bad, I did hardcode this while testing and forgot to properly solve it. Thanks for pointing it out :)

cppcache/src/CppCacheLibrary.cpp
gaussianrecurrence: See below comment

cppcache/integration-test/fw_helper.hpp
gaussianrecurrence: Well thing is this ASSERT_STREQ macro is only used in one place within integration-tests, so I'd rather remove this macro and use operator== in-place.
For reference here is the only place where it's used: https://github.com/apache/geode-native/blob/a072c8f69aa59335ab0b134f435f174c97d0ac22/cppcache/integration-test/testSystemProperties.cpp#L61

As you can see it's safe to remove this macro and use operator== from std::string

pivotal-jbarrett: Yes! So that! One less macro!

cppcache/integration-test/CMakeLists.txt
pivotal-jbarrett: Oh, well then if it is just these legacy tests... hack away.

pivotal-jbarrett (approved):

pdxcodemonkey:
@gaussianrecurrence please rebase to resolve conflicts, and I will merge ASAP.  Thanks!

gaussianrecurrence:
> @gaussianrecurrence please rebase to resolve conflicts, and I will merge ASAP. Thanks!
Sorry for the huge delay. I've updated the branch and  solved the conflict.

PR #929, opened 2022-02-22
GEODE-10074: Fix regex issue on Linux systems

pdxcodemonkey:
- Hex dump of a pointer on Linux is prefixed with '0x', so need to
  allow x or X character


mmartell (approved):

gaussianrecurrence:
Hi @pdxcodemonkey

I am not sure if you got wrong the ticket number since I created [GEODE-10074](https://issues.apache.org/jira/browse/GEODE-10074) :S

pdxcodemonkey:
Oops!  Sorry, yes this should be 10073.

PR #930, opened 2022-02-23
GEODE-10044: Allow FunctionAttributes update

gaussianrecurrence:
- Moved FunctionExecution and OnRegionFunctionExecution into a separate
   file in order to improve readability and code maintenance.
 - Moved ChunkedFunctionExecutionResponse into a separate file in order
   to improve readability and code maintenance.
 - Introduced FunctionAttributes class to improve code readability.
 - Modified code to introduce the usage of FunctionAttributes.
 - Modified TcrMessageTest UT to work with the inclusion of
   FunctionAttributes.
 - Implemented a mechanism to cleanup function attributes.
 - A new exception FunctionAttributesMismatchException is added to be
   thrown whenever the server detects a mismatch for the function
   attributes.
 - Now whenever FunctionAttributesMismatchException is catched while
   executing a function, its function attributes are cleaned up, so the
   client will request an update the next execution.
 - Added a new IT to verify that chaging a function attributes on the
   fly is possible.


gaussianrecurrence:
The rationale behind this change is allowing the server function upgrade scenario, which was previously discussed in here: [https://www.mail-archive.com/dev@geode.apache.org/msg26144.html](https://www.mail-archive.com/dev@geode.apache.org/msg26144.html)

gaussianrecurrence:

PR #932, opened 2022-02-24
gnmsg now parses all observed locator requests and responses

pdxcodemonkey:
LocatorListRequest/Response, QueueConnectionRequest/Response, and ClientConnectionRequest/Response are now handled.  Other messages should be reported out and not parsed, but will not raise an exception.


mmartell (approved):
Looks good. Only thing missing from ThinClientLocatorHelper is GetAllServersRequest/Response.

pdxcodemonkey:
Right - I don't know (yet?) under what circumstances geode-native sends GetAllServers, so I will add decoding for it when we find a log file in the wild that has it.  Or, someday, I will get around to writing a test that forces a geode-native code path that sends it, then decode it.

PR #936, opened 2022-02-25
GEODE-10085: Don't start JmxManager if already running

mmartell:
This adds support for multiple locators in the new .NET test framework.

### DIFF
--- a/clicache/integration-test2/Cluster.cs
+++ b/clicache/integration-test2/Cluster.cs
@@ -70,8 +70,9 @@ namespace Apache.Geode.Client.IntegrationTests
 
             for (var i = 0; i < locatorCount_; i++)
             {
+                var startJmxManager = (i == 0);
                 var locator = new Locator(this, new List<Locator>(),
-                    name_ + "/locator/" + i.ToString());
+                    name_ + "/locator/" + i.ToString(), startJmxManager);
                 locators_.Add(locator);
                 success = (locator.Start() == 0);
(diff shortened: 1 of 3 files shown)


mmartell:
Converting to Draft until fix also implemented in new cppcache test framework.

clicache/integration-test2/Cluster.cs
```diff
@@ -70,8 +70,9 @@ private bool StartLocators()
 
             for (var i = 0; i < locatorCount_; i++)
             {
+                var startJmxManager = (i == 0);
```
pdxcodemonkey, line 73: Okay, a few problems here:
i) I'd really like to do this without introducing a boolean parameter, though I understand it may be unavoidable.
ii) We don't need a new startJmxManager here, just pass i == 0 as the last parameter to the locator ctor
iii) This was here before you got to it, but this loop is still broken for multiple locators, because the success variable is broken.  If you start two, and the first one fails but the second one works, you've buried an error.  Please break out of the for loop if success is false.

pdxcodemonkey (changes requested):

clicache/integration-test2/Cluster.cs
mmartell: i) You can't start the JmxManager in gfsh independent of a locator.
iii) Good catch!

pdxcodemonkey (approved):

PR #938, opened 2022-03-02
GEODE-10096: Replaced macro definitions with enumerations.

mreddington:


cppcache/src/TcrConnection.hpp
```diff
@@ -33,22 +33,28 @@
...
+  REPLY_AUTHENTICATION_FAILED = 63,
+  REPLY_DUPLICATE_DURABLE_CLIENT = 64,
+  SUCCESSFUL_SERVER_TO_CLIENT = 105,
+  UNSUCCESSFUL_SERVER_TO_CLIENT = 106
+};
+enum class security : ::std::uint8_t {
+  SECURITY_CREDENTIALS_NONE = 0,
+  SECURITY_CREDENTIALS_NORMAL = 1,
+  SECURITY_MULTIUSER_NOTIFICATIONCHANNEL = 3
+};
```
pdxcodemonkey, line 57: Really love the separation of concerns here, it's much easier now to tell in the implementation that the switch statements are correct, for instance.  Quick question - do any of these values need to be in the header?  They don't seem to be referenced anywhere outside of TcrConnection code...

pdxcodemonkey (changes requested):

cppcache/src/TcrConnection.hpp
pdxcodemonkey: Ugh, build break - looks like REPLY_SSL_ENABLED is also used in ThinClientLocatorHelper :|

cppcache/src/TcrConnection.cpp
```diff
@@ -380,33 +386,33 @@ bool TcrConnection::initTcrConnection(
     }
 
     switch (acceptanceCode[0]) {
-      case REPLY_OK:
-      case SUCCESSFUL_SERVER_TO_CLIENT:
+      case static_cast<int>(acceptance_codes::REPLY_OK):
```
pivotal-jbarrett, line 389: If we have to static cast this switch case values than either `acceptanceCode[0]` is the wrong type or all the values of `acceptanceType[0]` are not covered by a single `enum` and the use of `enum` probably isn't appropriate.

pivotal-jbarrett (changes requested):

cppcache/src/TcrConnection.cpp
pdxcodemonkey: Oh right, we could just cast the byte we read rather than cast in every case statement.  That's way better.

mreddington: At the very least, an enum class doesn't seem appropriate.

pivotal-jbarrett: Usage of `enum class` should be preferred to scope the constants and avoid collision with others not to mention avoid the strangely behaving type decaying old school `enum` does.

pdxcodemonkey: Exactly, it's good already, just cast in one place instead of 5 (6?).

pivotal-jbarrett (changes requested):

cppcache/src/TcrConnection.hpp
```diff
@@ -33,22 +33,28 @@
...
-#define REPLY_SSL_ENABLED 21
-#define REPLY_AUTHENTICATION_REQUIRED 62
-#define REPLY_AUTHENTICATION_FAILED 63
-#define REPLY_DUPLICATE_DURABLE_CLIENT 64
-
-#define SECURITY_CREDENTIALS_NONE 0
-#define SECURITY_CREDENTIALS_NORMAL 1
-#define SECURITY_MULTIUSER_NOTIFICATIONCHANNEL 3
+
+enum class acceptor : ::std::int8_t {
```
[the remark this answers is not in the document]
pdxcodemonkey: I prefer 'Acceptor' to 'acceptor', but it's probably worth having a short conversation about what parts of the style guide we actually expect to conform to.  Directly below the blurb about type names, for instance, the guide dictates snake_case for variable names, and prefixing named constants with 'k', both of which I would give a "hard no" vote....

cppcache/src/TcrConnection.hpp
mreddington: Good call, I moved what I could to the source file.

cppcache/src/TcrConnection.hpp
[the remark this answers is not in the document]
pivotal-jbarrett: Yup, proposal to amend our guide have been made in the past. Send them to dev@geode mailing list. Then base on the feedback adjust the CONTRIBUTING.md style section with any amendments.

cppcache/src/TcrConnection.cpp
mreddington: Jesus, I'm getting rusty...

cppcache/src/util/to_underlying.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+/* Note: this implementation can be thrown away once we upgrade to C++14. */
+
+template< class T >
+using underlying_type_t = typename std::underlying_type<T>::type;
+
+/* Note: this implementation can be thrown away once we upgrade to C++23. */
+
+template< class Enum >
+constexpr underlying_type_t<Enum> to_underlying( Enum e ) noexcept {
```
pivotal-jbarrett, line 31: I am torn on this... I both like and dislike it.
Dislikes:
- The name, underlying doesn't feel right, but I can't put my finger on it. `to_ordinal`?
- It sort of undoes one of the key features of `enum class` by casting the enum something, you don't really see what, that can decay to another type.
- Doesn't have a reverse like `static_cast` does so there is no mirror operation going back and forth between enum and ordinal value.
Like:
- Hides the mess of casting to a int type.
🤷

mreddington: Counterpoints:

* This is a direct copy of an STL utility we just don't have in C++11.
* Look where I'm using it - at some point, we need to write these enum values to the byte stream. Now there are two ways we can do that:
  - We can map an enum to a value, and a value to an enum, like via some switch statement. If we do that, then we don't need to specify the enum values themselves, since we'd never cast the enum to or from an integer type. This is just redundant in every way.
  - We cast the enum where we write bytes to the wire.
* It doesn't need a reverse. You can static cast any integer to an enum and the compiler will let you know if you're narrowing, but if you want to cast from an enum to the numeric type used to implement it, that's built into the enum type but you wouldn't know it, or can't rely on it since it can change, either by the compiler, or in the case of an enum class, the implementer. The compiler always knows, so this leverages deduction to get it right for you.

But I feel you. I don't like the use of enums overall. There's just something about them that don't seem "good enough". I don't think C++ would have had enumerations in the form we have them today if we didn't inherit from C, so I'm always aware of the loose type safety they come with.

pivotal-jbarrett: Oh, I wasn't aware they added this in 14. Well then... if the C++ gods like it then we should too. Thanks!

PR #943, opened 2022-03-04
GEODE-10098: Removed explicit new and delete in favor of unique_ptr

mreddington:
I also got rid of an auto-pointer type class that could be replaced with unique_ptr.


cppcache/src/TcrConnection.cpp
```diff
@@ -686,30 +682,27 @@ char* TcrConnection::readMessage(size_t* recvLen,
...
   std::chrono::microseconds mesgBodyTimeout = receiveTimeoutSec;
   if (isNotificationMessage) {
     mesgBodyTimeout = receiveTimeoutSec * DEFAULT_TIMEOUT_RETRIES;
   }
-  error = receiveData(fullMessage + HEADER_LENGTH, msgLen, mesgBodyTimeout);
+  error =
+      receiveData(fullMessage.get() + HEADER_LENGTH, msgLen, mesgBodyTimeout);
   if (error != CONN_NOERR) {
-    delete[] fullMessage;
+    fullMessage.reset();
```
pdxcodemonkey, line 704: Technically not necessary - dunno how attached you are to this line, I suppose it goes some small way towards documenting what we're doing here...

pdxcodemonkey (changes requested):

cppcache/src/CacheXmlParser.cpp
```diff
@@ -494,8 +491,7 @@ void CacheXmlParser::setAttributes(Cache *) {}
  *
  */
 void CacheXmlParser::create(Cache *cache) {
-  // use DeleteObject class to delete cacheCreation_ in case of exceptions
-  DeleteObject<CacheXmlCreation> delCacheCreation(cacheCreation_);
+  std::unique_ptr<CacheXmlCreation> delCacheCreation(cacheCreation_);
```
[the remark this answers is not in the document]
pivotal-jbarrett: Yes, It looks like the purpose here was to prevent the `cache` from being leaked if there was an exception. I suspect the caller is already doing this if it's a smart pointer already. Perhaps a `Cache&` is even more appropriate to void even needing the `nullptr` check.

cppcache/src/TcrConnection.cpp
```diff
@@ -531,11 +532,9 @@ ConnErrType TcrConnection::sendData(const char* buffer, size_t length,
   return CONN_NOERR;
 }
 
-char* TcrConnection::sendRequest(const char* buffer, size_t len,
-                                 size_t* recvLen,
-                                 std::chrono::microseconds sendTimeoutSec,
-                                 std::chrono::microseconds receiveTimeoutSec,
-                                 int32_t request) {
+std::tuple<std::unique_ptr<char[]>, std::size_t> TcrConnection::sendRequest(
```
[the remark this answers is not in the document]
pivotal-jbarrett: Why not `std::vector`, use that in several other places for this exact same purpose.

cppcache/src/TcrConnection.cpp
```diff
@@ -659,16 +656,15 @@ char* TcrConnection::readMessage(size_t* recvLen,
       if (isNotificationMessage) {
         // fix #752 - do not throw periodic TimeoutException for subscription
         // channels to avoid frequent stack trace processing.
-        return nullptr;
+        return {nullptr, 0, error};
```
pivotal-jbarrett, line 659: I love tuple!

cppcache/src/TcrConnection.cpp
pivotal-jbarrett: Agreed, let the smart pointer do the work later. It doesn't do anything for us really to do it here explicitly.

cppcache/src/TcrMessage.cpp
```diff
@@ -1373,14 +1360,10 @@ void TcrMessage::handleByteArrayResponse(
 
     case TcrMessage::LOCAL_CREATE:
     case TcrMessage::LOCAL_UPDATE: {
-      int32_t regionLen = input.readInt32();
+      m_regionName.resize(input.readInt32());
```
pivotal-jbarrett, line 1363: What happens in the off chance this read a negative number?

cppcache/src/TcrMessage.cpp
```diff
@@ -1373,14 +1360,10 @@ void TcrMessage::handleByteArrayResponse(
...
       input.advanceCursor(1);  // ignore byte
-      char* regname = nullptr;
-      regname = new char[regionLen + 1];
-      DeleteArray<char> delRegName(regname);
-      input.readBytesOnly(reinterpret_cast<int8_t*>(regname), regionLen);
-      regname[regionLen] = '\0';
-      m_regionName = regname;
+      input.readBytesOnly(reinterpret_cast<int8_t*>(&m_regionName[0]),
```
pivotal-jbarrett, line 1365: Seems very unsafe! Do we know that all implementations of `std::string` have allocated a contiguous space large enough to do this? There is a reason that `c_str()` and `data()` are `const`.  Is the null termination correctly handled here?

cppcache/src/util/make_unique.hpp
```diff
@@ -0,0 +1,57 @@
...
+#ifndef UTIL_MAKE_UNIQUE_H
+#define UTIL_MAKE_UNIQUE_H
+
+#include <cstddef>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+// N3656 standard revision of make_unique.
+
```
pivotal-jbarrett, line 27: Should we put this in a namespace?

pivotal-jbarrett (changes requested):

cppcache/src/TcrConnection.cpp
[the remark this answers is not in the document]
mreddington: @pivotal-jbarrett You know, when I look back at how I implemented this, I made this type assuming `DataInput` took ownership of the pointer. I later found out it doesn't, but then I guess I forgot to scrap this in favor of `vector`. I know @pdxcodemonkey mentioned `vector` at some point when he asked me to do this, but the words didn't leave a lasting impression. An oversight; `vector` is better.

cppcache/src/TcrConnection.cpp
pdxcodemonkey: I do to as a concept, use it all the time in my Python code.  Just not a fan of the C++ syntax, like I said I find it clumsy.  Of course, that's true of a lot of the language :).

cppcache/src/util/make_unique.hpp
pdxcodemonkey: Have a look at how it was done in this other pending PR:  https://github.com/apache/geode-native/pull/891/files

Might be good to ping @gaussianrecurrence and coordinate before there's a conflict, or two definitions in the code base(!).

cppcache/src/TcrConnection.cpp
pivotal-jbarrett: C++17 makes it all good. 
https://en.cppreference.com/w/cpp/language/structured_binding

cppcache/src/util/make_unique.hpp
pivotal-jbarrett: There is an internal shared library, that used go be static linked into the main library, but is still linked into all the test libraries. It was intended to be a place to put common utils that both the main lib and test code could use without exporting them as symbols from the main lib. It is the perfect place to consolidate `make_unique` for both the main lib and tests. The two files currently in this library are obsolete and should be deleted. It would also need to be added back as a library dependency to `_apache-geode` library.

pivotal-jbarrett: When I say "internal shared library" I don't me it is compile as a shared library. It is compiled static and linked static.

cppcache/src/TcrMessage.cpp
mreddington: The code never cared before. But I'll add a `min`.

cppcache/src/util/make_unique.hpp
mreddington: I'm going to update the PR to address the rest and work on this one tomorrow.

cppcache/src/CacheXmlCreation.cpp
```diff
@@ -34,37 +35,21 @@ void CacheXmlCreation::addPool(std::shared_ptr<PoolXmlCreation> pool) {
...
-  }
-}
-
-void CacheXmlCreation::setPdxIgnoreUnreadField(bool ignore) {
-  // m_cache->m_cacheImpl->setPdxIgnoreUnreadFields(ignore);
-  m_pdxIgnoreUnreadFields = ignore;
+  std::for_each(std::begin(pools), std::end(pools),
+                std::bind(&PoolXmlCreation::create,
+                          std::bind(&std::shared_ptr<PoolXmlCreation>::get,
+                                    std::placeholders::_1)));
```
pivotal-jbarrett, line 43: Hmm... Does this modern C++ style make this any more readable?

Would dereferencing the `shared_ptr`s as local variables prior to these calls help with readability?
Would a lambda rather than `std::bind` read easier?

Without the side by side diff I honestly can't say I understood what is going on here without unwinding all the algorithm magic going on here.

cppcache/src/CacheXmlParser.cpp
```diff
@@ -562,27 +555,15 @@ void CacheXmlParser::startCache(const xercesc::Attributes &attrs) {
...
-
-  auto pdxReadSerialized = getOptionalAttribute(attrs, READ_SERIALIZED);
-  if (!pdxReadSerialized.empty()) {
-    if (equal_ignore_case(pdxReadSerialized, "true")) {
-      cacheCreation_->setPdxReadSerialized(true);
-    } else {
-      cacheCreation_->setPdxReadSerialized(false);
-    }
-  }
+  CacheRegionHelper::getCacheImpl(cache_)->setPdxIgnoreUnreadFields(
```
pivotal-jbarrett, line 562: Are you certain this logic does not override the default value of property. The original code only change the value if the values was provided in the XML, preserving the default value in the object.

pivotal-jbarrett (changes requested):

PR #946, opened 2022-03-11
Fix regex used to read server handshake requests

pdxcodemonkey:


mmartell (approved):
We can parse Handshake again!

PR #952, opened 2022-04-05
Fix Ubuntu build break due to cmake upgrade

pdxcodemonkey:
- New cmake version changes treatment of CMAKE_INSTALL_LIBDIR variable,
  which now evaluates to different things for xerces build and main
  client build.  As a result, we build Xerces, then go looking to link
  to it in the wrong place.  Hard-coding the path to just 'lib' for now
  appears to fix the Ubuntu issue without breaking anything else.


echobravopapa (approved):

pivotal-jbarrett (approved):

dependencies/xerces-c/CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ target_include_directories(${PROJECT_NAME} SYSTEM INTERFACE
 )
 
 target_link_libraries(${PROJECT_NAME} INTERFACE
-  ${INSTALL_DIR}/${CMAKE_INSTALL_LIBDIR}/${CMAKE_STATIC_LIBRARY_PREFIX}${LIBRARY_NAME}${CMAKE_STATIC_LIBRARY_SUFFIX}
+  ${INSTALL_DIR}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}${LIBRARY_NAME}${CMAKE_STATIC_LIBRARY_SUFFIX}
```
pivotal-jbarrett, line 67: As I feared, this breaks on RHEL.

I think the easiest work around right now is to have the CI set the `CMAKE_INSTALL_PREFIX` to something not `/usr` or `/usr/local` and leave this file alone.

pivotal-jbarrett (changes requested):

pivotal-jbarrett (approved):

PR #954, opened 2022-04-06
GEODE-10218: Allow server debug

mmartell:
To assist with protocol analysis it is very helpful to be able to step through the server code in the Intellij debugger while running native client test code. However, to be allowed to set breakpoints in the server code, the server must be started with the special flag below:

```
--J=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=127.0.0.1:FREEPORT
```
This ticket is to add support for the above flag in the native client test frameworks.

### DIFF
--- a/clicache/integration-test2/Cluster.cs
+++ b/clicache/integration-test2/Cluster.cs
@@ -38,7 +38,7 @@ namespace Apache.Geode.Client.IntegrationTests
         internal string keyStorePassword_ = "apachegeode";
         internal string trustStore_ = Config.SslServerKeyPath + "/server_truststore_chained_root.jks";
         internal string trustStorePassword_ = "apachegeode";
-        internal bool allowAttach_ = false;
+        internal bool useDebugAgent_ = false;
 
         public Gfsh Gfsh { get; private set; }
 
@@ -54,7 +54,7 @@ namespace Apache.Geode.Client.IntegrationTests
             return poolFactory;
         }
 
-        public Cluster(ITestOutputHelper output, string name, int locatorCount, int serverCount, bool allowAttach = false)
+        public Cluster(ITestOutputHelper output, string name, int locatorCount, int serverCount, bool useDebugAgent = false)
         {
             started_ = false;
             Gfsh = new GfshExecute(output);
@@ -63,7 +63,7 @@ namespace Apache.Geode.Client.IntegrationTests
             locatorCount_ = locatorCount;
             serverCount_ = serverCount;
             locators_ = new List<Locator>();
-            allowAttach_ = allowAttach;
+            useDebugAgent_ = useDebugAgent;
         }
 
         private bool StartLocators()
@@ -73,7 +73,11 @@ namespace Apache.Geode.Client.IntegrationTests
             for (var i = 0; i < locatorCount_; i++)
             {
                 var locator = new Locator(this, locators_,
-                    name_ + "/locator/" + i.ToString(), i == 0, allowAttach_);
+                    name_ + "/locator/" + i.ToString(), i == 0);
+                if (useDebugAgent_)
+                {
+                    locator.UseDebugAgent = true;
+                }
                 locators_.Add(locator);
                 if (locator.Start() != 0 ) {
                     success = false;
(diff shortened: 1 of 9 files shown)


clicache/integration-test2/Cluster.cs
```diff
@@ -71,7 +73,7 @@ private bool StartLocators()
             for (var i = 0; i < locatorCount_; i++)
             {
                 var locator = new Locator(this, new List<Locator>(),
```
pivotal-jbarrett, line 75: The second parameter here should be your `locators_` list so that each locators is aware of the previously created.

clicache/integration-test2/Cluster.cs
```diff
@@ -71,7 +73,7 @@ private bool StartLocators()
...
-                    name_ + "/locator/" + i.ToString(), i == 0);
+                    name_ + "/locator/" + i.ToString(), i == 0, allowAttach_);
```
pivotal-jbarrett, line 76: This constructor is starting to get wild. What about a builder patter?

clicache/integration-test2/Cluster.cs
```diff
@@ -200,6 +204,7 @@ public int Start()
                     .withMaxHeap("256m")
                     .withJmxManagerPort(cluster_.jmxManagerPort)
                     .withJmxManagerStart(startJmxManager_)
+                    .withAllowAttach(allowAttach_, Address.address)
```
pivotal-jbarrett, line 207: What is behind the choice of the word "attach"? We are starting the debugger agent, why not use a name that suggests that, like `withDebugAgent(...)`.

pivotal-jbarrett (changes requested):

clicache/integration-test2/Cluster.cs
mmartell: Good catch!

clicache/integration-test2/Cluster.cs
mmartell: I do like the builder pattern. With nearly 30 potential parameters for the start locator command it seems prudent.

clicache/integration-test2/Cluster.cs
mmartell: I like it! Good rename.

clicache/integration-test2/Cluster.cs
mmartell: Done.

clicache/integration-test2/GfshTest.cs
```diff
@@ -77,13 +77,16 @@ public void StartLocatorStringsTest()
                 .withSslKeyStore("some/path/keystore.jks")
                 .withSslKeyStorePassword("password1")
                 .withSslTrustStore("some/path/truststore.jks")
-                .withSslTrustStorePassword("password2");
+                .withSslTrustStorePassword("password2")
+                .withDebugAgent(true, "address");
             s = locator.ToString();
+            var withAttachPortRemoved = s.Substring(0, s.LastIndexOf(":", s.Length-1, 6));
```
pivotal-jbarrett, line 83: "attachPort"?

cppcache/integration/framework/Cluster.h
```diff
@@ -141,6 +142,9 @@ class Cluster {
   Cluster(InitialLocators initialLocators, InitialServers initialServers,
           UseIpv6 useIpv6 = UseIpv6{false});
 
+  Cluster(InitialLocators initialLocators, InitialServers initialServers,
+          UseDebugAgent allowAttach);
```
pivotal-jbarrett, line 146: rename variable.

cppcache/integration/framework/Cluster.h
```diff
@@ -124,6 +124,7 @@ using Password = NamedType<std::string, struct PasswordParameter>;
 using CacheXMLFiles =
     NamedType<std::vector<std::string>, struct CacheXMLFilesParameter>;
 using UseIpv6 = NamedType<bool, struct UseIpv6Parameter>;
+using UseDebugAgent = NamedType<bool, struct AllowAttachParameter>;
```
pivotal-jbarrett, line 127: Rename struct.

pivotal-jbarrett (changes requested):

clicache/integration-test2/GfshTest.cs
mmartell: Good catch. No longer using attach nomenclature.

cppcache/integration/framework/Cluster.h
mmartell: Done

cppcache/integration/framework/Cluster.h
mmartell: Done

pivotal-jbarrett (approved):

PR #959, opened 2022-04-14
GEODE-10237: Remove no-ack from testThinClientTransactionsWithSticky

mmartell:
This removes the no-ack functionality from this test, which fails regularly in the CI.


pdxcodemonkey (changes requested):
This is, IMO, too much code to duplicate, and adding the new file rather than editing the existing one really obscures the nature of this change.  Please try just editing `ThinClientTrasactions.hpp` and resubmit.

mmartell:
> This is, IMO, too much code to duplicate, and adding the new file rather than editing the existing one really obscures the nature of this change. Please try just editing `ThinClientTrasactions.hpp` and resubmit.

Good catch Blake! Not sure why I thought I needed to use a new header file. As you point out, just needed to refactor the original header.

PR #967, opened 2022-05-10
GEODE-10292: Bump to boost 1.79.0

moleske:
Initially complained that it did not know what `boost::filesystem::ofstream` before adding the header file

### DIFF
--- a/cppcache/test/statistics/HostStatSamplerTest.cpp
+++ b/cppcache/test/statistics/HostStatSamplerTest.cpp
@@ -18,6 +18,7 @@
 #include <gmock/gmock.h>
 
 #include <boost/filesystem.hpp>
+#include <boost/filesystem/fstream.hpp>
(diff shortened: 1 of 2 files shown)


moleske:
surprised at the sad time for windows.  I thought on my windows 10/vs2019 personal machine things were fine, but guess will have to check again

pivotal-jbarrett (approved):

moleske:
ok I think the problem is [_WIN32_WINNT](https://github.com/apache/geode-native/blob/develop/dependencies/boost/CMakeLists.txt#L167) needs to be updated, but I don't quite understand why bumping boost would require changing what windows version to target

ebluo:
Hello Michael,

Any plan when this will be merged? Thanks!

moleske:
Probably not anytime soon as I haven't made time to to figure out this problem.  Will close the PR since I'm not making progress on it

PR #968, opened 2022-05-10
GEODE-10291: Add ubuntu 22.04 (jammy) to CI

moleske:
I don't have ami credentials nor concourse credentials to test any of this, so I will need some help if this should be wanted in CI

### DIFF
--- a/docker/ubuntu-22.04/Dockerfile
+++ b/docker/ubuntu-22.04/Dockerfile
@@ -60,7 +60,7 @@ RUN installer=$(mktemp) \
     && bash ${installer} --skip-license --prefix=/usr/local \
     && rm ${installer}
 
-ARG GEODE_VERSION=1.14.4
+ARG GEODE_VERSION=1.15.0
 ENV GEODE_HOME /apache-geode-${GEODE_VERSION}
 RUN curl -L -s "https://www.apache.org/dyn/closer.lua/geode/${GEODE_VERSION}/apache-geode-${GEODE_VERSION}.tgz?action=download" | tar -zxvf - --exclude javadoc
 
--- a/docker/ubuntu-22.04/Dockerfile
+++ b/docker/ubuntu-22.04/Dockerfile
@@ -46,7 +46,7 @@ RUN apt update && apt -yq full-upgrade && apt-get -y install \
         git \
         doxygen \
         graphviz \
-        clang-format-11 && \
+        clang-format-12 && \
     apt-get -y autoremove && \
     apt-get autoclean
 
--- a/docker/ubuntu-22.04/Dockerfile
+++ b/docker/ubuntu-22.04/Dockerfile
@@ -1,0 +1,71 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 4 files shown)


docker/ubuntu-22.04/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+        bellsoft-java11 \
+        git \
+        doxygen \
+        graphviz \
+        python3-pip \
+        clang-format-6.0 && \
+    apt-get -y autoremove && \
+    apt-get autoclean
+
+RUN pip3 install --upgrade pip && \
```
pivotal-jbarrett, line 54: I think you can cut out the coveralls bits. We haven't really been using it. Then you can cut out pip above too.

docker/ubuntu-22.04/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+        libc++-dev \
+        libc++abi-dev \
+        zlib1g-dev \
+        libssl-dev \
+        bellsoft-java11 \
+        git \
+        doxygen \
+        graphviz \
+        python3-pip \
+        clang-format-6.0 && \
```
pivotal-jbarrett, line 50: Gosh, this should probably be clang-format-11.

pivotal-jbarrett (changes requested):

docker/ubuntu-22.04/Dockerfile
moleske: lol.  I just copied pasted the ubuntu-20.04 dockerfile, so I didn't actually look at anything beyond the from line.  I'll clean these up.

Also I don't think these are used in ci after another look, they are only here for convenience.  Not sure if anyone cares, just an observation

pivotal-jbarrett: Not used by CI. I occasionally use them locally to compile under the alternative platforms when trying to debug a platform specific issue. They get very little love.

moleske: per some outside of here conversation with @pdxcodemonkey it is clang-format-12.  There's a lot of inconsistency in the other dockerfiles, and clang-format-12 is not in the default repository for ubuntu 16.04 and 18.04, so will only fix it for ubuntu 22.04 for now and pr the other dockerfiles later

docker/ubuntu-22.04/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN installer=$(mktemp) \
+    && curl -o ${installer} -L $(curl -s https://api.github.com/repos/Kitware/CMake/releases \
+        | grep -P -i 'browser_download_url.*cmake-\d+\.\d+\.\d+-linux-x86_64\.sh' \
+        | head -n 1 \
+        | cut -d : -f 2,3 \
+        | tr -d '"') \
+    && bash ${installer} --skip-license --prefix=/usr/local \
+    && rm ${installer}
+
+ARG GEODE_VERSION=1.14.4
```
gaussianrecurrence, line 63: Considering latest version is 1.15.0, maybe this needs to be updated

gaussianrecurrence (changes requested):
Seems good, except for the version thing

docker/ubuntu-22.04/Dockerfile
moleske: ah joys of forgetting about prs for months on end, can bump it

gaussianrecurrence (approved):

gaussianrecurrence:
Hi @moleske & @pivotal-jbarrett, it seems to me that @moleke addressed all the comments, so I am not sure is this can be merged already?

PR #973, opened 2022-05-18
GEODE-10076. Align String serialization with Java client

albertogpz:
The Geode C++ native client library serializes strings of PdxSerializable objects with DSCode CacheableString.
Nevertheless, the Java client, in order to serialize Strings, it analyzes them and different DSCodes are assigned depending on the contents (only ASCII vs not only ASCII chars) and length. In the native client, it's always set to CacheableString, whenever for example in the case of an ASCII string should be CacheableASCIIString.

This can provoke that two objects with the same contents created one by the native client and the other by the Java classes return false when compared with the equals method.

The writeString() method of the DataOutput() class of the Geode C++ native client library has been aligned with its counterpart in the Java library so that the serialization is the same in both cases.


gaussianrecurrence (commented):
Changes looks OK. However you need to re-execute Windows pipelines, as by the time you published the PR were failing

gaussianrecurrence (approved):
All pipelines are passing and the change looks fine

PR #974, opened 2022-06-09
GEODE-10371: C++ client: Improve dispersion on connections expiration

albertogpz:


cppcache/src/TcrConnection.cpp
```diff
@@ -63,23 +63,10 @@ bool useReplyTimeout(const apache::geode::client::TcrMessage& request) {
...
-  // NOLINTNEXTLINE(clang-analyzer-security.insecureAPI.rand): TODO replace
-  int random = rand() % numbers + 1;
-
-  if (random > 10) {
-    random = random - numbers;
-  }
-  return random;
+  std::random_device rd;
+  std::default_random_engine generator(rd());
+  std::uniform_int_distribution<int> distribution(-9, 9);
```
gaussianrecurrence, line 68: It'd be good to add a couple of constants to add some semantic meaning to the distribution limits

cppcache/test/TcrConnectionTest.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+  // Create several connections at the same time
+  const int connections = 100;
+  std::unique_ptr<TcrConnectionTest> tcrConnections[connections];
+  for (int i = 0; i < connections; i++) {
+    tcrConnections[i] =
+        std::unique_ptr<TcrConnectionTest>(new TcrConnectionTest(
+            static_cast<const TcrConnectionManager>(nullptr)));
+  }
+
```
gaussianrecurrence, line 52: This is a suggestion for a change to this case or maybe for another case, but given the goal of this change is to ensure the variance is evenly distributed along the connections created within a close period of time, I'd suggest instead of comparing with the first value, I'd rather sample the distribution and check no bucket in the sample has more than x times (i.e double) the expected amount.

PR #982, opened 2022-08-23
GEODE-10400: Fix metadata completion detection

gaussianrecurrence:
- Whenever execution server functions with isHA=false, hasResult=true
   and optimizeForWrite=true, if the metadata is incomplete a request of
   type EXECUTE_REGION_FUNCTION_SINGLE_HOP is sent with an invalid
   bucket set partition. This causes an
   InternalFunctionInvocationTargetException exception
 - In order to solve this issue, now, whenever calling
   groupByServerToBuckets, if there isn't a valid location for one of
   the buckets, it returns nullptr, triggering the fallback mechanism
   which would be sending EXECUTE_REGION_FUNCTION to one of the nodes
   instead.

### DIFF
--- a/cppcache/integration/test/FunctionExecutionTest.cpp
+++ b/cppcache/integration/test/FunctionExecutionTest.cpp
@@ -19,6 +19,7 @@
 #include <thread>
 
 #include <gtest/gtest.h>
+#include <gmock/gmock.h>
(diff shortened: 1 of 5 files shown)


gaussianrecurrence:

PR #984, opened 2022-09-12
GEODE-10414: Implement Region::putIfAbsent method

gaussianrecurrence:
- Moved PUT actions from LocalRegion into its own modules.
- Added PutIfAbsent action.
- Added the necessary scaffolding for put requests to return a value.
- Moved TcrMessagePut implementation into its own file. Now this message
  type accepts an operation as input.
- Moved TcrMessageDestroy implementation into its own file. Now this
  message accepts an operation as input.
- Also, updated TcrMessageDestroy operation format so it uses a byte
  part instead of an object.
- Added putIfAbsent method to the Region interface and its
  implementations both for LocalRegion and ThinClientRegion.
- Modified LocalRegion and ThinClientRegion to comply with the above
  code changes.
- Modified PUT's old value parsing to handle null pointer.
- Also replaced Cacheable type by Serializable in some places in order
  in order to take advantage of fast-forward definition.

### DIFF
--- a/cppcache/integration/test/RegionPutTest.cpp
+++ b/cppcache/integration/test/RegionPutTest.cpp
@@ -1,0 +1,294 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gmock/gmock.h>
+
+#include <chrono>
+#include <future>
+#include <iostream>
+#include <random>
+#include <thread>
+
+#include <gtest/gtest.h>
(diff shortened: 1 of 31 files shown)


cppcache/integration/test/RegionPutTest.cpp
```diff
@@ -0,0 +1,294 @@
...
+  EXPECT_THAT(converted->toString(), Eq(value->toString()));
+  auto& txId = txMgr->suspend();
+
+  EXPECT_THAT(region->putIfAbsent(key, value), IsNull());
+
+  txMgr->resume(txId);
+  EXPECT_THROW(txMgr->commit(), CommitConflictException);
+}
+
+TEST(RegionPutTest, testPutIfAbsentExistingEntryTX) {
```
albertogpz, line 265: What is the difference between this test case and `testPutIfAbsentExistingEntryCaching()`?
Why is there a TX suffix in the name?

cppcache/src/LocalRegion.cpp
```diff
@@ -367,11 +372,27 @@ void LocalRegion::put(const std::shared_ptr<CacheableKey>& key,
...
 }
 
+std::shared_ptr<Serializable> LocalRegion::putIfAbsent(
+    const std::shared_ptr<CacheableKey>& key,
+    const std::shared_ptr<Serializable>& value,
+    const std::shared_ptr<Serializable>& cbArg) {
+  auto opTime = startStatOpTime();
+  std::shared_ptr<Serializable> retValue;
+  std::shared_ptr<VersionTag> versionTag;
+  GfErrType err = putIfAbsentImpl(key, value, cbArg, retValue, -1,
```
albertogpz, line 382: Is the Impl suffix a common convention in C++ to name methods that implement the logic of public methods?
Given that it is a convention in Java to name classes that implement an interface, I would change the name to something else to avoid misunderstandings.

gaussianrecurrence:

